PR #172, opened 2026-08-10
#168: dev editors give visible feedback on a bad Save As name

### LINKED ISSUE #168: Dev editors: a name with a backslash saves nothing, with no visible feedback
🤖 Claude says:

## The bug — dev playtest, 2026-08-10

**Saving an item whose typed name contains a backslash fails with no visible feedback.** Press Save, nothing happens, the content is not on disk.

## Root cause (read from the code; not yet reproduced live)

The typed name **is** the filename, raw, at three save-new sites:

- `ItemEditorTool.gd:114` — `var path := dir + entered_name + ".tres"`
- `AttackEditorTool.gd:157` — same shape
- `ScenarioTool.gd:178` — via `ScenarioManager.scenario_path(entered)`

All funnel into `DevWidgets.save_over(resource, path)`, where the failure is real but invisible:

1. `make_dir_recursive_absolute(path.get_base_dir())` — `get_base_dir()` splits on `/`, so a `\` stays inside the leaf filename and only the intended parent directory is created.
2. `ResourceSaver.save()` on Windows resolves the `\` as a path separator into a directory that doesn't exist → error.
3. `save_over` does `push_error` and returns `false` — but **`push_error`/`push_warning` only surface in the editor's Output panel**, and every caller checks the bool solely to skip its on-success refresh. In the running game the dev-tool window shows nothing. (The existing refusals have the same blind spot: `refuse_existing_file`'s "already exists" message is a `push_warning` too.)

The same silent failure is reachable with every other Windows-illegal filename character (`/ : * ? " < > |`), and `/` additionally breaks the `res://` path on every platform.

## Fix shape (proposed, not built)

- **One shared gate beside the existing one:** a `DevWidgets.refuse_illegal_name(name, noun)` mirroring `refuse_existing_file`, called at the three save-new sites before the path is built — refuse the characters rather than sanitize them (a silently renamed file is the same surprise one step later, and catalogs key on the name).
- **Make refusals visible in the tool window** — a status label or equivalent, covering both this refusal and `refuse_existing_file`'s. Without this half, the fix replaces a silent failure with a differently-worded silent failure; the visibility gap **is** the reported bug.

Severity note: this is dev-tool-only, but the failure mode is authored content silently not persisting — the kind that costs a session's work before it's noticed. Worth a priority bump if it bites again before it's fixed.

— Claude (Fable 5) · 2026-08-10

Phaazoid:
Closes #168.

A name with a backslash used to save nothing with no visible feedback — `push_warning`/`push_error` only reach the editor's Output panel, invisible in the running game where these tools actually live.

**`DevWidgets.gd`:**
- New `refuse_illegal_name(name, noun, status_label, allow_slash)`, banning Windows-illegal characters. `/` is banned separately: the item/attack catalogs scan flat (`ResourceDir`), so a `/` there would silently land a file no scan ever finds again — but `Scenarios/fixtures/`-style subfolder names are a real, working feature, so `allow_slash` is scenario's opt-out from the same gate.
- `refuse_existing_file`, `save_over` and `delete_saved_file` each take an optional `status_label` so the exact string already being `push_*`'d also reaches the tool window. `save_over` clears the label on success.

**`ItemEditorTool.gd` / `AttackEditorTool.gd` / `ScenarioTool.gd`** each get a status label wired into their save/update/delete handlers — including the two pre-existing "needs a name"/"no file to update" warnings in the same handlers, since leaving those silent while fixing the new refusals would be an inconsistent fix for the same gap the issue describes.

**`Scenes/DevOverlay.tscn`:** one static status Label per tool, placed after each Save As row (matches the existing `LoadedScenarioLabel`/`_objective_warning` precedents for a warning-colored readout in this scene).

**Verification:** new `tests/dev/test_dev_tool_save_feedback.gd` — pure-logic cases on the gate (every banned character, a clean name, the `/` asymmetry), a check that `refuse_existing_file`'s message reaches a label, and one wired case per tool pressing the real Save As button with a backslash name. No case reaches `ResourceSaver.save`, matching this repo's no-disk-writes-in-tests convention. Falsified by hand: disabled the label mirror, confirmed the suite failed for the right reason, restored. `tests/dev`: 22/22. `tests/ui` spot-check (shared `DevOverlay.tscn`): 114/114.

**Flagged, not built:** `ScenarioManager.save_scenario` duplicates `save_over`'s dir-creation + save + error-handling instead of calling it (a Law #4 duplicate that predates this ticket) — its own disk-error path still won't reach the label. Out of scope here since `save_scenario` has more callers than this fix needs to touch; spun off as a background suggestion.

Also folds in unrelated pending changes at the dev's request: a gdUnit4 addon `.tscn` re-save (editor format churn) and his own in-editor rune/carving test scratch resources.

### DIFF
--- a/Classes/dev/AttackEditorTool.gd
+++ b/Classes/dev/AttackEditorTool.gd
@@ -8,6 +8,7 @@ class_name AttackEditorTool
 @onready var update_button: Button = %UpdateAttackButton
 @onready var delete_button: Button = %DeleteAttackButton
 @onready var save_as_button: Button = %SaveAsAttackButton
+@onready var status_label: Label = %AttackStatusLabel
 
 # Authors TransmutationData carvings, WeaponAttackData weapon attacks, AND edits an established
 # family's/prototype's MAIN attack in place — three modes, one form (#30 / #72; folded from a
@@ -126,15 +127,19 @@ func _on_update_pressed():
 	if _mode == Mode.FAMILY_MAINS:
 		# The main is edited LIVE, never duplicated, so it already knows its own file.
 		if current.resource_path == "":
-			push_warning("%s's main attack has no saved file yet -- author it in Weapon Attack mode first" % target)
+			var msg := "%s's main attack has no saved file yet -- author it in Weapon Attack mode first" % target
+			push_warning(msg)
+			status_label.text = msg
 			return
-		DevWidgets.save_over(current, current.resource_path)
+		DevWidgets.save_over(current, current.resource_path, status_label)
 		return
 	var path: String = _items[target].resource_path
 	if path == "":
-		push_warning("%s has no file on disk to update" % target)
+		var msg := "%s has no file on disk to update" % target
+		push_warning(msg)
+		status_label.text = msg
 		return
-	if DevWidgets.save_over(current, path):
+	if DevWidgets.save_over(current, path, status_label):
 		_refresh_list(current.display_name)
 
 func _on_delete_pressed():
@@ -143,7 +148,7 @@ func _on_delete_pressed():
 	var target := DevWidgets.selected_name(load_dropdown)
 	if target == "":
 		return
-	if DevWidgets.delete_saved_file(_items[target].resource_path, "attack"):
+	if DevWidgets.delete_saved_file(_items[target].resource_path, "attack", status_label):
 		_refresh_list()
 
 func _on_save_as_pressed():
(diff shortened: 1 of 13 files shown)


Phaazoid:
Second commit on this PR, at the dev's request — the flagged follow-up plus an answer to "is scenario naming safe now?":

**`save_scenario` de-duplicated onto `save_over`** (`b24ead3`). The Law #4 duplicate is gone: `ScenarioManager.save_scenario(name, status_label = null)` delegates the write (dir creation, `take_over_path`, save, error mirror) to `DevWidgets.save_over` and keeps only the empty-name guard + `capture_scenario` + success-only `last_loaded_path`. Both `ScenarioTool` call sites thread their label through, so a genuine scenario **disk** failure now reaches the tool window too — not just a bad name. (Callers counted: exactly two, both in `ScenarioTool`.)

**Scenario naming, remaining hole closed:** with `/` legal for scenarios, `../Foo` built `res://Scenarios/../Foo.tres` — saved *outside* `Scenarios/` where the recursive scan never looks, i.e. the original #168 symptom one level up. `refuse_illegal_name`'s `allow_slash` branch now refuses `..`, `.`, and empty segments (`foo//bar`, trailing `foo/`); honest subfolders (`fixtures/Foo`, `missions/X`) still pass, pinned by test.

**Verification:** three new cases, including one driving the real `save_scenario` into a cross-platform-deterministic save failure (base "directory" is an existing file — zero bytes written on any OS; a backslash name would save *legally* on Linux CI, which is why it isn't the forcer). Falsified by restoring the old inline body — the wire case fails on the empty label. `tests/dev` + `tests/flow`: 177/177.

Issue #174, opened 2026-08-10
Dev authoring for elemental/tile state: paint burning tiles, set unit element states


Phaazoid:
🤖 Claude says:

**Demo milestone — filed off the #137/#138 content push.** `Prolog.tres` is the academy raid, and the raid wants the academy visibly on fire. Investigating turned up that the engine half is further along than assumed — this issue is the **authoring** half, plus one design fork it exposes.

## What already exists (don't rebuild — Law #4)

Burning tiles are built end-to-end *in play*:

- `Terrain.TileState.BURNING` (#50), drawn by `OverlayManager` with the Fire icon
- Occupant damage: `OrderExecutor.apply_burning_tile_damage` — `Terrain.BURNING_TILE_DAMAGE` (5) to any unit standing in fire when **its own faction's** turn ends; `MissionController.check()` runs right after it, so a burning tile can end a mission (#96)
- Duration: `TerrainStateManager.STATE_DURATIONS` — 3 turns, then out
- Serialization: `ScenarioData.terrain_states` round-trips with the scenario (#87)
- The one in-play deposit path: a map-hitting FIRE attack over a TREE tile (`Resources/TerrainReactions/Burning.tres`)

What does **not** exist is any way to author or dev-test this state directly: nothing in `Classes/dev/` references `Elemental.State` or `Terrain.TileState` (grep-verified). The Tile Brush paints static terrain `Kind` and zones only.

## The ask

1. **Tile Brush: a third paint mode for dynamic tile states.** `TileBrushTool.PaintMode { TERRAIN, ZONE }` gains a STATE mode with a picker over `Terrain.TileState` (BURNING / FROZEN / COVER) — left-drag paints, right-click erases, same interaction as the existing modes. Persistence is free (`terrain_states` already saves). This also decouples set-dressing from the reaction table: today fire can only ever appear on TREE tiles, but a burning *building* shouldn't need a tree under it.
2. **Unit element states in the dev editor.** A dev surface to view + toggle `Unit.element_states` (WET, ...), so testing a reaction (soak a unit, hit it with SHOCK) doesn't require authoring a WATER attack first. Battle-scoped state on the transient `Unit` — a dev write, not a new saved field.
[...]

Issue #166, opened 2026-08-10
Transmutation submenu as a catalogue: un-channelable carvings greyed, hover descriptions


Phaazoid:
🤖 Claude says:

## The gap

The Transmutation category lists only what the wielder can currently afford: `RuneData.choice_attacks` routes through the aura-filtered `channelable()`, so a carving you own but can't channel is **absent** rather than greyed — the opposite of the law `MainActionMenu._attack_entry` applies to weapons, where an unfireable pick stays **listed but disabled** (#73/#84). [`docs/design/visual-clarity.md`](../blob/main/docs/design/visual-clarity.md) has recorded this since 2026-07-29 (its "Transmutation submenu should read as a catalogue" item), but no issue carried it — it lived only in the doc line. This ticket is that item, migrated to the tracker per the dev's call at [#157](https://github.com/Phaazoid/Godoiosis/issues/157)'s planning session (2026-08-10).

## Why #157 doesn't remove the need

#157's equip gate is **"at least one channelable carving," never "all"** — so a one-of-three rune equips legally, and the other two carvings still owe the player a reason for not appearing. The gate killed the fully-dead case (both menu rows silently vanishing); the partial case is permanent and only a readout can explain it.

## Scope — the doc item's two halves, one surface

1. **Show un-channelable carvings, blotted out.** The menu reads `inscriptions` (everything inscribed) and asks channelability per entry, rather than being handed the pre-filtered `channelable()` list. Matching the weapons' listed-but-disabled law.
2. **Hover descriptions per carving** — the menu lists carving names with no indication of what any of them does (dev-requested at #88's build).

Also recorded in the doc (worth honoring while in there): the category deliberately opens at a **single** carving because with descriptions and blotted entries it is informative even when it offers no choice — that's the standing rationale for the ≥1 gate in `Unit.has_transmutations()`.

Likely touch points: `MainActionMenu` (the Transmutation submenu build), `RuneData.choice_attacks`/`inscriptions`, and whether `has_transmutations()` should open the row for a rune with carvings none of which channel (today it won't — with #157's gate such a rune can no longer be *equipped*, so the question only arises if the readout is built to show a partial rune's dead entries, which is exactly case 1).

## Related

- [#157](https://github.com/Phaazoid/Godoiosis/issues/157) — the equip gate this complements; its issue text calls this "the complement, and why it is not a rival".
[...]

Issue #135, opened 2026-08-04
Onboarding: nothing on screen explains any system


Phaazoid:
🤖 Claude says:

**Demo milestone.** From the milestone note: *the highest-risk item here.*

## The problem

Nothing on screen explains any system. The list a new player is expected to absorb unaided:

- **Squads** — that units belong to one, that a leader has a cohesion leash, that `has_squad()` means squad*mates*
- **Order queueing** — that you author a plan and then execute it, rather than acting per unit
- **Execution phases** — moves parallel, then attacks, then counters, then side channels
- **Will** — what spends it, what it gates
- **Crisis** — a prompt that appears mid-resolution offering a gamble, with no prior mention
- **Maiming and limb loss** — permanent consequences with no warning that permanence is on the table
- **Transmutation** — aura, affinity, carvings, channeling
- **Readiness economies** — Springspear spring, Carbine magazine, mace charge, chainsword rev, each with different rules

That is a lot of systems, and the demo's whole purpose is judging the combat core. A stranger who cannot form a plan is not evaluating the combat core; they are evaluating an interface.

## The bar

Minimum viable, per the milestone note: **a controls card, and a first mission that introduces one system at a time.**

The mission half is the real work and it overlaps the authored-missions issue — a tutorial mission is an authored mission with a teaching constraint. Worth building them in the same sitting.

## The design fork worth deciding early

**Diegetic tutorial vs. a reference card.** A card is cheap, ages well, and is skippable. A tutorial mission teaches better and costs level-design time you would otherwise spend on missions people actually play. The systems above split unevenly across those two: controls and the queue fit a card; Crisis, maiming, and readiness genuinely need to be *encountered* to land.

Leaning: **card for the verbs, first mission for the consequences.** Not decided.

## Related

- Authored missions (the tutorial is one of them) · **#44** visual clarity · the mission-status HUD issue, which is a prerequisite — a tutorial cannot state an objective the HUD cannot display.

**Milestone:** Demo — a build you can hand to a stranger

— Claude (Opus 5) · 2026-08-04

Phaazoid:
🤖 Claude says:

[...]

PR #181, opened 2026-08-11
Glossary page + registry-backed hover tooltips (#135 reference layer)

Phaazoid:
The reference layer for #135, per the approved in-session plan (scope + exclusions documented in [this comment](https://github.com/Phaazoid/Godoiosis/issues/135#issuecomment-5253616706)). Three commits, each independently green:

## 1 — The seam + the page
- **`Classes/core/Glossary.gd`** — the one term registry: 46 entries, each carrying `short` (the hover tooltip) and `long_text` (the page body). Numbers interpolate the constants that rule them (`DOWN_WILL_COST`, `COVER_DEF`, the new `Unit.DOWNED_TURNS`, …); elemental/terrain **interactions compose from the live reaction catalogs**, never hand-written prose. Bridges from `Stats.Stat` / `Terrain.TileState` / `Elemental.State` mean a new stat or state can't ship without its entry (`tests/law/test_glossary_coverage.gd`).
- **`GlossaryScreen`** — a locking ModalCard, browsable by category, opened from a new pause-menu row and a title-screen button. The pause round trip mirrors REPORT's (close → back to the pause menu, prior-state stash preserved).
- Support: `Elemental.state_display_name` + `Terrain.kind_display_name`/`tile_state_display_name` (the display-name seam's twins), `ACTION_DATA` rows gain their `term` keys.

## 2 — The tooltip surfaces
- **Main action menu**: rows now build through `_entry` (#166's row law finally applied at the top level) with the glossary short as hover readout. Hidden-when-unavailable policy untouched.
- **Stat rows** (inspect panel): tooltips lead with what the stat *means*, then the provenance breakdown when there is one. This deliberately ends the empty-when-unmodified tooltip — the emptiness was an anti-noise call about provenance alone. The provenance builders and their tests are untouched.
- **State icons** (hover card + inspect states bar): each state explains itself on hover — the unit half of visual-clarity's P6 doctrine.
- **Tile hover block**: the hover card is now a stack — any hovered cell that is *anything other than ordinary ground* (tile states with live clocks via the new `TerrainStateManager.turns_remaining`, water's traversal gate, above-norm move cost) grows a readout block, with or without a unit standing there. Plain grass shows nothing.

## 3 — Canon sweep
visual-clarity.md P6 progress note + marker bump; CLAUDE.md core/ + ui/ rosters.

## Verification
- New suites: `tests/law/test_glossary_coverage.gd` (8), `tests/ui/test_glossary_screen.gd` (3), `tests/ui/test_tile_hover_readout.gd` (4), plus a main-menu readout case in `test_menu_catalogue_rows.gd`.
- **Every new test falsified against its own bug** — blanked registry entry, deleted pause row, dropped pause reopen, unclaimed modal lock, dead tile composition, unrouted main menu, card-shown-on-empty: each went red before green.
- Full `ui` + `law` + `core` areas: **212/212**.
- **Not verified: feel.** Glossary layout, tile-block position/width (fixed 160px, wraps long lines), and tooltip lengths need in-game eyes — that's the playtest half of this handoff. The Prolog's authored academy fire (#174) is a ready-made board for the tile hover.


Phaazoid:
🤖 Claude says:

**Playtest round 2 pushed** (`c5c32d7`), per the in-session plan:

- **Tooltip delay** — global 0.5 → 1.0s (`project.godot`). Everything with a built-in tooltip waits the extra beat now, dev tools included — feel that deliberately, since it was accepted as a named consequence.
- **Attack targeting** — every attack row's readout ends with `(unit)` / `(tile)` / `(unit/tile)` (one spelling, `AttackData.targets_text`; the rune inventory tooltip inherits it), and the Glossary's Actions page gains an **Attack Targeting** entry explaining the axis.
- **Tile card** — now a real card (the unit card's own stylebox), on **every** tile: kind icon + name header, states with live clocks worked back in, ground rules, and which elements can touch the tile. That last list runs through `TerrainReaction.applies_to_tile`, extracted from the resolver's inline deposit gate so the card and the resolver share one predicate.

Verification: ui + law + terrain 225/225; the flipped every-tile case, the targeting token and the interactions list each falsified red before green.

**To feel-check:** the 1s delay across surfaces, the every-tile card's presence while sweeping the mouse (it's a lot more card than before — by request, but worth confirming it reads as information rather than noise), the 160px interaction lines ("Ice → sets Frozen"), and the parenthetical wording.

— Claude (Fable 5) · 2026-08-11

Issue #177, opened 2026-08-11
Cast characters: reference-first UnitData, starting kit, catalog spawn


Phaazoid:
🤖 Claude says:

**Demo milestone — the roster lane's first ticket, plan approved in-session 2026-08-10.** Dev decision: **the cast saves as references; snapshots return when mission-to-mission carryover arrives.** Today "Torv" cannot exist as one authoritative thing — `UnitData` can't express a kit, the spawner can't place a saved character file, and every tool save embeds an inline copy of the unit, so each authored mission would hold its own drifting Torv.

## Scope (approved plan)

1. **`UnitData` starting-kit block** — `starting_jobs` / `starting_inventory` (refs to standalone equippable `.tres`, granted via `copy_equippable`) / `starting_equipped_index` / `starting_worn_index` / `starting_proficiency` / `starting_prosthetics`. All defaults inert.
2. **`Unit._seed_starting_kit()`** in `_ready` after `initialize()` — grants through the gated doors, warns on refusal.
3. **Provenance**: `UnitFactory.create_unit` stashes `unit.unit_data_source` when the handed `UnitData` is a standalone file (`resource_path` non-empty and `::`-free — a scenario-embedded sub-resource has a `::` path and must NOT count).
4. **`ScenarioUnitEntry.state_saved := true`** — the `affinity_saved` pattern widened to the whole state block; default true is back-compat (existing saves are snapshots). Reference entries write false and are never `apply_unit_state`'d. Snapshot apply gains a clear-before-append so it stays authoritative over a seeded kit.
5. **`ScenarioManager.capture_scenario(name, authored := false)`** — authored saves write cast units as `ExtResource` references with no state; `BugReporter` keeps the default and stays a byte-exact #87 snapshot. Loader passes `entry.unit_data` directly (the outer `duplicate(true)` severed provenance) and gates apply on `state_saved`.
6. **`UnitCatalog`** (scan of `Resources/Units/`) + a Character dropdown on the SpawnTool.
7. **Scenario tab checkbox** (default ON): authored save vs exact mid-battle snapshot.

## Tests

Kit seeding (copies, not shared), the reference round-trip (edit the character file → the mission sees it), snapshot-not-doubled (falsified against the missing clear), back-compat load of pre-change saves, and the `::` provenance boundary.

Full plan + rationale preserved in the session plan file; canon sweep (CLAUDE.md persistence seam, missions.md, stats.md, jobs.md + the #70 aside correction) rides the closing PR.

[...]

PR #185, opened 2026-08-11
Character editor: author the Resources/Units/ cast files in the dev tools (#179)

### LINKED ISSUE #179: Character editor: edit and create the Resources/Units/ cast files in the dev tools
🤖 Claude says:

**Demo milestone — the roster lane's authoring tool, requested by the dev 2026-08-10 right after #177/#178.** The character files that #177 made load-bearing cannot be touched by any dev tool.

## The gap, verified

- `UnitEditorTool` edits the **live selected unit** only — every control writes a staged buffer applied to the unit; nothing in the file writes a `.tres` (grep-verified: no `ResourceSaver`/`save_over` call). Its edits used to reach disk indirectly via scenario snapshots; with #177's reference semantics, an authored-save cast unit's live edits deliberately reach **nothing**.
- The `SpawnTool` form builds ad-hoc `UnitData` and (since #177) *places* character files, but never edits one.
- The #177 **starting-kit block** (`starting_jobs` / `starting_inventory` / equip+wear picks / `starting_proficiency` / `starting_prosthetics`) has **no dev surface at all** — Godot-inspector only.

Since a reference save means *the file IS the character*, file editing is character editing. This is the tool that makes the #137 balance pass (author the Bleeding Hearts with real numbers) a dev-tools workflow instead of inspector surgery.

## The ask — the Attack Editor's shape, applied to characters

Model named by the dev: **the weapon-family / main-attack editing modes** (Attack Editor: pool authoring's new/load/name/save + FAMILY_MAINS' load → edit in place → overwrite). A Character editor tab (or a mode on the existing Unit tab) with:

1. **Load** — dropdown over `UnitCatalog` (`Resources/Units/`).
2. **Edit** — identity (display name, base stats, faction, sprites, portrait, base aura, affinity, innate abilities) **and** the kit: jobs from `JobCatalog`, starting inventory **picked from the weapon/armor/rune catalogs** (which makes the hand-inlined base-class-`WeaponInstance` trap — the one `_seed_starting_kit` warns about — unhittable by construction), equipped/worn picks, proficiency, prosthetic slot assignments.
3. **Update** — overwrite the loaded file in place. Same heavier-handed semantics as FAMILY_MAINS, and now with the same justification: the edit changes that character **in every mission that references them**, which is exactly what #177 was built to mean.
4. **Save As** — create a new character file, from scratch or **from the live selected unit** (`UnitEditorTool` already stages one — "save this unit as a character file" is the cheap create bridge).

## Sharp edges this must ride (all documented, all already bitten once)

- `take_over_path` on every save site — whoever writes a path must claim it (#99).
- `OptionButton.select(-1)` before any dropdown rebuild's re-select loop.
- `DevWidgets.save_over` + a status label, per #168's visible-failure rule.
- **The SpawnTool's Character dropdown is built once in `init`** — a character save must refresh it (JobCatalog's `refresh()`-after-runtime-authoring precedent), or a freshly authored character is invisible until restart.

## Related

#177 / PR #178 (the reference semantics that make this the character editor) · #137 (the balance pass it unblocks) · #99 (Save As / Update / Load verbs) · #168 (save feedback)

— Claude (Fable 5) · 2026-08-10

Phaazoid:
Closes #179.

## What this is

The Character tab in the dev tools — the authoring surface for `Resources/Units/` cast files (#177 made them load-bearing; nothing could edit one). Three commits:

1. **Plumbing** — `AbilityCatalog` (first enumerator of `Resources/Abilities/`), SpawnTool's Character row rebuilt inline with a stored dropdown + `refresh_characters()` fired on Spawn-tab entry, sprite-triple scan hoisted to a shared static.
2. **The tab** — Item Editor pool chrome (load-gated Update, Save As refusals, confirm-on-delete, visible status label) over a staged `UnitData`. Every identity + kit field gets a control: stats, faction, sprites/portrait, affinity/aura (Rebecca-rule guarded), alkahest, innate abilities, jobs, 6 kit slots picked **from the catalogs** (slots stage the catalog file itself, so saves write ExtResource references — the inlined-`WeaponInstance` trap is unhittable from here), equip/wear picks, prosthetics (gated by `can_install_as_prosthetic`), proficiency (absent = 3). **Capture selected unit** projects the live Unit-Editor unit through `ScenarioUnitEntry.capture_unit_state` onto the kit block; the four unrepresentable losses (inline items, auto-equip on captured-unarmed, limb states, >6 items) are all said in the status label, never silent.
3. **Tests** — 9 cases, no disk writes, all four guard falsifications proven red (details in the issue comment).

## Verification

- `tests/dev` 63/63; **full suite 1186/1186, 0 failures** (178.9s Elapsed).
- Falsified: Update gate removed, New-reset removed, aura-erase removed, refresh wire cut — each turns exactly its case red. The first falsification run caught a weak assertion (a gate-less Update still writes a "no map sprite" note into the label) and overwrote `Aster.tres` in the process — restored from git, assertion sharpened to the refusal wording.

## Playtest checklist (before merge)

- [ ] New → pick sprite/portrait → stats/kit from catalogs → Save As → new character appears in the Spawn tab's Character dropdown after switching tabs → spawn it, kit seeds
- [ ] Load Aster → tweak → Update → respawn shows the change (also: the load flags Aster's out-of-affinity aura — fixing it here is one checkbox + Update)
- [ ] Capture a live unit → read the inline-item notes → Save As
- [ ] Feel check: tab order (Character sits after Unit Editor), form section order, the red status label carrying informational capture notes

One cosmetic diff to expect: re-saving a cast file drops the editor-written `metadata/_custom_type_script` line (code-path `ResourceSaver` doesn't emit it — harmless).

### DIFF
--- a/Classes/dev/CharacterEditorTool.gd
+++ b/Classes/dev/CharacterEditorTool.gd
@@ -1,0 +1,557 @@
+extends MarginContainer
+class_name CharacterEditorTool
+
+# Dev-only character editor (#179): authors the standalone UnitData files under Resources/Units/
+# -- the cast (#177). An authored save REFERENCES those files, so Update rewrites the character
+# in every mission that references it; the file IS the character. Load/Update/Delete/Save As
+# ride the Item Editor's pool chrome; Capture stages the live unit selected in the Unit Editor
+# tab as a new character. Kit slots stage catalog FILE resources, so saves write ExtResource
+# references (UnitData's own authoring guidance) and a path-less entry is called out as inline
+# rather than silently embedded.
+
+const KEEP_LABEL := "(keep current)"
+const PORTRAIT_DIR := "res://Art/Units/Portraits/"
+
+@onready var editor_container := %CharacterEditorVbox
+@onready var load_dropdown: OptionButton = %CharacterLoadDropdown
+@onready var name_input: LineEdit = %CharacterNameInput
+@onready var update_button: Button = %UpdateCharacterButton
+@onready var delete_button: Button = %DeleteCharacterButton
+@onready var status_label: Label = %CharacterStatusLabel
+
+var current: UnitData = null
+var _items := {}
+# Which catalog entry current was loaded from ("" = a New character). Load stages a COPY with no
+# resource_path, so nothing else records this -- and Update's load-gate needs it.
+var _loaded_name := ""
+
+func _ready():
+	_refresh_list()
+	_on_new_pressed()
+
+func _refresh_list(select_name := ""):
+	load_dropdown.clear()
+	_items = UnitCatalog.get_characters()
+	for k in _items:
+		load_dropdown.add_item(k)
+
+	# add_item auto-selects index 0 -- Update must never aim at an entry nobody picked.
+	load_dropdown.select(-1)
+	for i in load_dropdown.item_count:
+		if load_dropdown.get_item_text(i) == select_name:
+			load_dropdown.select(i)
+			break
+	_refresh_buttons()
+
+func _refresh_buttons():
+	DevWidgets.refresh_update_button(update_button, DevWidgets.selected_name(load_dropdown), "character", _update_block_reason())
+	DevWidgets.refresh_delete_button(delete_button, DevWidgets.selected_name(load_dropdown), "character")
+
+# "" = allowed. Update only overwrites the character that is actually LOADED (#166 shape) -- and
+# here an overwrite reaches every mission referencing the file, the blast radius that made the
+# gate a rule in the first place.
+func _update_block_reason() -> String:
+	var target := DevWidgets.selected_name(load_dropdown)
+	if target == "" or target == _loaded_name:
+		return ""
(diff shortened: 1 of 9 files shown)


Phaazoid:
🤖 Claude says:

**Review feedback folded in (commit 4): Character is now a sub-tab of a renamed Unit Authoring tab beside Spawn** — top level is back to six tabs. The Unit Editor's sub-tab precedent, but scene-built since these pages are static.

One design point worth knowing when you playtest: the Spawn-dropdown refresh now has **two doors**, because the natural authoring loop — save a character, flip to the Spawn *sub*-tab, place it — never fires an outer tab change. Both the outer tab entry and the inner sub-tab switch refresh the pickers through one `_refresh_spawn_pickers()`, and the suite drives each door through its real hook (both falsified: each cut wire turns exactly its own case red).

Full suite **1187/1187** (181.3s). Checklist addendum: the earlier "appears in the Spawn tab after switching tabs" step now means switching *sub*-tabs within Unit Authoring — that's the new wire to feel out.

— Claude (Fable 5) · 2026-08-11

Issue #50, opened 2026-06-27
Map elemental interactions


Phaazoid:
We'll want to start adding in some basic elemental interactions with the map tiles.  

Things like bush + fire = burning

Water + ice = frozen

etc.  

This will tie in with #47 , since we will apply elemental effects this way.  This will also tie in with #30 design session.  

We'll need to put another toggle on attacks, as well.  Some attacks will hit just the enemy they target with elemental effects, some will hit the enemy and the map.  Some may only hit the map, even if there's an enemy there.  

I've closed #14  but its third and final point may come into play, depending on how we implement things here.

Phaazoid:
🤖 **Claude says:** Status checkpoint. The #30 rune-firing arc — the dependency this issue needed — has landed (committed, suite green at 116 cases). MAP-targeting attacks now actually exist, so terrain deposits have a real source. Here's the done / left split.

## Done — slices 1 & 2

**Slice 1 — headless plumbing**
- **Separate `Terrain` vocab** (kept independent of `Elemental.State`, dev call): `Terrain.TileState {NONE, BURNING}` (dynamic per-cell condition, serializes as int, append-only) + `Terrain.Kind {NONE, GRASS, MUD, ROCK, TREE, WATER}` derived from the tileset's `terrain_type` string via `_KIND_BY_NAME`.
- **`TerrainStateManager`** store — `Dictionary[Vector2i, Array[TileState]]` with `apply` / `states_at` / `has_state` / `cells_with`.
- **Cell-effect channel** — `ResolvedPlan.cell_effects: Array[ResolvedCellEffect]`, populated by `PlanResolver._resolve_cell_effect` **only when a board is supplied** (unit-only resolver callers stay untouched).
- **First reaction** — `FIRE × tree → BURNING` (`TerrainReaction` `.tres` + `TerrainReactionCatalog`).
- **The attack toggle this issue asked for** — `TargetMode {UNIT, MAP, BOTH}` on `WeaponData` and `TransmutationData` (default UNIT), gating the deposit. Covers "hit the unit / hit the map / hit both."

**Slice 2 — live & previewed (Law #2)**
- Execution applies effects (`game._apply_cell_effects`), and the action queue **previews** the pending ignite as a ghost marker before it fires (`OverlayManager` terrain sprites, `TerrainIcons/Fire.png`).
- `SquadManager.resolve_plan` threads the board + `TerrainReactionCatalog` so preview == execution.
[...]

PR #199, opened 2026-08-12
FROZEN is permanent (playtest call)

Phaazoid:
Playtest feedback, reported in-session: Ice shouldn't be on a timer. **Stacked on #198**; the stack is #195 → #196 → #197 → #198 → this.

## What changed

`Terrain.TileState.FROZEN` carried a 3-turn `STATE_DURATIONS` entry, same clock as BURNING. Removed — FROZEN now follows **COVER's exact mechanism** (already the established pattern for a "permanent unless destroyed" tile state): no entry means `tick_states` never touches it, and it's removed only by a destructive `states_removed` deposit. That matches the design intent already on record in `marketing.md` — *"freeze the river, march your squad across, melt it behind them"* — where the player decides when the ice goes away, not a clock.

**One thing surfaced while tracing this that's worth flagging separately:** `docs/design/terrain.md`'s #50 banner claims slice 3 shipped "both authored `.tres` reactions" for ICE→FROZEN and FIRE→FROZEN. That's not what's in the repo — `Resources/TerrainReactions/` doesn't exist; those two reactions live only inline in `tests/terrain/test_ice.gd`, proving the machinery works but never wired to a real carving/weapon. Practical effect right now: nothing in live play removes FROZEN at all except the dev brush's right-click erase — "melt it behind them" isn't actually playable yet. Left a note in the doc; didn't fix it here since it's a different, bigger question (author the reaction) than what was asked.

## Changes

- `Classes/terrain/TerrainStateManager.gd` — the one-line mechanical fix (drop the entry).
- `Classes/terrain/Terrain.gd`, `Classes/core/Glossary.gd` — enum comment + player-facing tooltip text, dropped the "for N turns"/"until it thaws" wording.
- `docs/design/terrain.md` — dated banner note + the stale reaction-authoring flag above.
- `tests/terrain/test_ice.gd` — the melt-after-duration case replaced with a permanence case (ticks 20 rounds, asserts it holds), mirroring `test_burrow.gd`'s existing Cover-permanence case.

## Verification

- Full suite **1269/1269**.
- The new case falsified against the old behavior: reintroduced the duration entry as a mutant, confirmed it goes red, reverted.
- **Your half:** playtest the actual feel — is permanent-until-removed what you wanted, given nothing removes it yet in live play but the brush?

Includes your latest Level_1.tres repaint progress.

## Version bump — pending one commit

Owes a `0.6.1 → 0.6.2` bump (bugfix-only). Waits on Godot being closed, same as the last two PRs.


Phaazoid:
🤖 Claude says:

**Retraction for this PR's "nothing melts ice" flag** (found while building #201, which stacks on this branch): the finding is **false**. `Resources/TerrainReactions/{Burning,Frozen,Melt}.tres` have been authored and tracked since 2026-07-01 (`41db51c`), `SquadManager.resolve_plan` feeds `TerrainReactionCatalog.get_all()` into every live resolve, and `Fireball.tres` carries FIRE at `TargetMode.BOTH` — so fire has melted painted ice in play all along (the Glossary's FROZEN entry even documents it). The misread came from `tests/terrain/test_ice.gd`, whose reactions are deliberately built inline; the authored files live beside them and were missed by a bad directory listing.

#201 fixes the `Terrain.gd` comment and the `terrain.md` note that carried this; the flag in **this PR's body** is yours to amend. The frozen-permanent change itself is untouched by the correction — permanent ice plus a working melt is exactly the "freeze, cross, melt it behind you" loop.

— Claude (Fable 5) · 2026-08-12

Issue #220, opened 2026-08-13
HD-2D Stage 4b: the input bridge - play the real game in 3D


Phaazoid:
🤖 Claude says:

**Stage 4b of the #176 HD-2D overhaul: the input bridge.** Stage 4a (#215, PR #216) made the battle *watchable* in 3D — `Battle3D.tscn` runs the real `Main.tscn` hidden, `BoardMirror`/`UnitMirror` render it read-only. 4b makes it **playable**: clicks and hover in the 3D view route into the real game, through the same doors the 2D presentation uses.

## Scope

1. **The click seam** — extract a cell-level entry from `game.gd`'s `_unhandled_input` (shape: `handle_board_click(cell: Vector2i)`) so both presentations feed one door. Battle3D resolves the click via `BoardPicker` (boards are flat — drop y) and calls it. Every caller/reader of the input path gets counted at plan time (the Group Move lesson).
2. **The hover seam** — `HoverPresenter`'s mouse→cell read becomes injectable, so `hovered_cell_changed` / the hover card flow from the 3D picker exactly as they do from the 2D mouse.
3. **UI hosting — DECIDED (dev, 2026-08-13): 2D corner picture-in-picture for 4b.** The hidden `GameContainer` re-shows small in a corner: every menu/panel/queue usable on day one, near-zero throwaway code, and the dev sees the 3D board + real UI together before 4c commits to its anchoring. A bridge, not the shipping look — 4c owns real UI anchoring over the 3D view.
4. **Driver changes** — the player faction stops being AI; `auto_play` becomes explicit demo-mode semantics. (This is also where the 4a feel-check note lands: "the 4-player team are supposed to squad up" — squadding becomes the player's own opening move again.)
5. **The mission-entry seam** — 4a's driver calls `_on_mission_chosen` directly, a declared demo liberty; 4b formalizes it.

Overlay polish from the #176 feedback batch (z-fight candidates, overlays-go-unshaded) rides 4b or 4c — decided at plan time.

Formal Plan Mode before any edit (this stage touches `game.gd`).

— Claude (Fable 5) · 2026-08-13

Phaazoid:
🤖 Claude says:

[...]

Issue #176, opened 2026-08-11
HD-2D presentation overhaul: 3D diorama board, billboarded pixel sprites, lighting + post stack (Octopath / Triangle Strategy style)


Phaazoid:
🤖 Claude says:

## What this is

Adopt the **HD-2D presentation style** (Octopath Traveler / Triangle Strategy) for Iosis battle scenes: real 3D diorama environments skinned in pixel art, 2D character sprites billboarded inside that world, dynamic 3D lighting, and a heavy post-processing stack. Parked far-future deliberately. The dev confirmed (2026-08-10) that **all current presentation is placeholder and will be overhauled wholesale**, so when this wakes up it is a *greenfield build of the real presentation layer*, not a port of the current 2D one — the migration cost that would normally dominate this ticket does not exist. Everything below is the distilled 2026-08-10 conversation, written so nothing has to be re-derived later.

## Why it fits Iosis specifically

- **Verticality finally gets a home.** Combat doctrine already treats height as core — #116: shoving a unit off a height is an outright kill — but the current board is flat and elevation has zero representation in code. A 3D board renders elevation naturally instead of faking it with 2D tricks.
- **Relighting is free content.** The same map assets become night / rain / interior scenes purely via lighting + grading — one of HD-2D's best tricks, and it only works if the art stays flat-lit (see conventions below).
- **The rules engine doesn't care.** Combat logic and the test suite are presentation-agnostic; they plug into whatever draws the board. This overhaul threatens no game-rules code.
- **The Aseprite pipeline carries over unchanged.** Same PNGs, mounted on billboarded 3D quads instead of `Sprite2D`.

## What HD-2D actually is (four ingredients)

1. **3D diorama world with pixel-art skin** — low-poly tile geometry textured with low-res pixel art, nearest-neighbor filtered. Source of the depth, parallax, and "model railroad" feel.
2. **Billboarded 2D sprites** — flat quads that stay upright and face the camera; they *participate in 3D lighting*, casting shadows onto terrain and receiving light color.
3. **Real-time 3D lighting** — one directional sun with shadows, plus warm point lights (torches, lanterns, spell glow). Art is flat-shaded in the texture, dynamically lit in the scene.
[...]

Issue #131, opened 2026-08-04
The feedback loop: getting a bug report off a stranger's machine


Phaazoid:
🤖 Claude says:

**Demo milestone · sub-issue of the export.** Everything that carries information from a stranger's machine back to you.

A demo you cannot get feedback out of is a demo that taught you nothing. This group is the milestone's actual purpose — the rest of the packaging just makes the game reachable.

## 1. The bug reporter is dead in an exported build

`Classes/dev/BugReporter.gd`:

```gdscript
const REPORT_DIR := "res://bug-reports/"
```

`res://` is read-only once exported — CLAUDE.md already flags this (*"note `res://` is only writable in an editor run"*), and #128 shipped against that constraint knowingly. Moving to `user://` is the fix, plus **telling the player where the file landed**: nobody finds `%APPDATA%\Godot\app_userdata\...` unaided. An OS-native "open the folder" affordance is the cheap version.

Watch the two paths that must NOT follow it: reports still have to stay outside `Scenarios/` so Mission Select and #9's folder scan never see them, and `bug-reports/.gdignore` exists to stop the editor importing screenshots.

## 2. A visible affordance, not just F3

F3 is correct for you and invisible to a stranger. Needs a button they can find and **on-screen confirmation that a report was written** — a silent success reads as a broken feature and they stop using it.

## 3. Error visibility

`push_error`/`push_warning` go nowhere a player can see. The bug reporter's log tail is already the mitigation, which is a second reason item 1 is load-bearing rather than cosmetic. Folded in here rather than filed separately — it has no fix of its own.

## 4. The intake channel — a decision, not a task

How does a report physically get from their machine to you? Manual upload to a form, an itch.io page, a Discord, or automatic upload. **The rest of this issue's shape depends on this answer**, so decide it before building the affordance. Automatic upload is the only option that pulls in item 5.

## 5. Consent, if anything uploads automatically

A report carries a screenshot and a full board snapshot. If it leaves their machine without explicit opt-in, that is data collection and it needs disclosing up front. If the intake channel is manual, this item evaporates — which is a real argument for manual on a first demo.

## 6. "What did you think" is a different question from "what broke"

Bug reports capture defects. They do not capture *"I was bored"*, *"I had no idea what to do"*, or *"the shove was the best part"* — which is most of what a demo is for. Needs its own prompt, and it should not ride the bug-report path.

## Related

[...]

Issue #103, opened 2026-07-28
AI squad soft-locks the game when its plan is invalid: ghosts persist, squad never acts, turns loop forever


Phaazoid:
🤖 Claude says:

Filed from `docs/SCRATCHPAD.md`. Reported by @Phaazoid: in `Scenarios/Castle Assault.tres`, with both player squads set to **AI/Rushdown** and the enemy squad to **AI/Hold**, the game played itself into a permanent stuck state once the enemy leader was the last hostile standing.

## Observed (dev, 2026-07-28)

> the player AI in rushdown somehow ended their turn in a state where they were planning an invalid move, so when it shifted to the enemy turn, the ghost movement projection and arrows were still on the map. At this point, the enemy AI would attack one of the ghost states which did nothing, then pass. The player would then not be able to do anything, stay in that state, and pass, and it went on forever.

The enemy leader was still **active** (never downed) throughout.

## Root cause: the invalid-plan guard is a player affordance with no AI counterpart

`OrderExecutor.execute_orders` (`Classes/actions/OrderExecutor.gd:37`):

```gdscript
if game.squad_manager.squad_has_invalid_actions(squad):
	for action in squad.action_queue:
		if not action.is_valid:
			action.actor.visuals.play_invalid_flash()
	return
```

For a human this is exactly right — flash the bad rows red, hand control back, let them fix the plan and press Execute again. The early `return` is *before* three things that only ever run on the success path:

- `action.actor.visuals.set_projected(false)` (line 53) — so the **projection ghosts stay on the board**
- the `remove_action` loop (line 76) — so the **invalid orders stay queued**
- `set_has_acted(squad, true)` + `clear_planned_path` (lines 79-81) — so the **squad is never marked as having acted and the path arrows stay drawn**

`AIController.take_faction_turn` (`Classes/ai/AIController.gd:42`) calls the archetype and then `await execute_orders(...)` with no return value and no check. **The AI cannot see a red flash.** So it plans, gets silently refused, and moves on — leaving every piece of that state behind. Each observed symptom falls straight out:

| Symptom | Mechanism |
|---|---|
| Ghosts + arrows persist into the enemy turn | `set_projected(false)` and `clear_planned_path` are below the early return |
| Enemy AI swings at a ghost and does nothing | AI target queries read `get_projected_destination()`, which still points at the stale planned cell |
[...]

PR #221, opened 2026-08-13
HD-2D Stage 4b: the input bridge - play the real game in 3D (#220, v0.13.0)

Phaazoid:
Closes nothing on merge — **#220 closes on your feel-check**, per pattern.

## What this is

Stage 4b of the #176 HD-2D overhaul: **the battle is now playable in 3D.** Clicking and hovering in the Battle3D view drive the real game; the 2D UI lives on as a corner picture-in-picture (your ruling); the player faction is human again — which is where your 4a note lands: squadding up the Prolog four is the player's own opening move now.

## The mechanism (recorded on #220 before building)

**Synthetic input, not method calls.** The bridge maps 3D pointer → cell (`BoardPicker`, its first non-LookDev consumer) → the cell's position in GameView coordinates → the PiP's screen coordinates, and feeds ordinary mouse events to `Input.parse_input_event` — the same entry physical input uses. The container forwards them into the hidden game; `game.gd:192` stays the only mouse→cell site; **`game.gd` has zero edits**; the board lock, the modal freeze, menu placement, and hover all fire natively. Hover works from both surfaces **for cells the hidden 2D camera is showing** — a far cell's mapped position falls outside the container rect and is dropped, so the PiP's hover card can lag on the last near cell while the 3D bracket stays correct (self-review finding; options for closing it are on the review comment, your call).

Three engine facts were measured on the way (probe scripts, deleted after):

1. **`Viewport.push_input` is a dead door against `handle_input_locally = false`** — pushed events into GameView never process. The container is the only live entry.
2. **Unconsumed events BUBBLE back out of the SubViewport to the root** — a PiP board click reaches Battle3D's `_unhandled_input` too. Hence the PiP-rect guard, and the pinned rule that the bridge never re-reads such events as 3D pointing.
3. **A synthetic the bridge re-reads as 3D pointing forges a self-sustaining event loop that hangs the flush** (found when a falsification mutant froze the suite twice). Every synthetic now carries a meta tag; the tag check is the loop-breaker, not hygiene.

**The camera snap**: a pushed click's position must map back to the intended cell through the live 2D camera transform, so `CameraController.snap_to_position` (instant + clamped, built from `_apply_pan_position`) fires before each mapped click. Structural consequence, by design: **the PiP recenters on every 3D click** — it tracks your action. (Discovery en route: `ActionMenuController._place_panel` already clamps menus on-screen, so the snap's payload is placement coherence + PiP tracking, not menu reachability.)

**Driver**: `demo_mode` (new export, default off) preserves the 4a AI-vs-AI diorama; `auto_play` keeps meaning "start the mission on launch". Playable mode leaves `ScenarioData.ai_factions` standing — all four `is_ai_faction` readers (board lock, End Turn visibility, the #103 concede, pacing) flip to the human path at once.

## Knobs (the tuning rule)

- `pip_scale` (default **0.35**) — PiP display size; the container keeps its native 1280×720 SIZE (GameView keeps full resolution; `custom_minimum_size` guards this too) and only the display scales.
- `pip_margin` (default 12,12) — corner inset, bottom-right.
- The hover bracket reuses `BoardOverlays`' existing `bracket_*` exports.

## Verification (honest report)

`tests/presentation/test_input_bridge.gd` — 10 cases, the repo's first true click-through wire tests: events enter at the top of the input pipeline (`parse_input_event`) and must survive the whole chain. Full suite **1358/1358, 0 orphans** (149.8s).

Every mechanism falsified (mutant in → red → restore), 11 mutants:

| Mutant | Case that went red |
|---|---|
| pick maps (x,y) not (x,z) | click wire |
| camera snap removed | snap equality (selection stayed green — the math is position-robust; the snap is the camera contract) |
| `can_process()` gate dropped | modal freeze — **after two blind clauses were caught and replaced**: end-state asserts miss a pierced click grazing the pause card's button gaps, and the viewport-mouse comparison is blind because the snap maps every click to the exact center a fresh SubViewport already reports. The clause that works: the bridge's pointer state stays virgin |
| `game.gd:188` board lock disabled | board lock (the sharp clause is RMB — for LMB the dispatch `match` is a second guard) |
| hover motion dropped | hover (presenter clause) |
| hover bracket dropped | hover (bracket clause) |
| PiP left hidden | suite-wide from case 1 — **an invisible container forwards nothing; PiP visibility is load-bearing for the whole bridge**, not just the UI |
| naive container shrink | native-resolution — only after ALSO zeroing `custom_minimum_size`: the authored minimum is a second guard that makes the naive bug a no-op on its own |
| RMB branch dropped | RMB cancel |
| unconditional both-AI | driver |
| PiP-rect guard removed | PiP board click (unconsumed PiP clicks would double-act as 3D picks) |

**What headless cannot see — your half:**
- PiP readability and feel at 0.35 scale (knob), and whether bottom-right is the corner you want.
- The click-snap camera behavior as a feel (every 3D click recenters the PiP).
- Real-mouse 3D picking feel; menus opening inside the PiP.
- **WASD pans BOTH cameras** (the rig polls physical W/S/A/D, the 2D camera polls `cam_*` bound to the same keys) — declared 4b jank, self-healing on the next click; the 4d camera pass owns real arbitration.
- Esc pauses and the report card works from 3D (the modal claims ALWAYS and the PiP keeps the click path alive — the #131 architecture, untouched); typing WASD into the card no longer pans the rig (freeze-mirror, `_process`).

Deliberately not here (per plan): overlay mirroring beyond the hover bracket, z-fight fixes, unshaded conversion (4c); camera arbitration + chooser + hidden-viewport power (4d); dev F-keys in Battle3D (4d's compatibility check). Pre-existing and unchanged: the Prolog SteamRune load noise flagged on #216.

### DIFF
--- a/Classes/board/CameraController.gd
+++ b/Classes/board/CameraController.gd
@@ -1,3 +1,7 @@
+# The 2D board camera: WASD-scrolled with grid snapping, clamped to the board,
+# with AI-turn locks (set_ai_locked/follow) and the fixed-duration pan_to beat.
+# center_on_position glides via the _process lerp; snap_to_position is the instant
+# form (the 3D input bridge maps clicks through the live transform, #220).
 extends Node2D
 class_name CameraController
 
@@ -41,6 +45,15 @@ func center_on_position(world_pos: Vector2):
 	lock_manual_input = true
 	target_position = world_pos
 	clamp_target_position()
+
+# Instant, clamped reposition. The 3D input bridge (#220) maps a click's viewport
+# position through the LIVE canvas transform, so the camera must already be showing
+# the clicked cell when the synthetic event lands — a lerp target isn't enough.
+func snap_to_position(world_pos: Vector2) -> void:
+	target_position = world_pos
+	clamp_target_position()
+	_apply_pan_position(target_position)
+	camera.force_update_scroll()
 
 func clamp_target_position():
 	var viewport_size = get_viewport_rect().size
--- a/Classes/flow/MissionController.gd
+++ b/Classes/flow/MissionController.gd
@@ -85,7 +85,7 @@ func open_mission_select() -> void:
 		if not missions.has(path):
 			others.append(path)   # root playtest saves + fixtures/ -- selectable during development
 	_select_screen = MissionSelectScreen.open(game, missions, others)
-	_select_screen.mission_chosen.connect(_on_mission_chosen)
+	_select_screen.mission_chosen.connect(begin_mission)
 	_select_screen.load_game_chosen.connect(_on_load_game_chosen)
 	_select_screen.sandbox_chosen.connect(_on_sandbox_chosen)
 	_select_screen.glossary_chosen.connect(func(): GlossaryScreen.show_screen(game))
@@ -98,7 +98,9 @@ func _close_mission_select() -> void:
 		_select_screen.queue_free()
 	_select_screen = null
 
-func _on_mission_chosen(path: String) -> void:
+# The one path-taking mission entry, public since #220 — the Mission Select signal
+# and the Battle3D driver both start missions through this door.
+func begin_mission(path: String) -> void:
 	_close_mission_select()
 	game.scenario_manager.load_scenario(path)   # routes through clear_board() -> reset()
 	_begin_turn()
@@ -134,7 +136,7 @@ func restart_mission() -> void:
 	game.scenario_manager.reload_current()
 	_begin_turn()
 
-# Player-facing resume (#144): the same two-step arrival _on_mission_chosen makes, except the
+# Player-facing resume (#144): the same two-step arrival begin_mission makes, except the
 # board comes from a save slot and last_loaded_path is aimed at the ORIGIN mission -- so Restart
 # and F2 reload the mission start, and no dev tool can ever aim Update at a slot. _begin_turn no
 # longer resets actions (a menu arrival trusts the file -- see game._on_turn_started), which is
--- a/Scenes/Battle3D/battle3d.gd
+++ b/Scenes/Battle3D/battle3d.gd
@@ -22,10 +22,10 @@ extends Node3D
 # PiP knobs (aesthetics get a knob, not a guess):
 @export var pip_scale := 0.35
 @export var pip_margin := Vector2(12.0, 12.0)
-
-# The 2D game's native resolution (Main.tscn authors GameView and the container's
-# minimum size to this); the PiP pins the container SIZE here and scales the display.
-const PIP_NATIVE := Vector2(1280.0, 720.0)
+# Bring-up diagnostic (4b): the bridge narrates each link on the help label —
+# pointer pick, click mapping, refusals — so a live run pinpoints which link is
+# dead without a debugger. Untick to silence; drop the mechanism at 4c.
+@export var debug_readout := true
 
 # Stamped on every synthetic event so the bridge never mistakes its own echo for
 # 3D pointing (InputEvent is a Resource — meta survives in-process delivery).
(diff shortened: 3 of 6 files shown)


Phaazoid:
🤖 Claude says:

**Self-review round** (8 finder angles → adversarial verification), applied in `c086799`. Also: your live run's "clicks do nothing" was the bridge working *quietly* — menus open inside the PiP at 0.35 scale (you found them). `pip_scale` is the knob if you want the corner bigger; whether PiP-hosted menus are livable until 4c is exactly the feel call the PiP existed to inform.

## Fixed (each falsified: mutant in → red → restore)

1. **RMB cancel was dead after hovering a far cell** — the cancel reused the last hover's mapped position, which falls outside the container for cells the hidden camera isn't showing, so the event was silently dropped and the player was stuck in the mode. Cancel is position-blind (`game.gd:279`), so it now pushes at the viewport center, which always forwards.
2. **A click the board lock would refuse still snapped the hidden camera** — yanking the PiP off the AI's `pan_to`/`follow` mid-turn. The bridge now reads `game._board_locked_for_player()` (the one predicate) before acting; that same read also covers the MENU hole — Mission Select deliberately opts out of the modal lock, so the `can_process()` gate alone can't see it.
3. **demo_mode's rig died forever when the hidden end-of-mission banner claimed the modal lock** — the freeze-mirror now carves out demo mode (4a parity: the diorama's camera survives the battle ending).
4. The click branch regained its null arm; `_pip_native` reads the authored `custom_minimum_size` instead of restating 1280×720; two comments corrected to the **measured same-drain delivery model** (events parsed mid-dispatch deliver in the same flush — the "next frame" wording was wrong and is what led two review angles to a phantom camera-drift bug).
5. New `debug_readout` export: the bridge narrates picks/mappings/refusals on the help label for live bring-up.

## Refuted against probe evidence (for the record)

- Parse-to-delivery camera drift (two finders): the flush drain re-checks its queue, so there is no frame gap between the camera snap and the synthetic click's delivery.
- `get_global_rect()` ignoring container scale (two finders): measured scaled in three probes on this engine build.

## Proposals — new seams, your call (not built)

1. **`board_loaded` signal on `ScenarioManager.apply_scenario`** — the PiP exposes Restart / Load Game / Return-to-Title, and a *different-board* arrival (Load Game from another mission's save; Mission Select → another mission) leaves the 3D terrain mirror and the picker's `_tops` describing the dead board. Verified: `turn_started` deliberately does NOT fire on menu arrivals (#144's `_begin_turn` bypass), so the load funnel itself is the one honest hook — `apply_scenario` is the single point every swap crosses. Same-mission Restart/F2 are accidentally safe (identical tile_data).
2. **Reuse consolidations**: a `BoardSpace` flat-cell pair (the (x,z)↔(x,0,y) convention is hand-rolled in 5 places), a `GridUtils` cell→world helper (the bridge is the 9th copy of `to_global(map_to_local(cell))`), and the LookDev `cell_picker_demo` now duplicates the bridge's pick-and-highlight loop.

## Pre-existing finds (not this PR's; file if you agree)

- Esc during a demo/AI turn opens an **invisible** report card that freezes the diorama (4a-era; the card lives in the hidden container).
- `CameraController.CELL_WORLD := TILE_SIZE * 2` (32) disagrees with the actual 16px cell — camera clamp bounds compute at 2× the board's extent.
- Mid-execution clicks aren't state-locked (no EXECUTING state; harmless today — nothing pans during player playback).

## Declared 4b jank (unchanged, 4c/4d own them)

Wheel over the PiP also zooms the 3D rig (same global-poll family as the WASD double-pan); the PiP hover card lags for far cells (see the corrected PR body); menus live in the PiP until 4c.

— Claude (Fable 5) · 2026-08-13

Issue #228, opened 2026-08-14
Right-click cancels the most recently queued action (LIFO undo)


Phaazoid:
🤖 Claude says:

**Dev request, 2026-08-14:** *"When I have actions queued, and I right click, I want that to cancel them. Each right click cancelling the most recently queued action."*

A LIFO undo on the action queue, driven by the button the player's hand is already on. This is a **gameplay/UX** ask, not presentation — it should behave identically in the flat 2D game and the 3D view.

## What right-click does today

`game._on_right_click()` is **position-blind** and means one thing: *leave the current mode* (`exit_current_mode()` — drop the aim, the move choice, the target pick). It has no relationship to the queue. Cancelling a queued order today is a queue-panel affordance only.

So the ask adds a second meaning to the same button, and the interesting part is the precedence rule.

## The design question (dev's call)

**When a mode is open AND orders are queued, which does right-click do?** Two coherent readings:

- **A — mode first, then queue** (recommended): right-click exits the mode if one is open; if nothing is open, it pops the most recent order. One button, unwinds in the exact reverse order the player built it up. Predictable, and never surprises someone mid-aim by eating an order.
- **B — queue always**: right-click always pops the newest order regardless of mode. Fewer presses, but a player who opened Attack and changed their mind loses a queued move they never meant to touch.

Two smaller ones that need answering either way:

1. **Whose most-recent — the active squad's, or the whole board's?** The queue is per-squad (`Squad.action_queue`), and the panel shows the active squad. Almost certainly "the active squad's newest order", but say it out loud.
2. **Volley siblings** — an AoE attack is N `AttackAction`s sharing one `volley` array. Popping one must pop the whole volley or the queue lies (Law #2). The existing note on this ("volley cancel propagation needed if per-row queue cancel UI ever arrives") is exactly this problem arriving.

## Existing machinery to reuse (Law #4 — do not build a second cancel)

`SquadManager.remove_action` is the one removal path, and `game._on_unit_action_cancelled` already owns the fallout (clear the planned path, redraw icons/leader range/projected units). `revert_if_only_hold` handles the squad-left-with-nothing-but-holds case. A LIFO pop should be a thin caller over those, never a parallel implementation.

## Watch for a collision

[...]

Issue #232, opened 2026-08-14
Suspected: downed units go invisible in the 3D view


Phaazoid:
🤖 Claude says:

**Suspected regression in the 3D view, derived from code and NOT yet observed in play.** Filed rather than fixed because it needs one live look to confirm, and because if it is real it makes rescue unreadable in 3D.

## The claim

**A downed unit's 3D sprite is probably set invisible every frame**, one line after it is given its downed texture.

## The chain

1. `Unit._go_downed` calls `_show_downed_sprite(true)`, which sets `map_sprite.visible = false` (`Classes/units/Unit.gd:510-514`) — in 2D this is correct, because a *separate* downed sprite node is shown in its place.
2. `UnitVisuals.sprite` **is** `$MapSprite` (`Scenes/Unit.tscn:32-34`, `sprite = NodePath("../MapSprite")`).
3. `UnitMirror._sync` copies `sprite.visible = unit.visuals.sprite.visible` every frame (`Classes/presentation/UnitMirror.gd:100`) — added by #222 so the mirror inherits the projected-ghost / knockback-ghost hide (`UnitVisuals.set_projected`, `OverlayManager`'s ghost pairs).
4. But `UnitMirror._sync:90-93` has *already* handled downed correctly by swapping `UnitSprite3D` to its downed texture.

So the visibility copy — which exists for the ghost-hide case — appears to catch the downed case as collateral, overriding the texture swap that was doing the right thing.

## Why no test caught it

`tests/presentation/test_unit_mirror.gd:69-75` asserts only `sprite.is_downed()` — the texture state — and never checks `visible`. `test_overlay_mirror.gd` has no downed case at all. The #222 falsification battery covered the ghost-hide direction (a mover's real sprite must hide behind its ghost) but not the downed direction.

## Why it matters if confirmed

Downed bodies are gameplay-critical, not decoration: they are rescue targets (`RulesService.is_rescueable`, the whole #33/#124 rescue flow), they hold their cell against enemy passage rules, and they carry a visible downed clock in the panels. A body you cannot see on the board is a body you cannot plan a rescue around — in the view that is now the primary one.

## To confirm

Load a mission in Battle3D, get a unit downed, look. Expected-if-real: the unit vanishes entirely from the 3D board while still listed in the 2D panels and still occupying its cell.

## Likely shape of the fix (not decided)

[...]

PR #233, opened 2026-08-14
HD-2D stage 4d: the camera pass (#176, v0.17.0)

Phaazoid:
## HD-2D stage 4d: the camera pass (#176, v0.17.0)

Stage 4c made the 3D view solo-playable. Surveying what was left turned up something worse than polish: **the camera was inert during play.** The only production writes to the rig were a one-shot fit at board load and your own keys. Meanwhile `AIController` pans the *hidden 2D* camera and awaits the full 0.7s beat, plus 0.6s of plan-read — so watching an enemy turn in 3D was **~1.3 seconds of dead air per squad in front of a motionless frame**, and a squad outside the frame was never seen at all.

Six more verified defects sat beside it. This closes all seven.

### The camera follows the action now

By **mirroring the 2D camera**, which already knows where the action is — no second follow seam, and because the 2D owns the tween the 3D inherits `Pacing.AI_SQUAD_PAN` exactly, so that beat stays one number with one reader. Same poll-don't-wire trick the overlays use.

The gate is `ai_locked`, deliberately **not** the broader `_board_locked_for_player()` — that one also covers MENU, and Mission Select opts out of the modal lock, so mirroring on it would yank the rig to a stale position the moment a menu opened. There's a test for each direction; the wrong-predicate mutant reddens the menu case by name.

### `_fit_camera` actually fits now

It was passing a board span **in cells** to `set_zoom`, which wants a camera **distance**, and the clamp silently ate the result — Prolog needed ~82 and got 24. Both authored missions opened showing a fraction of the board.

`frame(volume)` now solves per AABB corner against the camera's own projection matrix, so it accounts for fov, aspect, pitch and perspective rather than approximating. The zoom ceiling is derived from the board it just fitted, so a fit can never be eaten by its own clamp again. The falsification for this reproduces the shipped bug exactly: Prolog's corners off-camera at load.

### Free orbit, per your ruling

Yaw orbits freely and **rests where you leave it**; **Q/E step to the next 90° detent**, realigning in one press from any angle. That repeals the locked snap-only contract, and `presentation-effects.md` records why it was safe to: facing already judges against the *live* camera basis, so nothing structural depended on a quantized resting yaw.

**Orbit is on middle-drag by default**, and the click/drag threshold is built — so flipping `orbit_button` to right-drag in the inspector is genuinely all it takes, and you can settle middle-vs-right by feel in one session. Right-click keeps its press-to-cancel meaning at the default, which is what keeps #228's queue-undo clean.

### The rest

- **WASD drove both cameras.** The 2D one now stands down under the existing `board_input_delegated` — and only its *keyboard branch*, because `follow()` needs that same `_process` and gating the loop would have killed the feature this stage exists to build. There's a test for that specific over-reach.
- **R reset to the wrong place** (home was captured before the fit ever ran) — it returns to the opening shot now.
- **The rig's pan was unbounded** — it's clamped to the board plus a margin.
- **A menu left the camera and pointer live behind it** — both refuse now.
- **SPACE was dead** (the delegation gate sits above it) — it recentres the diorama on the pointer cell, handled in Battle3D so `game.gd`'s one bool stays the arc's only edit to its input path.
- **Facing was only re-judged when a unit moved**, so rotating left every standing sprite facing the old angle. Snapping never protected against that either; the mirror re-judges on camera turn now.
- The six camera constants became `@export`s (the tuning rule).

### Verification

- Full suite **1405/1405 green** (+18), zero orphans.
- **Eight falsification mutants**, each reddening its own named clause. The three that decide the architecture: a hover must not drag the rig, a menu must not yank it, and delegating input must not kill `follow()`.
- One note on process: the wrong-gate mutant reddened an *earlier* case first and aborted the suite before reaching the menu case — so I isolated it and re-ran to prove the menu case fails on its own merits, rather than accepting a composite red.

### Feel-check

1. **Watch an enemy turn.** That's the headline — the camera should travel to each squad and the beat should read as deliberate rather than dead.
2. **Opening framing** on both missions — `fit_margin_cells` is the knob if it reads too far back.
3. **Orbit** — sensitivity, and whether middle-drag is right. Flip `orbit_button` to `RIGHT` in the inspector to try the other; both work.
4. **Texel shimmer at off-axis rest** — the accepted cost of free rotation. Your call whether it reads as characterful or broken.
5. Q/E realign, R, SPACE, and that WASD now moves one camera.

Also worth knowing: whether orbit/zoom should stay live *during* an AI turn is a one-line fork, noted at the source — right now the camera is fully handed over while the AI acts.


Phaazoid:
🤖 Claude says:

Pushed `833139c` — the four camera-side findings from your v0.17.0 playtest, folded into this PR since they are all the same subsystem. **v0.17.1**, so the next build you play is distinguishable from the one you gave feedback on.

**Opening zoom.** The fit was doing exactly what it was told and the instruction was wrong: it framed the whole 64×40 board. `frame()` now takes a **shot** and a **bounds** — the shot is what is on screen at load (your own squad, `opening_view_cells = 18`), the bounds are the box the view may never leave, and *only the bounds* set the zoom ceiling and the pan limit. Solving the ceiling off a close shot would have clamped you out of ever reaching the far corner, which is the same class of bug as the cells-passed-as-a-distance one this stage already fixed.

**AI turn squares up.** On the *edge* into the turn, not every frame — idempotent either way today, but if orbit is ever allowed to stay live under an AI turn, a per-frame snap would fight your drag instead of squaring up once.

**Sensitivity** 0.35 → 0.25. Your `orbit_button = RIGHT` edit is committed as yours — that is the knob doing its job, and the doc now records the flip rather than the original default.

**Help text.** It still said MMB after you flipped the binding, so it now *reads* `orbit_button` instead of restating it. And `RMB cancel` was misleading — see below.

### Right-click is not broken

RMB cancels the current **mode** (an open aim, a move pick). It has never touched queued orders, so with nothing open it correctly does nothing visible. Cancelling *orders* is #228, unbuilt. My label was the bug; it now says "cancel aim", and `SHORTCUTS.md` spells out the distinction. Since you are hitting it in play, #228 looks worth pulling forward.

### Two falsification notes

A mutant aiming the opening window at the **board's centre passed**. Both authored squads start near enough to the middle that "every player unit is on screen at load" is satisfied by a window aimed anywhere central — so the obvious assertion pinned nothing. The aim itself is now asserted, with a guard that reds if a future map does start its squad dead centre.

Separately, the follow suite's existing non-vacuity guard caught its own obsolescence unprompted: the rig now *opens* over the squad, so panning to a player unit no longer moves it. That case now displaces the rig first.

`test_both_authored_missions_open_with_the_whole_board_in_frame` asserted the contract this commit deliberately changes, so it split in two — "opens on the player's squad" plus "zooming fully out still shows the whole board". The second keeps the original bug pinned: that bug made the far corners unreachable at **any** zoom, not merely unframed at load.

**1409/1409, 0 orphans.** Four new clauses, each falsified by its own mutant.

Not included: your live-editor churn in `Battle3D.tscn` / `LookDev.tscn` / `lookdev_meshlib.tres` (uid stamps, `rotation_degrees` → `transform`, and the board you painted in LookDev) — left in your working tree.

### Still open from your list, not in this PR

- **Downed units invisible** (#232) and **enemy sprites not tinted red** — both diagnosed, both mirror-side, next PR.
- **Boot into 3D** — your call, its own change; flagging that it also needs the dev-overlay lookup fixed or F1 stops working by default.
- **Bug reporting in 3D** — verified gap: the screenshot grabs the SubViewport, which is now transparent with the board hidden, so `board.png` has no diorama in it.

— Claude (Opus 5) · 2026-08-14

Issue #231, opened 2026-08-14
Dev tools in the 3D view: what do they even do there? (design session)


Phaazoid:
🤖 Claude says:

**This is a DESIGN-SESSION seed, not a port plan.** The dev's framing (2026-08-14):

> *"If we're porting the game to 3D, the editing needs to be in 3D too. It would be very weird to have to switch to the 2D view to edit a level, then switch back to see how the changes look. If this is a big deal, then building back the full functionality of the dev tools in 3D can be its own step, one tab at a time. We don't have to include it in this pass."*

and, on how to approach it:

> *"We're probably going to need a proper grill back and forth on what the functionality of the dev tools will even do in 3D, if there's anything new, how things adapt, etc, so don't make too many assumptions there."*

So: **the answers below are deliberately not proposed.** What follows is the verified current state plus the questions a grill session (`/grill-me`) would work through. Nothing here should be read as a decision.

## Verified current state — why nothing works today

Two independent causes, both confirmed:

1. **`game.dev_overlay` is `null` for the entire Battle3D session.** `game.gd:106` resolves it by **absolute path** — `get_node_or_null("/root/Main/DevOverlay")` — and under Battle3D the tree is `/root/Battle3D/Main/DevOverlay`. No crash (every reader is guarded; `tests/ui/test_game_scene_without_dev_overlay.gd` exists to pin exactly those guards), but silently: **F1 is a hard no-op, dev spawn is dead, the tile brush is dead, the unit editor never opens on a dev click, and the dev-mode checkbox never syncs.**
2. **4c's delegation gate sits above the dev arms by design.** `game.gd:192` returns on `board_input_delegated` before both the tile-brush branch (`:195`) and the SPACE spawn arm (`:210`) — commented as deliberate. So even with cause 1 fixed, board editing only unblocks in `View.FLAT_2D`.

Also true today, worth knowing before designing anything:

- The DevOverlay Window itself is fine — its own lookups are **relative** (`Classes/dev/DevOverlay.gd:11-12`) and survive re-rooting. It is only unreachable, not broken.
- **F2 works correctly in 3D** and rebuilds the mirror (`reload_current` → `board_loaded` → `battle3d._on_board_loaded`).
- **F3 works but ships the wrong screenshot** — `BugReporter.capture_frame` grabs `game.get_viewport()`, which is the SubViewport that 4c set `transparent_bg = true`. The `board.png` is 2D UI on transparency; the diorama is not in it.
[...]

Issue #149, opened 2026-08-06
F3 during an AI turn crashes: game.selected_unit dangles when the selected unit dies


Phaazoid:
Reported from play 2026-08-06: **pressing F3 to file a report during an AI turn crashes the game.**

```
Trying to assign invalid previously freed instance.
  at Classes/dev/BugReporter.gd:213
```

## Root cause (read from the code, not inferred from the message)

`BugReporter._plan_squad()` falls back to the selection when no squad is active:

```gdscript
func _plan_squad() -> Squad:
	var active: Squad = game.squad_manager.active_squad
	if active != null:
		return active
	var selected: Unit = game.selected_unit     # <- line 213
	if selected != null:
		return selected.squad
	return null
```

**`game.selected_unit` is never cleared when the unit it points at dies.** `game._on_unit_died` (`game.gd:776`) does three things — `overlay_manager.handle_unit_death`, `squad_manager.handle_unit_death`, `refresh_action_queue` — and none of them touches the selection. The only two writers are `_click_idle` (sets it) and `exit_current_mode` (nulls it, `game.gd:463`).

So the field holds a **dangling reference** to a freed `Unit` from the moment your selected unit dies until you next click or change mode. `if selected != null` does not catch it — a freed instance is not `null` — and the crash is the *typed assignment* on line 213 (`var selected: Unit = ...`), which is why the message is "assign" rather than a nil-deref.

An AI turn is simply the most likely place to meet it: it is the one stretch where units die while the board is locked and you cannot click anything to refresh the selection. It is **not AI-specific** — the same dangling read is reachable any time your selected unit dies, e.g. to a counter-attack or a burning tile on your own turn.

## Why the fix probably does not belong in BugReporter

Guarding line 213 with `is_instance_valid()` fixes this one crash and leaves the dangling field for the next reader. `game.selected_unit` is read by (per CLAUDE.md's #107 note) all four mode click-handlers plus their three `HoverPresenter` twins, so the honest fix is at the writer: clear the selection in `_on_unit_died` when the dead unit *is* the selection. Worth checking `went_downed` at the same time — a downed unit is not freed, so it does not dangle, but staying "selected" through a down may or may not be intended.

Deliberately not fixed as part of the #127 AI sweep it was found during — different seam, and the fix is a decision about selection lifecycle rather than a typo.

## Repro

1. Any board with an AI faction enabled and a player unit the AI can kill.
2. Click one of your units to select it.
[...]

PR #244, opened 2026-08-14
Live diorama + a pickable ground plane: the presentation half of dev-tools-in-3D (#231)

Phaazoid:
**PR 1 of 2 for [#231](https://github.com/Phaazoid/Godoiosis/issues/231) — the presentation half. No input changes, so nothing here alters how the game plays; it makes the diorama tell the truth while you edit.** PR 2 is the input plumbing (brush routing, the RMB/SPACE bindings, the orbit rebind, the 3D block ghost).

Split at the risk boundary per your call after the scope grew on contact with the code.

## What was actually broken

The [design session](https://github.com/Phaazoid/Godoiosis/issues/231#issuecomment-5296418311) found the issue's own "verified current state" was stale — cause 1 was already fixed by boot-into-3D. What remained, and what this PR fixes:

1. **Erasing a tile in 3D made that cell permanently unpaintable.** `BoardPicker` builds its pick table from `get_used_cells()`; no mesh → no column → `NO_CELL` forever. Repair meant F4 to flat 2D and back — the round trip this arc exists to kill, reintroduced by the erase button.
2. **Painting showed nothing until F2.** `rebuild` fired only on `board_loaded`, `refresh_states` only on `turn_started`.
3. **PATROL zones had no 3D twin at all** — and PATROL is the tile brush's default kind, so zone painting in 3D drew nothing.

## The picker plane

A column with no block now answers if it lies inside the caller's `plane` rect — an implicit floor at `BoardSpace.FLAT_TOP_LEVEL`. Resolved **inside** the DDA walk rather than as an analytic intersection before or after it, so a real column in front of a hole still wins by ray order instead of by a hand-written comparison — invisible on a flat board, wrong on LookDev's towers.

**`FLAT_TOP_LEVEL` is a named constant because y = 0 reads plausible and is wrong.** `of_flat` parks every cell at y-index 0, so the face things sit on is one `CELL_SIZE` up; a plane at the slab's bottom resolves the wrong column at grazing angles. Same class as the cells-passed-as-a-distance bug 4d fixed. `UnitMirror.COLUMN_TOP` derives from it now rather than restating it.

The plane is a **required** parameter, not a defaulted one — a forgotten default would silently un-click every hole. LookDev's two demo callers pass `Rect2i()`, a no-op there.

**Behaviour change worth flagging:** with the apron always on (your ruling — not dev-mode-only), a click one cell off the board now reaches `_on_left_click` and, in `CHOOSING_MOVE`, exits the mode. That is **2D parity restored** — `game.gd:210` has always derived a cell for any mouse position — but it is a change, and it has its own pin.

## Live mirroring

**Terrain** syncs per cell, driven while `DEV_MODE` is active. The sim never paints terrain, so gating there keeps an O(board) diff entirely out of the shipping game while still catching every writer — including future ones — with no trigger site to remember.

> An engine signal was the first choice and is not available. **Measured, with a property write as the control: `TileMapLayer.changed` does not fire on `set_cell` or `erase_cell` in 4.7**, same frame or across frames.

**Tile states** go live for everything, not just dev paints, retiring the declared v1 approximation. `_fire_markers` becomes a keyed reconcile (UnitMirror's #149 pattern) — strictly cheaper than the free-and-rebuild-everything it replaces, so no pooling. Polled from `OverlayMirror`: a `states_changed` signal would fire *inside* the resolver's per-effect loop and churn markers many times per pass, where the poll coalesces a frame into one reconcile.

A repaint can add or erase a **column**, so `_tops` and the camera bounds refresh with it. Bounds only, never a re-frame — the new `look_dev_camera.rebound()`, which `frame()` also calls for its ceiling/pan half so the two cannot disagree. Inert for LookDev, which never calls `frame()`.

## Zones

PATROL and the picked-zone highlight now mirror, **visibility-gated**: the 2D shows them only while the Tile Brush tab is up, so mirroring their cells alone would leak authoring scaffolding into the play view. Mirror the question — *should this be on screen* — not the field the cells sit in.

**Needs your eye (headless can't see it):** the zone band moved from sort −2 to −3, with the highlight alone at −2. `sort` drives both render priority *and* the geometric lift, so zones now sit 0.004 lower. −1 was unavailable — SQUAD lives there and would share the lift, i.e. z-fight. Worth a look at a capture zone under a squad-range overlay.

## Law #4 collapses riding along

- **`BoardMirror._has_fire` enumerated `FIRE_STATES` itself**, which `Terrain.gd` forbids by name (*"no reader may enumerate fire members itself"*). It and a third copy in the test suite are gone; `burning_cells()` is the one form.
- **`battle3d._board_volume` derived the board bbox a second time** — `BoardPicker.used_rect`/`max_top` are statics now and both sides read them.
- **The apron is not a new number**: `paint_apron_cells` defaults to `CameraController.EDIT_MARGIN_CELLS`, which is already how far 2D lets you pan out to paint past the edge.

## Verification

**Full suite 1438/1438.** Eleven mutants, each isolated where a red sibling would otherwise have hidden the intended case.

**One mutant PASSED, and that was the finding.** The first version of the standing-flame case asserted node identity across idle frames — which proves nothing, because `OverlayMirror` value-diffs the burning set and never calls the reconcile twice on a static board. Free-all-and-recreate looked identical through it. It only distinguishes a reconcile from a rebuild if something changes meanwhile, so the case now lights a *second* fire and checks the first marker survived it.

Knobs added, per the tuning rule: `paint_apron_cells` on `Battle3D`.

**No `project.godot` edit** — the merge Action reads the label. Labelled `type/feature`.

You are on `feat/231-dev-tools-3d-presentation` — launch it.


Phaazoid:
🤖 Claude says:

**Feel-check round 1 folded in (`ccbb382`). Suite 1439/1439, 0 orphans. Still on `feat/231-dev-tools-3d-presentation`.**

## The bug: authoring zones drew twice

Reported in play — with the Tile Brush tab up in the 3D view, zone tiles rendered in the flat 2D style overlaid on screen, on top of the 3D fills this PR added.

`battle3d._set_board_visible` hides every 2D board visual in the 3D view but **deliberately skipped** `zone_overlay` and `zone_highlight_overlay`. That skip was *correct* before this PR: those layers were the only way to see zones in 3D at all, and a blanket restore would reveal PATROL zones in play. Giving them a 3D twin made it wrong, and nothing said so.

The fix is not just deleting the skip, because of a trap I walked into. `_fill_gated` decided whether to mirror by reading the 2D layer's `.visible` — so hiding the 2D layer takes the mirror down with it. That field had quietly acquired **two writers with two meanings**: *"the Tile Brush tab is up"* (DevOverlay) and *"the 2D board is hidden because we're in 3D"* (battle3d). [#232](https://github.com/Phaazoid/Godoiosis/issues/232)'s shape exactly, inside the function whose own comment cites #232.

So the intent gets its own name on `OverlayManager`:

| | means | written by |
|---|---|---|
| `zones_authoring_visible` | the Tile Brush tab is up | DevOverlay |
| `board_rendering` | the 2D board draws at all | Battle3D |
| `.visible` | the **product** — purely a render fact | computed |

The 3D mirror reads the **intent**; the 2D layers read the product. Both halves hold at once: flat tiles stay dark behind the diorama while the 3D fills draw, and F4 brings them back because the intent never changed. `board_rendering` defaults true, so the flat game and every headless fixture are untouched.

The skip in `_set_board_visible` **stays**, and now has a reason that survives: those layers are computed from both inputs, so a blanket true would reveal patrol zones in play and a blanket false would be re-overridden the next time the tab changed.

## Correction to my own launch note

I told you to check "paint a tile and it appears immediately" on this PR. **That was wrong** — PR 1 has no input changes by design, so the tile brush cannot receive input in the 3D view at all; the `board_input_delegated` gate at `game.gd:199` still returns above its arm. Making it work is PR 2's whole job. I described PR 2's payoff as PR 1's.

What is actually reachable here: **F4 to flat 2D, paint, F4 back** — the diorama is already updated with no F2, which is the live-sync mechanism working. Mid-turn fire is observable in normal play. Pre-existing zones mirror (and no longer double-draw).

## Falsification, under the new triage

One mutant, on the one suite file, 9 seconds: drop `set_board_rendering` and `test_the_2d_zone_layers_stay_dark_in_3d_while_the_mirror_draws_them` reds. It qualifies under the narrow rule — a `visible`-quirk assertion sitting several hops from the change.

For the record, the eleven mutants earlier in this PR were over-spend under that rule; the picker ones were one-hop assertions against a function I'd just written. Three were warranted (the vacuous flame case, the terrain wire, the visibility gate) and those are the ones that found something.

— Claude (Opus 5) · 2026-08-14

Issue #281, opened 2026-08-15
Path arrows don't climb ramps — they stay at one level across a slope


Phaazoid:
🤖 Claude says:

**Found in the #273 feel-check** (dev, 2026-08-15): *"path finding arrows don't slope up slopes, they stay flat on the top level."*

#273 lifted the board, units, overlay fills and markers onto their cells' real surfaces. Path arrows are the one channel that did not follow the slope.

## Why they are different from everything else #273 fixed

Overlay **fills** are cells — `OverlayMirror._fill` maps each one through `BoardSpace.of_cell(cell, level)` and the layer lies on the surface. Arrows are **sprites**, mirrored by 2D pixel position through `_anchor_px`, which resolves the pixel to a cell and takes that cell's `surface_point`. So each arrow sprite *should* already pick up its own cell's height — and per the report it does not, which means either the pixel→cell resolution is landing on the wrong cell for arrow sprites, or the arrows are anchored once for the run rather than per sprite.

Worth checking before assuming a fix: `_anchor_px` floors `px / TILE_SIZE`, and a sprite whose `global_position` is a cell CENTRE lands correctly, but one anchored at a corner or carrying a half-cell offset resolves to a neighbour — which on a slope is exactly a one-level error.

## The harder half, which is design not plumbing

Even with the anchor right, an arrow **crossing** a ramp is a quad spanning two surfaces at different heights. Options, none picked:

- Leave each arrow flat at its own cell's surface — a visible stair-step along a diagonal run.
- Tilt the arrow to match the slope it sits on, which needs the ramp's rise on the presentation side (available: `BoardHeights.ramp_rise_at`).
- Lift arrows onto a small constant clearance above the surface and accept the step, the way the flame already clamps its ground gap.

The middle one is probably what reads best and is the most work; the first is nearly free and may be enough at this camera distance.

## Done when

A move path that climbs a ramp shows arrows on the ramp's surface rather than at one level, from every orbit angle. Falsified against a mutant that anchors every arrow at level 0.

Out of scope for #273 by the dev's own call. Related: #273 (the 3D elevation render), #227 (path arrows read flat on the 3D board — the *lift* question, and worth reading together with this).

— Claude (Opus 5) · 2026-08-15

Phaazoid:
🤖 Claude says:

**Built — [PR #306](https://github.com/Phaazoid/Godoiosis/pull/306).** Awaiting the feel-check.

[...]

Issue #260, opened 2026-08-15
Dev tools: paint elevation and ramps (scroll sets level, button resets to 0)


Phaazoid:
🤖 Claude says:

**Follow-on from the #218 verticality design.** Spec: [`docs/design/verticality.md`](../blob/main/docs/design/verticality.md). Needs slice 1's height store to exist first.

Authoring surface for elevation in the dev tools. **Deliberately minimal** — the dev scoped it himself:

> *"for now, we can keep it simple - scrolling up and down can determine what level I want to place the tile at, with a button dedicated to resetting it to 0. Eventually I'll want to do things like paint textures on walls, etc, but that doesn't have to come first."*

## Scope

- **Scroll wheel sets the level the brush places at.** Current level shown somewhere visible.
- **A dedicated button resets it to 0.**
- Ramp painting (axis + direction), since a height map with no ramps is unreachable terrain.

This is a **paint mode** on the existing Tile Brush, not a palette expansion — elevation is per-cell data, so the shape #174 established for Tile States is the one to copy. That is precisely why the design put elevation in a per-cell store instead of tileset custom data: per-tile elevation would have needed a distinct tile per level and blown up the palette.

## Explicitly out of scope

- Painting textures onto wall faces (the dev named this as a later want).
- **3D-native authoring** — the dev has said he expects that to be the issue after [#231](https://github.com/Phaazoid/Godoiosis/issues/231), and it stays a separate ticket. His constraint on it, recorded on #218: *"I want it to still work in both modes — we'll just need new sprites to represent height differences."*

Design: [#218](https://github.com/Phaazoid/Godoiosis/issues/218). Dev tools in 3D: [#231](https://github.com/Phaazoid/Godoiosis/issues/231).

PR #341, opened 2026-08-16
One brush: tiles carry their elevation, ramps wear their ground (#340)

### LINKED ISSUE #340: One brush: paint tiles at any elevation, and ramps wear the tile painted on them
Elevation and terrain are one brush in name only. #260 made elevation a paint **MODE** inside `TileBrushTool` rather than a second tool, which was right at the time — but authoring still means switching modes to say how high a tile sits, and a ramp is a property of the cell that the tile you painted knows nothing about.

Dev ruling (2026-08-16):

> I want the elevation brush and the tile brush to not be separate. Ramps are tiles, and tiles should be paintable at any elevation.

and, on where a ramp's slope direction should live:

> Same as [keep rise on the cell, just unify the brush], but I want to be able to paint tiles on ramps. only tiles that are flat, not things like rocks, lanterns, etc. grass, mud, etc.

So the rise stays a per-cell fact — no tileset ramp column, no `RulesService` change — the mode collapses into one brush, and a ramp renders wearing the flat tile painted on it.

## Why now

**Every ramp currently renders as `dirt_ramp`** — one hardcoded procedural mesh (`BoardMirror.RAMP_ITEM_NAME`), grass-top / dirt-side, regardless of the terrain painted on the cell. A stone ramp is dirt. That is the bug this ask fixes, and it exists precisely *because* a ramp isn't a tile.

**Nothing needs migrating.** No `.tres` under `Scenarios/` carries a `ramps` entry, and no ramp art exists on disk. The store and the meshlib can be reshaped freely.

## What changes

1. **`PaintMode.ELEVATION` is deleted.** The enum becomes `{ TERRAIN, ZONE, STATE }`; the level and rise rows show under TERRAIN, and the wheel + Z/C keys stay live while the terrain brush is armed.
2. **Painting a terrain tile writes the height too.** `_paint_tile` gains the `board_heights.set_cell` call `_paint_elevation` holds today. Raising a cell = repaint it at the new level; there is no separate height-only gesture any more, which is the point.
3. **A rise is refused for a standing prop; the level is not.** A rock at level 5 is fine, a rock that *slopes* is not. The gate is the existing `GridUtils.stands_up_of()` — derived from `prop_shape`, so no new predicate — checked against the brush's selected tile so the illegal state is never authored.
4. **The ramp wedge wears that tile's art.** The generator already emits one meshlib item per tile via `BoardMirror.tile_item_name`; the ramp variant is one more `_add_item` in the same loop, gated on `not stands_up`. `BoardMirror.ramp_item_for_cell` replaces the bare `ramp_item()` at its two call sites, with `dirt_ramp` staying as the fallback for the cases `item_for_tile` already skips.

## Flagged, not decided

- **Meshlib size.** The tileset has 322 tiles, 15 of them non-FLAT; the meshlib holds 325 items today, so a ramp variant per flat tile roughly doubles it. Geometry is negligible (a wedge is a handful of triangles) but file size and load time are not measured — measuring first, and the fallback is to gate variants on walkable FLAT tiles only.
- **TUFT.** `stands_up_of` reads a flowery grass tuft as "standing", so it will refuse a rise despite being walkable ground and arguably the "grass" named above. Excluded initially on purpose: its plants would need planting on a tilted face via #281's `BoardSpace.surface_transform`. Say the word and it becomes a follow-up or part of this.

Phaazoid:
Closes #340.

Elevation stops being a paint **MODE** and becomes always-live rows on the **TERRAIN** brush, so one click paints a tile *at* a height. Height stays in the per-cell `BoardHeights` store for #260's reason — a per-*tile* elevation would need one grass tile per level — but "how high" was never a different QUESTION from "which tile", only a different store, and a mode per store made authoring a raised cell two gestures.

## What you'll notice

- **The Elevation mode is gone.** Level and Ramp Rise sit under Terrain; the wheel and Z/C work while the terrain brush is armed.
- **Raising a cell = repainting it at a new level.** There is no height-only gesture any more.
- **Painting virgin board now works in one click** — it creates the ground it raises. The old "elevation refuses a groundless cell" guard is answered by ordering (tile first, then height) rather than a refusal.
- **A ramp wears the tile you paint on it.** Previously every ramp rendered as the one generated `dirt_ramp` regardless of terrain, so a stone ramp read as dirt.
- **The Ramp Rise picker greys out for a tile that stands up** (rock, lantern) — your "only tiles that are flat" rule. A prop can still be painted at any *level*, it just can't slope.

## The one thing worth pushing back on after you feel it

`verticality.md` had parked this exact question, and the merge answers it: **repainting a tile over a terrace now flattens it to the brush's level.** That is the merge doing what was asked (a tile is painted AT a height), but it's a feel call, not a code one. If it turns out to be annoying, the fix is a "texture only" toggle on the brush — *not* a second mode.

## How it's built

| | |
|---|---|
| Rise stays per-cell | `RulesService.can_step` is untouched. This retires the tile-orientation candidate `verticality.md` had parked as an open Law #4 question — a ramp's direction varies per **placement**, which is exactly what a per-tile column cannot say. |
| Ramp art | `gen_lookdev_assets.gd` emits a wedge variant per FLAT tile beside its block, wearing that tile's own atlas UV; `BoardMirror.ramp_item_for_cell` picks it by name exactly as `item_for_cell` picks the block. **293 variants; meshlib 325 → 618 items, 1.35 → 1.85 MB** (measured before building the rest — a wedge is cheaper than a block, so it's +38%, not the 2× I'd budgeted). |
| The flat gate | `GridUtils.stands_up_of` — the existing `prop_shape` derivation, no new predicate. It lives in `TileBrushTool.selected_rise()` so the **ghost** reads it too; gated at the paint site instead, the preview would show a sloping rock the click then refuses. |
| `dirt_ramp` | Stays as the fallback for the cases `item_for_tile` already documents (empty cell, rotated alternative, multi-cell art) plus a prop on an already-sloped cell. Without it those cells render as a hole. |

**TUFT is excluded, deliberately and provisionally** — `stands_up_of` reads a flowery grass tuft as standing, so it refuses a rise despite being walkable ground and arguably the "grass" you named. Its plants would need planting on a tilted face via #281's `BoardSpace.surface_transform`. Say the word and it gets its own ticket.

## Tests

The cases that pinned the mode split were **re-aimed at ZONE/STATE rather than deleted** — the rule they guard (a key that retunes a brush you cannot see is worse than a dead key) is unchanged, only which modes can see it. Two reversals are recorded in place rather than quietly dropped: a groundless cell now *does* owe a ghost, and the 2D view now *does* draw one.

Four mutants, each redding exactly the case aimed at (checked by name, not just by the suite going red):

| Mutant | Case that reddened |
|---|---|
| drop the height write from the merged paint | `test_painting_writes_the_tile_the_level_and_the_rise_together` |
| drop the `stands_up_of` rise gate | `test_a_tile_that_stands_up_paints_flat_however_the_rise_is_set` |
| always fall back to the generic wedge | `test_a_ramp_wears_the_tile_painted_on_it_rather_than_the_generic_wedge` |
| remove the fallback entirely | `test_a_tile_with_no_ramp_variant_falls_back_to_the_generic_wedge` |

`dev` 171/171, `presentation` 243/243, 0 orphans. Canon swept: `verticality.md` (marker bumped to #340, both parked questions recorded as answered), `docs/SHORTCUTS.md`, and the tab-5 tooltip that had never mentioned Elevation at all.


Phaazoid:
🤖 Claude says:

Both feel-check findings fixed in `1638fb2`. You're still on `feature/ramp-tiles` — pull and relaunch.

**Water slopes were papered with the whole tilesheet — my regression, introduced here.** `_block_mesh` takes a `side_uv` and threads it to every side face; the `_ramp_mesh` call I added passed `top_uv` only, so the wedge's sides fell back to `Rect2(0,0,1,1)`. That's harmless for dirt and stone, whose side material *is* a single tile — but a water tile wears its own surface down the sides, where the side material is the composited atlas, so 0..1 mapped the entire sheet onto every water slope. `_ramp_mesh` now takes `side_uv` beside `top_uv`, and `_tri` grew the `uv_rect` its two triangular sides needed.

Four water variants changed; the meshlib is otherwise byte-identical. I checked determinism explicitly — two consecutive regenerations are identical to each other.

Pinned as a **law over every emitted variant** rather than a water case: both meshes are handed the same UV pair by the same loop, so any divergence is the bug, and a tile that grows an atlas-backed side later is covered without anyone remembering to add a case. Falsified against the *real* bug by restoring the previous commit's meshlib — it names all four water tiles, side UV `(1.0, 1.0)` against the block's `(0.033, 0.055)`.

**The Z/C lag is not a ghost bug.** The 3D preview is polled every frame in `battle3d._process` with no early-out, so it redraws the instant `selected_rise()` changes. What never changed was `selected_rise()`. `embed_subwindows` is off, so the game and `DevOverlay` are two real OS windows and a key event reaches only the **focused** one — and every dev key lived in `DevController._input`, inside the game subtree. So **F1, F2, F3 and Z/C were all dead whenever the dev-tools window had focus**, which is exactly where authoring leaves you: pick a tile from the palette, press Z to turn the rise, nothing happens until you click the game window back.

It's pre-existing — Z/C since #260, F1/F2/F3 longer — and #340 surfaced it by making Terrain the mode you turn a rise in. Very likely your other four sightings.

Fixed as a **forward, never a second handler**: `DevController.handle_dev_key` is the one public entry and `DevOverlay._input` calls it, so two windows are two input *sources* for one set of bindings. Idempotent by construction, since only one window has focus. A focused `LineEdit`/`TextEdit` is excluded first, or typing a scenario name would toggle the overlay and reset the board.

**The rule is now canon**, in `CLAUDE.md`'s sharp edges as you asked: in a multi-window project, *"which window is this handler in?"* is a question every binding must answer, and nothing in the code says it out loud — the handler compiles, runs, and is simply never reached. `docs/SHORTCUTS.md` gained a short "which window are these live in?" note.

`dev` + `presentation` 417/417, 0 orphans, exit 0. Both new cases falsified, each redding only the case aimed at.

One thing I did **not** chase: the regenerated meshlib's composited atlas image differs by one line from the build I made in the isolated worktree, while being stable across repeated runs here. Almost certainly a fresh-vs-established import cache reading a source PNG slightly differently. Benign, since the meshlib is a committed artifact CI never regenerates — but flagging it rather than leaving it unsaid.

— Claude (Opus 5) · 2026-08-16

Issue #272, opened 2026-08-15
Object Editor tab: terrain objects own their own fields (light, height, anything prop-specific)


Phaazoid:
🤖 **Claude says:**

**REFRAMED 2026-08-15 on the dev's ruling.** This issue was *"per-object flame and light presentation: the global knobs want a per-instance override"*. His words:

> *"We should just have an object editor tab, similar to the unit editor. Except instead of unit stats, it will edit object specific fields, like torch light, height, etc. Anything unique to terrain objects."*

That is a better model and the retitle reflects it. The old framing said *"override the global knob"*; his says **objects own their own fields**, with a tool to edit them. Same want, cleaner shape — and it makes the Law #4 question explicit instead of hiding it (below).

## The want

A dev-tools tab on the `UnitEditorTool` pattern, editing whatever is unique to a terrain object rather than to a unit. Named candidates: **light** (a torch/brazier's energy, range, colour), **height**, and *"anything unique to terrain objects."*

## What has to be sorted first, and it is not the UI

**There is no per-object store to edit.** Measured, not assumed: a prop today is a **TileSet custom-data layer** — `prop_shape` is read via `GridUtils.prop_shape_of(TileData)` off `grid.get_cell_tile_data(cell)`. So "an object" is a *tile type* carrying a non-FLAT shape, and #264's whole prop system hangs off that. There is no resource, and no per-placement record.

That splits the ask in two, and they are different features:

- **Per tile TYPE** — *"a lamp emits this much light."* Adding custom-data layers to the TileSet. Authorable in Godot's TileSet editor already; the tab's job would be surfacing it in-game so it can be tuned live on a real board instead of in the editor. Cheaper, and probably what covers most of the want.
- **Per PLACED instance** — *"THIS lamp at (4,7) is dimmer than that one."* Has **no store whatsoever** today. The existing precedent is `ScenarioData.elevations` / `ramp_rises`: sparse `Vector2i -> value` dicts saved with the board. That is a real feature with save/load and dev-brush implications, not a tab.

**Which of the two he wants is the first fork**, and the answer might be "type now, instance later".

## The Law #4 question the reframe exposes

Today `flame_light_energy`, `flame_light_range`, `block_height_scale` and friends are **global knobs on `BoardMirror`**, tuned in the Look tab and captured into `LookPreset`s. If an object carries its own light, what happens to the global?

[...]

Issue #253, opened 2026-08-15
Look tab presets: save/load/update/delete, attach one to a mission, and an authored starter set


Phaazoid:
🤖 Claude says:

**Dev ask, 2026-08-14, straight after the Look tab landed (#212 slice 1, PR #251).** His words:

> *"There are a lot of settings here... First of all, a list of saved presets, and a way to add to them/delete/update, same as all the other dev tabs. I'll also want to attach one of these saved settings to a mission, so they'll always load with it. I'll also want you to create a bunch of presets for me - standard ones, interesting ones, crazy ones, ones you think I should see, etc."*

Three parts, in dependency order.

## 1. Saved presets, on the dev-tab idiom

A preset is a named capture of the Look tab's `KNOBS` values. The surface should be the same one every other dev tab already has, not a new one: a Load dropdown, **Save As** (refuses an existing name), **Update** (load-gated to the loaded file), **Delete** (behind `DevWidgets.confirm_delete`), and a status label. `ScenarioTool` and `ItemEditorTool` are the shapes to copy.

Non-obvious constraints this inherits, all already paid for elsewhere:

- **Scan through `ResourceDir`**, never a raw `DirAccess.get_files()` — a source-extension filter matches nothing in an exported build (the sharp edge that made eleven catalogs silently empty).
- **`take_over_path` before every save**, or a re-save leaves later `load()`s handing back the stale object.
- **`select(-1)` before the dropdown's match loop**, since `add_item` auto-selects the first entry and the Update button writes to whatever the dropdown names.

## 2. Attach a preset to a mission

`ScenarioData` grows a reference so a board always loads with its look. Two things to decide before building:

- **What a preset STORES — the whole table, or only what differs from the authored scene.** Whole-table is self-contained and reproducible (a mission's look cannot drift when someone re-tunes `Battle3D.tscn`); diff-only is composable and survives new knobs gracefully. These pull opposite ways and it is a real fork, not a detail.
- **A preset naming a knob that no longer exists, and a mission naming a preset that has been deleted.** The missions doctrine already has the answer-shape for the second one: a declared-but-missing objective reads PENDING rather than being silently dropped, because silently dropping it converts a broken map into a different playable one. Same logic applies here — a mission pointing at a deleted preset should say so loudly, not quietly render with whatever the scene happens to hold.

[...]

Issue #326, opened 2026-08-16
Burrow cover should pop up the way tufts do


Phaazoid:
*Filed by Claude from the dev's playtest notes (2026-08-16), not written by the dev directly.*

## The ask

Dev, playtesting v0.39.0: *"burrow terrain should pop like the flowers do."*

This closes a scratchpad bullet that has been sitting in the Inbox since 2026-08-14, where the dev had already described the shape and doubted it was reachable:

> **Cover should POP UP in 3D the way units do — but as three individual mud bumps cropping up, NOT the whole sprite rotating vertical.** *"I worry this might not be as easy as a simple transform."*

He was right that a transform could not do it, and #280 has since built the thing that can.

## Why #280 is the answer

`PropShape.TUFT` splits a predicate that had one answer for every prop: the generator's bake question stopped being *does this tile stand up* and became *does its art LEAVE the ground* (`GridUtils.art_leaves_ground_of`). A tuft's art is drawn **twice** — flat, and popped — on purpose, and the popped half is produced by keying out the background, splitting what remains into 8-connected clusters, and standing **each cluster up at its own place**.

That is precisely "three individual mud bumps cropping up, not the whole sprite rotating vertical", generalized: the cluster count comes from the art, so three lumps in the drawing become three lumps on the board with nothing authored.

It also carries the finding that makes it correct rather than merely plausible — **a ground tile is drawn top-down, so a thing's y inside it is DEPTH INTO THE CELL, not height** (learned twice now: #263's `fence_ver` and #280's flowers). Each cluster is planted at x = its centre column, z = its **bottom** row, because the lowest drawn pixel is where the object meets the ground. Standing cover up any other way puts the back lumps in the air.

## The part that does not transfer

**Cover is not a tileset tile.** It is a dynamic terrain **state**, deposited by `BurrowAction` through `TerrainStateManager`, and drawn as a flat `Layer.TERRAIN` sprite quad by `OverlayMirror._terrain` — the same path as the FROZEN icon. `prop_shape` / `PropShape.TUFT` live on the **tileset's custom data** and are baked into the meshlib per tile by the generator, which a runtime state has no entry in.

So the mechanism transfers and the plumbing does not. What has to be decided:

[...]

Issue #308, opened 2026-08-15
OverlayMirror._fill diffs on cells only, so a rise painted under a drawn fill does not re-tilt it


Phaazoid:
🤖 Claude says:

**Found while building [#281](https://github.com/Phaazoid/Godoiosis/issues/281) (PR #306), flagged rather than fixed** — it was outside that ticket's scope and the dev asked for it to be tracked separately.

## The gap

`OverlayMirror._fill` value-diffs on the **cell list** before pushing:

```gdscript
cells.sort()
if _last_cells.get(layer, Array()) == cells:
    return
```

Since #281, how a fill *lies* on its cell also depends on that cell's **ramp rise** — and a rise is not in the cell list. `BoardSpace.of_cell(cell, level)` encodes only x / level / y, so painting a rise onto a cell whose **elevation does not change** produces an identical `Array[Vector3i]`, the diff short-circuits, and the fill keeps the tilt it already had.

Painting an *elevation* is fine: the level moves, the `Vector3i` changes, the push happens. It is specifically the rise-only edit that is invisible.

## How to reach it

The dev-tools elevation brush ([#285](https://github.com/Phaazoid/Godoiosis/issues/285)) with a fill layer already drawn — in practice the zone layers, since those are the ones visible while the Tile Brush tab is up. Paint a ramp under a painted zone and the zone tile stays flat over the new slope.

**It self-heals** the moment the cell set next changes (any brush stroke that adds or removes a cell, any mode change, `clear`), which is why this is low severity rather than a visible defect in play. No player-facing path reaches it: move/attack fills only exist during a mode the brush cannot be open in.

## Shape of the fix

The diff key has to include what the render now depends on. Options, unranked:

- Fold the rise into the compared value (a parallel `Array[Terrain.RampRise]`, or cells as a `{cell: rise}` dictionary) — correct, costs a small per-frame build on a hot path that currently allocates one array.
- Give `BoardHeights` a mutation counter and compare it alongside the cells — cheapest per frame, but it is a new seam on a class that has none, and per Law #4 worth checking against `zones_changed`'s precedent first.
- Have the brush invalidate `OverlayMirror._last_cells` on any paint, which is the narrowest change but puts knowledge of the mirror's cache in a dev tool.

`docs/performance.md` records that this poll runs every frame for every layer, so measure before picking the first option.

## Done when

[...]

PR #352, opened 2026-08-16
Objects tab: a tuned world-construction value saves itself to source (#272 slice 1)

Phaazoid:
Closes the frustration that started this: you tuned **Cover bump scale**, liked it, and there was nowhere to put it.

## The gap this fills

There are three destinations for a tuned value and only two had doors:

| destination | door |
|---|---|
| mission mood | `LookPreset` + Save (#253) |
| per-object variation | #272 slice 2 |
| **game constant** | **nothing — this PR** |

`cover_scale` was correctly `PRESET_EXCLUDED`: it is world construction, not mood — one number for every cover tile on every board forever. So a preset was never the answer, and Copy Values → paste-into-the-scene was the only path.

## What changed

**The three prop-geometry knobs LEAVE `LookKnobs.KNOBS`** for a new `ObjectKnobs` table — `block_height_scale`, `tuft_scale`, `cover_scale`. A move, not a copy, and the consequences are the point:

- Their `PRESET_EXCLUDED` entries go with them, so *"a mission cannot restyle world construction"* is now **structural** rather than three names on a list — the same shape as `DefaultLook.tres` sitting outside the scanned folder.
- **Copy Values stops emitting them**, so each knob has exactly one way to persist. Never two.

**The Objects tab (dev-tools tab 8) writes a tuned value into the `@export` default that declares it.** Measured, not assumed: `Battle3D.tscn` overrides none of these, so the script line *is* the single authority.

**This does not eat the Effects tab.** Every flame knob is scene mood, is preset-captured, and stays exactly where it is — only the three world-construction rows moved.

## The failure this is shaped around

A save that silently writes nothing. So `ObjectKnobs.rewrite_export_default` is a pure string transform, and a property it cannot find returns `""` — never the source unchanged, which a caller would happily save and call success. It carries a `: set = _x` suffix through (two of the three own a setter, and dropping it would leave a slider that moves nothing) and accepts both spellings of a default.

Two laws guard the same failure arriving by the other door: **every `ObjectKnobs` property is findable in its own script**, and **the scene overrides none of them**. If either stops being true, Save writes a line nobody reads and the panel reports success.

## Shared rather than forked

`DevWidgets` gains `add_knob_row` (LookTool's row builder, moved — both panels draw the same kind of row) and `literal_for` (both write the same literals, to different destinations).

## Verification

`tests/dev` 182/182, `tests/flow` 183/183 (the one non-dev reader of `KNOBS`). Falsified three ways, each red on exactly the aimed case:

- a miss returning the source unchanged → `test_an_unknown_property_returns_empty_rather_than_the_source`
- a renamed knob property → `test_every_object_knob_is_declared_in_the_script_its_node_carries`
- `cover_scale = 0.9` authored into `Battle3D.tscn` → `test_the_scene_overrides_no_object_knob_property`

## Your check

Objects tab → move **Cover bump scale** to the value you had → **Save to source** → `git diff` shows one changed line in `BoardMirror.gd`. Editor runs only, which is the whole population (`custom_features=""`, so no build carries dev tools).

Godot will notice the file change when you refocus the editor — if you have `BoardMirror.gd` open with unsaved edits, save or close it first.

## Not in this diff

- **Slice 2** — per-tile-TYPE fields off the TileSet's custom data (a lamp's own light, per-type height), with the global staying as the default an object may override, per your call.
- **Per-placed-instance** overrides — no store exists; a separate feature.
- `prop_light_*` — `@export`s read at prop-build time, so they need a rebuilding setter before they can be honest knobs. They belong with slice 2's light work.
- Canon sweep of `docs/design/presentation-effects.md` (it calls two of these "Look knobs") — at issue close, per the canon-sweep rule.


Phaazoid:
Second commit on this branch, from the feel-check.

## Prop block height was a slider that moved nothing

Two separate things were going on, and only one of them was correct behaviour.

**The Lantern is a BILLBOARD** — a camera-facing sprite with no mesh — so `block_height_scale` cannot touch it by construction. It only reaches `SOLID_SHAPES` (crates, chests, rocks, barrels, the pot, fences), and it is a **vertical stretch of the generated mesh**, nothing to do with a Z or a lift. That part is right; the label just didn't say it. Tooltip fixed.

**But it also had no setter**, since #264. The value was read at build time and props reconcile only when their tile changes, so it did not reach anything already standing — the exact lying-slider failure the Look-tab law exists to forbid. Its two siblings got setters in #280 (`tuft_scale`) and #326 (`cover_scale`); nobody went back for this one, which is why cover and tuft felt right and this didn't. It now sweeps the prop roots and re-stretches their `MeshInstance3D` children, which excludes billboards and tuft sprites *by type* rather than by a mark.

## Fire joins the Objects tab

Your call, and the measurement supports it: **all twelve shipped presets carried byte-identical flame values**, so nothing had ever been tuned per mission and leaving costs no authored content. All twelve knobs move, including the #217 accessibility switch and the #298 clearance — both reasons still hold and are now **structural** rather than `PRESET_EXCLUDED` entries, since a preset cannot reach a knob that isn't in that table at all. The eleven flame keys are stripped from each preset, so no load reports an unresolved key.

`flame_size:x`/`:y` forced a real seam: a **component** knob has no declaration line of its own (`x = 0.4` is not something a script can say), so Save asks `declaration_prop()` which line to write, and the two axis knobs collapse to the one they share. Copy Values reached the same answer for `.tscn` lines; this is that rule in the other destination.

**One consequence worth naming:** `flame_size:x` was the *only* component knob in `LookKnobs.KNOBS`, so LookTool's own whole-vector branch now has no reachable caller from its table. I kept it — it's the rule for the next vector look knob — and repointed its case to drive `_paste_split` with a synthetic knob rather than deleting the guard. A rule with no live caller is exactly the kind that rots unwatched.

`EFFECTS_NOT_MOOD` is down to a single label. That's the list shrinking because the ruling got *stronger*, and last round's names-a-real-knob law is what keeps it honest.

## Verification

`tests/dev` 184/184, `tests/presentation` 265/265. Both new guards falsified:

- removing the setter reds `test_the_block_height_knob_restretches_props_already_standing` with its own message
- a stale label reds `test_every_not_mood_label_names_a_real_effects_knob`

## Still open, your call

Look's **Effects** sub-tab now holds exactly one row — `brush_ghost_alpha`, dev chrome. Worth folding into Markup and dropping the tab (your original "it eats the Effects tab", arrived at by a different route), but I left it alone rather than restructure unasked.

Phaazoid:
Slice 2 — the half the issue is named for. An object type now carries its own light and its own size, with the global as the default it overrides.

## The finding that shaped it: per-type light was already built, in the wrong place

`BoardMirror.LIT_PROPS` was a hardcoded `["Lantern"]` matched against a tile's **display name**, and its own comment said the quiet part out loud — *"which props glow is content."* So this doesn't add per-type light beside it; it **retires it into the column that should always have held it**. `LIT_PROPS` is deleted.

The plan flagged this as the riskiest assumption. It's now measured, not assumed: **exactly one tile carries that name**, so the migration is 1:1 and nothing else was silently being lit.

## The shape

Seven custom data layers on `TestTiles.tres`, **declared in this diff rather than added at runtime** — a schema migration that runs once on one machine is a path no test executes again. `ObjectKnobs.FIELDS` is the table: layer, type, which `PropShape`s it applies to, and the global it falls back to. That's what makes a new field a row, and what makes a meaningless row impossible to draw rather than merely hidden — a tuft is never offered a block height, and the four light rows only appear once a tile says it emits at all.

## One thing changed from the approved plan

**The sentinel is `0`, not `-1`.** I stopped and asked when the storage disagreed with the plan: `TileData.has_custom_data` answers whether the *layer* exists, never whether *this tile* authored a value, so an untouched float arrives as the type's own `0.0`. Any other sentinel would have to be hand-authored onto every field of every object tile, with nothing enforcing it for the next tile added.

The cost is that a literal zero stops being authorable, and for these six fields that costs nothing real: a prop scaled to 0 is invisible, and a lightless light is what `prop_lit = false` says better.

Colour taught the same lesson one type along, and I only found it by running the test — **a `Color` layer defaults to opaque black, not transparent**, so blackness is its sentinel and alpha isn't part of the question.

## Two things that would have been silent bugs

**`drop_props()` exists because `_reconcile_prop`'s diff key is the TILE.** A prop whose tile is unchanged is deliberately left standing — right for painting, and blind to the tile itself gaining a field. That's exactly the #308 shape (a poll drawing through a store its own diff key can't see) and it takes the same answer: invalidate at the source rather than widen the key into a copy of everything the build reads.

**Both global setters now re-derive from each prop's own override** rather than writing the new global flat. Without that, touching a global slider would silently stomp an authored height and the board would disagree with the tileset until the next repaint. The override rides the node as meta, because the sweeps have no grid to ask — the mirror is handed one per call and stores none, deliberately.

## Effects tab, folded

Its whole population left in slices 1 and 1b; the one survivor is `brush_ghost_alpha`, re-filed as **Dev chrome** and sharing the Markup sub-tab. A tab holding a single row has stopped meaning anything. Your original "it eats the Effects tab" is now literally true.

## Verification

`tests/dev` 197/197, `tests/presentation` 266/266. Three mutants, each red on exactly its aimed case:

- dropping the `prop_lit` authoring → the migration law reds (*"the light that used to come from LIT_PROPS is gone"*)
- making the resolve ignore overrides → the override case reds
- no-opping `drop_props` → the live case reds (*"a changed field would never reach the board"*)

One incidental fix: the old light test asked for `"Lantern"` and `"Tree"` **by authored name** — a content-razor violation that predates this work. It now derives the lit/unlit pair from the column.

## Your check

Objects tab → pick the lantern → untick **inherit** on Light energy → it changes on the board and nothing else does → **Save object fields** → `git diff` shows the value in `TestTiles.tres`. Then confirm a crate is unmoved by the lantern's number, and that the lantern still glows at all on a fresh load — that last one is the migration, and it's the thing a law can't check for you.

## Not in this diff

Per-**placed**-instance overrides ("*this* lamp is dimmer") — no store exists; a sparse `ScenarioData` dict and its own feature.

Issue #357, opened 2026-08-18
Element states are invisible on the board: an icon row above the health bar


Phaazoid:
*Filed by Claude from the dev's direction in conversation (2026-08-18), not written by the dev directly.*

## The ask

Dev, ruling on #346's open half: *"I want to represent what element states units have more visually. At the very least, I want an icon for what states a unit has to appear just above their health bar, centered left. Those elemental icons will work fine for that for now."*

So: a row of state icons anchored just above each unit's `UnitHealthBar`, left-aligned to the bar's left edge, growing rightward, reusing the `StateIcons` art as-is. This is the first occupant of the head channel #346 freed — the fact it carries is *what this unit IS*, per the rule in `docs/design/visual-clarity.md`.

## Where it lands

- `UnitMirror` (#229) already builds one `UnitHealthBar` per mirrored unit, anchors it by measuring the art, and faces it once per frame. The icon row joins that cluster — children laid out in ordinary local space above the bar's quad, inheriting its facing, per visual-clarity's *one display means ONE billboard* finding. It does not get its own billboard basis.
- `StateIcons` (`Classes/ui/panels/StateIcons.gd`) stays the single source of truth for the art — this row is its third consumer after the hover card and `UnitStatesBar`. New state art lands in `StateIcons.ICONS`, never here.

## Visibility — one gate, not two

The bar cluster's visibility is #350's business (always-on unless the player toggles it off, dev direction 2026-08-18), and the icon row rides the same gate. If this ticket grows its own visibility rule, that is the bug — the same sentence #350 already says about the bar. When bars are toggled off, the persistent board read for element states is the sprite-effect channel (#358); until that ships, the hover card is the fallback.

## Known gaps, accepted by the dev

- **CHILLED has no icon.** `StateIcons.ICONS` carries WET alone; CHILLED falls back to a text label, which will read rough as a world-space row. Dev accepted the current art "for now" — when a 16px CHILLED icon exists it lands in `StateIcons.ICONS` and all three consumers pick it up at once.
- **Crowding.** Today's vocabulary is two states (WET, CHILLED), so the row cannot crowd. #229's deferred crowding question re-triggers when the state list grows or the row takes more tenants (downed/maimed already have icons in `Art/Icons/StateIcons/`).

## Constraints

- #354 is in flight in `UnitMirror._sync_bar` (latching predicted bars through the pass) — coordinate, don't collide.
[...]

Issue #350, opened 2026-08-16
Let the player pin every health bar on, not just the hovered one


Phaazoid:
🤖 Claude says:

**Dev ask, 2026-08-16, at [#313](https://github.com/Phaazoid/Godoiosis/issues/313)'s merge:** an in-game button the player can push to toggle every unit's health bar to always-visible, instead of only-on-hover.

## What this is

[#229](https://github.com/Phaazoid/Godoiosis/issues/229) put a health readout in the volume above a unit and made it **hover-only, by dev call**. #313 gave it a second reason to be up — the queued plan is about to change that unit's HP. This is a **third** reason, and unlike the first two it is a *player preference* rather than a derivation: show them all, all the time, because I want to read the board at a glance.

## Why the model half is nearly free

`UnitMirror._sync_bar` already computes visibility as one expression — `hovered or foretold`. This adds one more disjunct. There is no new source of truth, no new per-unit state, and nothing to derive: the bar is already built for every mirrored unit and simply hidden. **If this ticket grows a second visibility rule, that is the bug.**

Two knobs it can reuse rather than reinvent: `UnitMirror.ghost_shows_number` already exists for exactly this crowding question one level down (a prediction-only readout is bar-only by default, and hovering reveals the digits). An always-on mode almost certainly wants the same treatment, and possibly wants to *share* that knob rather than add a twin.

## The questions it raises

1. **Where does the button live, and the answer is not obvious, because the game has no options menu.** [#217](https://github.com/Phaazoid/Godoiosis/issues/217) (photosensitivity toggle) already says so and explicitly reserves being that menu's first tenant. So the live options are: a HUD affordance on the board, a pause-menu row, a keybind, or **wait and land both toggles together on one settings seam**. That last one is the Law #4 answer if an options pass is anywhere near — two player toggles built two ways, weeks apart, is how a settings surface ends up with no owner.

2. **A keybind has a trap with a name.** *A keybinding lives in ONE OS window and is DEAD in the other* — canon in `CLAUDE.md`, and it has bitten five times. The game and `DevOverlay` are two real OS windows, so a bare handler is silently unreachable whenever the other window has focus. If this becomes a key, it forwards through one public entry the way `DevController.handle_dev_key` does; it does not get a second handler.

[...]

Issue #346, opened 2026-08-16
Two marker channels, one rule: the ground says what you are doing, the head says what a unit is


Phaazoid:
*Filed by Claude from the dev's playtest notes (2026-08-16), not written by the dev directly.*

## The ask

Dev, testing #316's ground marker: *"This thing you've implemented actually overlaps with another feature we already have — the target reticules that pop over unit's heads for different things. But funny enough, I think this is actually more readable, and is a good direction to go for 325 as well. Perhaps above the head stuff can be saved for elemental statuses, etc."*

So: **two marker channels, and a rule for which fact goes in which.** The ground says what the *interaction* is; the head says what the *unit* is.

## The overlap is literal, not conceptual

`game.create_squad` calls `draw_create_squad` — which puts an `IconType.TARGET` billboard over every eligible recruit — and then `enter_target_pick_mode`, which paints `TARGET_ATLAS_COORDS` on those same units' cells. Since #316 made the ground half visible, Squad Up marks each candidate **twice**, in two visual languages, for one fact. That is Law #4 with both answers now on screen at once.

## What the head channel was actually doing

`IconType.TARGET` is created in exactly **one** place — `game.gd:947`, the Squad Up flow. Rescue, Intimidate and Join Squad all ride the same `enter_target_pick_mode` and never got it. So before #316 those three modes marked their candidates **nowhere, in either view**, and Squad Up looked fine only because someone had patched that one flow with head icons.

The head reticle was never the general answer to "which unit may I pick". It was a local fix for one screen, and the general answer — the ground marker — was the broken one nobody could see.

## Why freeing the head is worth something

A unit's element states are **hover-only** today: `hover_info_grid_container` calls `StateIcons.populate(states_row, unit.element_states)`, and nothing else draws them. On the board, a Wet unit and a dry one are identical, so the player has to interrogate units one at a time to find the fact that decides whether a Lightning carving is worth firing. That is the content the head channel is well suited to and the ground channel is not — it is true of the *unit*, it persists across modes, and it should be readable on every unit at once without hovering.

## The rule to decide

- **Ground / tile** — what this interaction is about. Transient, mode-scoped, cursor-driven: pickable candidates, reach, selection, the cells a thing would affect.
- **Above the head** — what this unit *is*. Persistent, mode-independent: element states, downed / maimed, Crisis, and whatever later joins that list.

Each existing `OverlayIcon.IconType` then needs a home, and these are dev calls, not obvious:

[...]

Issue #292, opened 2026-08-15
2D/3D parity (umbrella): keep the flat view a real alternative, not a degraded fallback


Phaazoid:
🤖 Claude says:

Umbrella / tracking issue for **keeping the flat 2D view and the 3D diorama at parity**, so the question lives in one place instead of being re-answered ad hoc by whichever ticket trips over it next.

**Core problem (@Phaazoid, 2026-08-15, right after #285's feel-check):** the game has two views and features keep landing in only one of them. Each divergence was individually reasonable; the aggregate is that 2D is quietly becoming a degraded fallback rather than a real alternative.

## The doctrine this exists to enforce

Two rulings already on the record, both of which this umbrella is the bill for:

- **Dev ruling on [#218](https://github.com/Phaazoid/Godoiosis/issues/218) (2026-08-14):** elevation is **not** a 3D-only feature — the flat view has to render and play it. That sharpened the earlier *"a Z level is just a variable"* line from *2D survives elevation* to **2D must render it, and here is the art cost.**
- **`CLAUDE.md` on FLAT_2D:** it is *"the real escape hatch and the embryo of #176's 2D/3D chooser"* — it hands input back on purpose, because *"a flat 2D game that cannot take its own clicks is not a playable fallback."*

## The sharpest current instance

**`Level_1` now has painted elevation, and the 2D view renders it flat.** The rules read `BoardHeights` either way — `can_step`, movement cost, cohesion — so the flat view is currently *rules-correct and visually blind*: you can play a map whose heights you cannot see, on a board that looks like it has none. F5's `HeightDebugOverlay` exists only to paper over exactly this, and it was declared throwaway scaffolding when it shipped.

That one is [#266](https://github.com/Phaazoid/Godoiosis/issues/266) and it is the flagship child here. It is blocked on a tileset decision, not on code.

## The forward-looking rule, which is the actual point

> **When a feature lands in one view, the other view's answer is part of the ticket** — even when that answer is "deliberately nothing".

A declared asymmetry is a design; an undeclared one is drift. [#285](https://github.com/Phaazoid/Godoiosis/issues/285) is the model: its ghost is 3D-only, and the reason is stated at the source (the 2D ghost is a `TileMapLayer` holding a tile pick, and a height is not a tile) rather than left to be rediscovered. That is Law #4's *"declare it out loud"* applied to the two views.

## 👉 The list is the first comment below

[...]

PR #372, opened 2026-08-19
#325 verdict: the crown goes back over the head, the rings stay

### LINKED ISSUE #325: Squad membership reads as a green square: give it a 3D-native form
*Filed by Claude from the dev's playtest notes (2026-08-16), not written by the dev directly.*

## The ask

Dev, playtesting v0.39.0: *"We should change the effect of showing who is in a squad (currently green square over head) to something more fitting in the 3D, perhaps a light effect or something."*

## What is there now

`OverlayIcon.IconType.SQUADMEMBER` -> `res://Art/Icons/BoardIcons/SquadHighlightIcon.png`, created per member by `OverlayManager.redraw_squad_icons` alongside the leader's `CROWN`. `OverlayMirror._icons` mirrors the whole `icons_by_unit` table into `BoardOverlays.Layer.ICONS` as camera-facing billboards, with a per-type height stagger so co-celled icons do not z-fight.

So the 3D is faithfully reproducing a 2D affordance: a flat green square, drawn in screen space above a head, in a scene that is otherwise lit and volumetric. It reads as a HUD sticker pasted onto a diorama, which is the complaint.

## What makes this one interesting

Squad membership is not a *marker* on a unit, it is a *relationship between* units — which is exactly the kind of fact a flat per-unit icon is worst at expressing and a 3D scene is best at. Candidates, none picked:

- **A light effect** (the dev's own suggestion) — a tinted rim, a ground pool, or a subtle emissive under each member. Ties the members together by sharing a colour, and it is the one option that reads at a glance without adding a second thing to look at.
- **A ground ring / footprint tint** on each member's cell. Cheapest, uses the fill machinery already there (`Layer.SQUAD` exists), and stacks legibly with terrain because it lies on the surface rather than floating.
- **Tethers** — a line from each member to the leader. Says "relationship" most directly of anything, and it is the only option that also shows the cohesion leash, which is currently invisible until a move is refused.

Worth noting that `Layer.SQUAD` and `Layer.SQUAD_RANGE` fills already exist and already draw the squad's footprint and range in orange — so part of the answer may be that the icon is redundant with a fill the player is already looking at, rather than needing a replacement at all.

## Constraints

- **The 2D is the authority** for markup (`OverlayMirror`'s whole doctrine: cells, textures and tints arrive by copy). A 3D-native form that has no 2D counterpart is the first markup that breaks that, so it needs declaring on **#292** rather than quietly diverging — a declared asymmetry is a design, an undeclared one is drift.
- Anything emissive lands in the same band as the carried-light-sources ticket filed alongside this one; they should not each invent their own idea of what a unit-attached light is.
- Whatever ships gets Look-tab knobs, per the tuning rule.

Related: #292 (2D/3D parity umbrella), #44 (visual clarity umbrella), #176.

Phaazoid:
Closes #325. Your verdict, built: the rings stay for membership, the **original crown goes back over the head** for leadership (hover any squad member, the leader wears it — unchanged from before the ticket), the bar badge is gone, and the legacy green squares are deleted outright.

## Mostly deletion, and the crown's production path was never the problem

`game.get_squad_icons` has produced `[SQUADMEMBER, CROWN]` for the leader the whole time, in 2D, untouched by #325. What the experiment did was stop the **3D mirror** from carrying the crown — one `continue` in `OverlayMirror._icons` — and badge the health bar instead. So this is that `continue` removed, the badge removed, and the mode fork removed.

Net **-55 lines**.

## Things worth ruling on

1. **`OverlayMirror._icons` routes by TYPE, not by mode.** CROWN → head channel, everything else → ground. The per-type y-stagger went with the squares: it existed to keep several head-icon types off each other's plane, and CROWN is that channel's only tenant now.

2. **`ICON_TEXTURES` became the ONE answer to what a marker type looks like** — applied once at `setup`, so `_style_icon` now only decides tint and z. This is a **fix I found by breaking it**, not a tidy-up: my first pass retired the `SQUADMEMBER` entry from that table on the grounds that `_style_icon` overrode the texture anyway. It has a second reader — `create_unit_icon`'s `icon.setup(ICON_TEXTURES[type], …)` — and the suite went red immediately. Law #4's "count all callers" catching me in the act; the corrected shape is strictly better than what I planned, because the texture now comes from exactly one place instead of two.

3. **The leader wears both a ring and a crown.** A ring because they are a member, a crown because they lead. That is this section's two-channel rule working, not an exception to it — the ground says what the interaction is about, the head says what the unit is.

4. **`SQUAD_MARKER_RINGS` is gone**, with `_style_icon`'s fork and the Look tab's checkbox. **Ring opacity stays** — it tunes a shipped feature rather than picking between two designs. `SquadHighlightIcon.png` is now unreferenced; I left the file on disk rather than deleting art with its `.import` sidecar in this diff.

## Why the badge lost, recorded in canon

It read fine in isolation. What killed it is that it made **leadership conditional on a different marker's visibility** — and #350 landed a day later making the health bar a player preference that is *off by default*, so a squad leader silently had no persistent 3D marker at all. Neither ticket was wrong alone; it existed only in composition, which is why I filed it on #44 rather than as a bug against either.

**The rule that falls out, now in `visual-clarity.md`: a marker answering "what is this unit" must not ride another marker's visibility gate.** That is also the standing caution for #357, whose state-icon row is specified to ride exactly that gate — riding it is a choice about defaults, not a free consolidation.

## Verification

- `tests/presentation` + `tests/dev` **473 cases**, `tests/ui` + `tests/squad` **332 cases** — all green, 0 failures, 0 orphans. Full tree is CI's.
- **New case pinning what you actually asked for**: hover a **non-leader** member → the crown appears over the **leader** (asserted by position, not just by count) → look away → it clears. Nothing asserted that end to end before, because every existing case drew the icons by hand and so could not tell "produced in 2D" from "mirrored into 3D" apart.
- **Falsified**, and it needed isolating: restoring the `continue` reds an *earlier* case first, which truncates the file — so I disabled the earlier cases one at a time and confirmed **both** new cases red on their own, with the right messages. That wire (2D icon → 3D mirror) is the #103 shape and the only place a green suite could go blind here.
- The three #350 always-on cases are untouched and still green — the regression guard that this deletion did not disturb that gate.
- **Yours**: crown height over the head, the ring-plus-crown stack on a leader, and whether the crown reads at low camera pitch.

## Canon

`docs/design/visual-clarity.md`'s #346 section rewritten to the settled rule. I kept the genuinely good part of the earlier #325 sweep — the #228 correction that `Layer.SQUAD` was never a membership marker — since that is true regardless of which style won. The `#292` asymmetry that sweep declared is **gone**, not moved: both views draw the same head crown now.

### DIFF
--- a/Classes/board/OverlayManager.gd
+++ b/Classes/board/OverlayManager.gd
@@ -91,19 +91,21 @@ enum OverlayType {
 	INVALIDMOVE
 }
 
+# The ONE answer to what each marker type looks like, applied once by create_unit_icon's setup --
+# so _style_icon below never touches a texture, and there is no second place a marker's art can come
+# from. Membership is a per-squad ring underfoot, leadership the crown over the head (#325, settled
+# 2026-08-19); the legacy green square lost, and SquadHighlightIcon.png is now unreferenced.
 const ICON_TEXTURES = {
 	OverlayIcon.IconType.CROWN: preload("res://Art/Icons/BoardIcons/CrownIcon.png"),
-	OverlayIcon.IconType.SQUADMEMBER: preload("res://Art/Icons/BoardIcons/SquadHighlightIcon.png")
+	OverlayIcon.IconType.SQUADMEMBER: preload("res://Art/Icons/BoardIcons/SquadRingIcon.png")
 }
 
-# --- Squad marker style (#325 experiment) --------------------------------------------------
-# Rings underfoot vs the legacy over-the-head squares. static, the ATTACK_MODULATE pattern, so
-# the Look tab's toggle reaches it without a node ref; read by _style_icon here and by
-# OverlayMirror._icons, so flipping it moves BOTH stacks. Style only -- icon LIFECYCLE (when a
-# marker exists) is untouched either way. The losing style gets deleted when the experiment ends.
-static var SQUAD_MARKER_RINGS := true
+# --- Squad markers (#325, settled 2026-08-19) ----------------------------------------------
+# The dev played both styles and took a MIX: membership is a per-squad coloured RING underfoot,
+# leadership is the ORIGINAL crown over the head. The legacy green squares lost and are deleted --
+# no toggle, no second style. SQUAD_RING_ALPHA stays a live Look-tab value; it tunes a shipped
+# feature rather than picking between two.
 static var SQUAD_RING_ALPHA := 0.9
-const SQUAD_RING_TEXTURE := preload("res://Art/Icons/BoardIcons/SquadRingIcon.png")
 # Per-squad hues, dealt lazily by SquadManager when a squad first gains a squadmate: cool for
 # friendly squads, warm for enemy ones, so "which squad" and "whose side" read from one glance.
 # Plain consts, one per line -- edit freely; WHITE is reserved as the not-yet-dealt sentinel.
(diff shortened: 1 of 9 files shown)


Phaazoid:
**Conflict resolved — merged `main` in at `01532bc`.** #357 (the element-state row) landed while this was open, onto the same three files, which was structural rather than bad luck: both tickets ride `UnitHealthBar`, so #357's row was written *directly adjacent* to the crown badge this PR deletes.

Resolution in every case was the same — **keep #357's row, drop #325's badge**. Four conflicted files (`UnitMirror`, `UnitHealthBar`, `LookKnobs`, `visual-clarity.md`) plus one that auto-merged (`test_unit_health_bar.gd`), which I checked by hand rather than trusting: a clean auto-merge is exactly where an adjacent line goes missing quietly. All 15 of its cases are present — 6 from #229/#313, 3 from #350, 6 from #357 — and only my 3 badge cases are gone.

**One thing worth flagging, because it is the kind of stale that survives a green suite.** #357's code and canon described themselves *relative to the badge*, since the badge was the thing beside them at the time:

- `# pixel art, like the badge`
- `# the same reference the crown measures from, so the two marks share a silhouette`
- *"Everything else follows the crown badge #325 put on the same bar"*
- and a knob comment justifying icon sizing as *"the crown is bar-height because the dev asked for it AT bar height"*

Each is now a comparison to something that does not exist. All four are rewritten to stand on their own — the sizing rationale in particular is now stated positively (nothing ties a status icon to how thick the gauge is) rather than as a contrast with the crown.

**Re-verified after the merge**: `tests/presentation` + `tests/dev` + `tests/ui` — **662 cases, 0 failures, 0 orphans**, #357's six new cases and this PR's two green side by side. `git diff main...HEAD --stat` shows only this ticket's 9 files, so nothing of #357's was dragged in or reverted.

Phaazoid:
**CI is red on this PR and it is NOT this PR — `main` is red the same way.** Recording it here so the check isn't misread as a merge-resolution mistake.

- **This branch**: 1738 cases, **1 failure**, 432 orphans — `tests/presentation/test_input_bridge.gd > test_releasing_the_button_ends_the_brush_drag`.
- **`main` at `4d07237`** (the #370 merge): 1741 cases, **1 failure**, 432 orphans — the *same case*, the same orphan count. The last green run on main was `b293a617`, one merge earlier.
- **It passes locally on this branch**: that suite alone is 28/28, 0 orphans, on Windows.

So the branch inherited it by merging `main` in, and the orphan count is identical on both — nothing here introduced either.

## The lead, offered as a hypothesis rather than a finding

The failing assertion is a **precondition**: *"the press never armed the drag, so the release proves nothing"* — the brush never entered its painting state, so the case aborts before testing anything.

`test_input_bridge`'s fixture **loads `Prolog`** (line 39), and #370 is *"The Prolog squad lesson: authored steps, beats, and Torv's voice"*. That is a direct overlap: the suite now boots a board that has a tutorial authored onto it. If a lesson beat opens a prompt that claims the modal lock, the Game subtree freezes and a board click cannot reach `dev_controller` — which is exactly "the press never armed the drag".

**Why CI-only fits that**: it would be a race, not a logic error. `Dialogic.start()` is a frame late, so whether the beat lands inside or outside the test's `_pump()` window depends on machine speed — green on Windows, red on the CI runner. That also explains the 432 orphans appearing in the same run and not locally.

I have **not** verified this; I stopped at the overlap because #370 is someone else's ticket in flight. Whoever picks it up: the cheap first check is whether `ScenarioDirector` fires on the `test_input_bridge` fixture at all, and if so whether the suite should disarm it in `before_test` the way the bar suites disarm `PlayerSettings`.

Issue #218, opened 2026-08-13
Verticality design brainstorm: what elevation brings to the rules (lob vs block, ramp costs, LoS, Triangle Strategy study)


Phaazoid:
🤖 Claude says:

## What this is

The dev's requested design session (2026-08-13): *"we should probably do a whole brainstorming session on what verticality brings to this game. Including looking at Triangle Strategy."* The HD-2D board (#176) gives elevation a home for the first time; this session decides what the RULES do with it. Grill-style when it happens — this issue is the agenda seed so nothing on it evaporates.

## Seeds already on the table (the dev's, 2026-08-13 unless noted)

1. **Ramp movement costs**: going up costs more than flat; going down costs less. ("File that away as a thought.")
2. **Lobbed vs. blocked ranged attacks**: arcing attacks (fireballs, water) clear walls; flat-trajectory attacks (guns) are blocked by them. Hook: the weapons layer already reserves the `arcs_over_obstacles` flag (CLAUDE.md's targeting notes) — this is its design session.
3. **Fall damage scales with height, weight as a multiplier** (2026-08-12, recorded on #116) — softens the original shove-off-height kill doctrine; interlocks with #120's weight-modified knockback.
4. **The traversal ruling already in force** (2026-08-12, presentation demo + future sim rules): climbs happen only via ramps; ramps connect exactly their high/low sides, no sideways entry.
5. **30° two-tile slopes** (presentation experiment, noted on #176) — if adopted, traversal and cost rules must cover them.

## Questions the session should settle

- Height advantage/disadvantage for attacks (damage? reach? the backstrike-facing seam's vertical sibling?)
- Line of sight: does elevation block, reveal, or ignore? (Interacts hard with #2.)
- Do move ranges pay per-step height costs (#1) or flat costs with ramp gating only?
- Shove interactions: off ledges (#116), up/down ramps, into walls.
- **Triangle Strategy study**: how the tactical sibling telegraphs all of the above at a distance (already shelved in `docs/design/presentation-effects.md`'s reference section).
- What the AI needs to *see* of all this (#117's standing lesson: a feature the AI can't use degrades playtest signal).

## Related

#176 (the board that renders it), #116 (shove/fall), #120 (weight), #209 (replays would capture great verticality moments), `docs/design/presentation-effects.md` (the reference shelf).

— Claude (Fable 5) · 2026-08-13

Phaazoid:
🤖 Claude says:

[...]

Issue #120, opened 2026-07-30
Give Weight teeth: shove resistance, push tiers, and knockback distance


Phaazoid:
🤖 Claude says:

Filed from `docs/SCRATCHPAD.md` (sweep 2026-07-29), out of a @Phaazoid ↔ @c3potheds conversation
about shoving units into water. The water half is captured on
[#116](https://github.com/Phaazoid/Godoiosis/issues/116); **this issue is the weight half**, because
it is the first concrete proposal for the question `CLAUDE.md` currently records as open:

> **Weight is tracked but feeds no rule, on purpose.** Live options are **wire it to something** or
> **cut it entirely**; both are open (dev, 2026-07-27).

The dev, in the conversation: *"I was also just thinking this could be a way to make weight mean
something."*

## The proposal, in the dev's own terms

Two mechanics, both reading Weight, plus the tradeoff they create together:

1. **A weight/push tier gate.** *"could have a weight/push level thing, where like, the weakest
   tier 1 wind could only work on units of up to a certain weight."* A shove effect has a power
   tier; a unit too heavy for that tier does not move at all.
2. **Weight modifies shove distance.** *"weight could potentially determine how many tiles a unit
   gets knocked back, too."*
3. **The tradeoff that makes it a real axis:** *"heavier units can't move as far, but lighter units
   are more susceptible to getting shoved around."*

**Clarified by the dev 2026-07-29 (correcting an earlier misreading of this issue): "can't move as
far" means *pushed* as far, not MOV.** This has nothing to do with encumbrance or the unwired
Weight×MOV step — the whole trade lives inside knockback. The concrete shape, in the dev's words:
**certain weight bands lower knockback from ALL sources by 1 per band.**

That is a much cleaner mechanic than a two-system tradeoff:

- **One term, one place.** Resolved distance = the attack's authored `knockback` − the target's
  weight-band reduction, floored at 0. It folds into `_source_knockback`/`_resolve_knockback` and
  nothing else moves.
- **All sources, no exceptions** — not per-family, not per-attack. A band is a property of the target,
  so it reads the same against wind, a mace, or a water jet.
- **The trade is self-contained:** heavy gear buys shove resistance and costs you inventory room and
  whatever else weight eventually gates; light units stay mobile and get thrown around. No MOV
  involvement is required for it to be a real choice.
- **It composes with the tier gate** (1) rather than duplicating it — a tier gate is "does this move
[...]

Issue #116, opened 2026-07-29
Shove units off cliffs to kill them: knockback into a killing drop (airships, trains, cliffside cities)


Phaazoid:
🤖 Claude says:

Filed out of [#115](https://github.com/Phaazoid/Godoiosis/issues/115), where the knockback half of
"may THIS unit end up here?" turned out to be blocked on a mechanic that doesn't exist yet. Recording
the dev's framing (2026-07-29) so the intent survives.

## The mechanic

**Shoving a unit off a height kills it outright.** Not damage — removal. This is where knockback's
power budget has always been headed, and it is the reason shove effects are costed the way they are.

Dev, verbatim in spirit: *"eventually we're going to have levels in airships, on moving trains, and
in cities on a cliff side. These are all excellent locations to shove units off of high things, never
to be seen again."*

## Why it is filed now rather than built now

`PlanResolver._resolve_knockback` currently stops a shove at the first cell that fails
`BoardContext.is_walkable` — a wall, the board edge, or an occupant. That is a **safe placeholder,
not a decision**, and #115 has now declared it as such in a comment at the call site so nobody
"fixes" it by quietly swapping the predicate. Today an unwalkable cell simply stops the shove; under
this mechanic some unwalkable cells must instead *consume* the unit.

## Power gating — the dev's model

Two levers, both deliberately outside the rules engine:

1. **Map design.** Ledges are placed, not procedural. A level with no drop has no instant kills.
2. **Knockback distance.** Most things shove 1 tile. Reach to a ledge is the scarce resource.

Worth stating because it means the mechanic needs **no balance knob of its own** — no damage number,
no save-vs-death, no percentage. It is binary, and its frequency is authored.

## What has to be decided before building

- **What marks a killing drop?** The candidates are a `Terrain.Kind` (a `CHASM`/`VOID` kind), a
  `Terrain.TileState`, or a new tile custom-data flag. Note the shape already exists twice: #109 just
  established that `BoardContext.is_walkable` is the single reader for "may a unit stand here", and a
  chasm is a cell that answers **no** for a reason that matters. Law #4 applies — this should extend
  the existing terrain vocabulary, not open a parallel one.
- **Does it read as a kill or as a removal?** `Unit.die()` vs something new. It interacts with Will,
  Crisis, permadeath (`UnitInstance.permadead`), and rescue — a body at the bottom of a chasm cannot
  be rescued, which is exactly what makes it different from a normal down.
[...]

PR #428, opened 2026-08-20
Falls, shoves and tumbles: the airborne shove (#259)

### LINKED ISSUE #259: Verticality slice 3: falls, shoves and tumbles
🤖 Claude says:

**Slice 3 of 3 from the #218 verticality design.** Spec: [`docs/design/verticality.md`](../blob/main/docs/design/verticality.md). Needs slice 1's height store. Closes the interlock with [#116](https://github.com/Phaazoid/Godoiosis/issues/116) and [#120](https://github.com/Phaazoid/Godoiosis/issues/120).

## Two kinds of edge

- **A vertical drop** deals **fall damage**, scaled by height fallen and modified by weight (#120).
- **A void** — chasm, airship edge, train side — is **removal**, not damage. #116's original kill doctrine, intact.

These reconcile rather than compete: a void is a drop with no floor. The cliff city gets both, and #116's power-gating argument survives untouched — the level designer decides which edges exist, so the mechanic still needs no balance knob of its own.

## Shoves

- **You cannot be pushed uphill.** A shove that would climb stops dead; high ground braces you.
- **Only vertical drops give fall damage.** Slopes never do.
- **A shove onto a descending ramp becomes a tumble.**

## The tumble rule

> A shove that steps onto a descending ramp keeps descending while the next cell is another ramp continuing down, and ends on the first flat or level cell it can legally enter. A wall, a rise, or an occupied cell stops it where it stands. No fall damage.

```
shove →
  F3   R2   R1   R0   F0        one unbroken flight: tumbles F3 → R2 → R1 → R0, ends on F0
  F3   R2   F2   R1   R0  F0    landing at F2 catches it: ends on F2, never reaches R1
```

**A landing stops the tumble** (dev: *"it makes visual sense for flat tiles to stop a tumble"*), and that is the point — it hands the level designer the dial. One unbroken flight down a cliff-city terrace is a genuine hazard; a landing halfway makes it survivable. Verticality's danger stays a map-design decision rather than a number anyone has to balance.

Two consequences, both ruled the conservative way by the dev:

1. **The tumble is free — it does not spend knockback distance.** Otherwise a 1-knockback weapon could never push anyone down a flight, which is the whole image. Knockback moves its N cells normally; if a step lands on a descending ramp, the slide is a bonus, and the tumble **ends** the shove rather than resuming the remainder.
2. **The tumble stops at a lip, it does not launch you off it.** If a flight bottoms out at a sheer drop, the unit stops on the lip. Fall damage stays restricted to shoves that push a unit *directly* over a vertical edge, so the two never compound into a two-stage prediction. Tumble-then-plummet is a fine later addition once the base case has been felt.

## Where it goes

`PlanResolver._resolve_knockback` already publishes `knockback_from` / `knockback_to` and threads the new position into the hypo, so **a tumble is just a longer `knockback_to`** — no new channel, and the queue preview shows the landing cell for free.

**The existing declared placeholder at that call site stays declared and stays correct**: a shove asks the CELL-level `BoardContext.is_walkable`, never the per-unit `RulesService.can_traverse`, so a Waterwalker is not shoved onto water. Being thrown is not walking. Do not "fix" it by swapping the predicate.

**The fall must resolve inside the resolver, not at playback** — a shove off a ledge is a damage and lifecycle event, and the queue preview has to show it (Law #2).

## Note

A chasm/void should extend the existing terrain vocabulary (`Terrain.Kind` / `TileState`), **not** open a parallel one — #109 made `is_walkable` the single reader and Law #4 applies. This is already recorded on #116.

Still open on #116 and **not** resolved by this ticket: what a shove into **water** should do. The dev has deliberately left it unpicked. Do not resolve it unilaterally.

Design: [#218](https://github.com/Phaazoid/Godoiosis/issues/218).

Phaazoid:
Closes #259 — the last slice of the #218 verticality arc: falls, shoves and tumbles, built on the airborne model you ruled this morning.

## The model — one flight, one landing

> "The drop occurs where the shove would move the unit to. This means you could potentially blow allies over holes to safety. Also, when a unit lands, if they land on a slope, they tumble down that too."

- **Flight**: the unit travels the knockback distance at its starting elevation. Higher ground braces it (no push uphill, as before), **VOID cells are flown over**, and walls/water/bodies stop it exactly as today — the water shoreline stop is deliberately untouched (#116's water fork stays open, pinned by its own test).
- **Landing** (wherever horizontal travel ends): on a VOID → **removed** ("Into the void!" — KILLED rung + `removed`, so the AI, queue and lifecycle all light up unchanged; both executors call `die()` since a 0-damage hit can't kill). Lower → **fall damage** ("Fell N!"), which **bypasses DEF** and joins before the Iron Will cap, so the cap stays absolute. On a ramp → **tumble down the slope's own downhill**, after paying any drop — the doc's original connected-ramp entry stays a free tumble.
- The path can now bend once (perpendicular-ramp landing), so `ResolvedOutcome.knockback_path` carries the whole trail and the preview arrow draws it — including across a hole. Counter shoves are previewed now too (pre-existing Law #2 gap, closed).

Both canon worked examples from `verticality.md` (F3 R2 R1 R0 F0 and the caught-at-F2 variant) reproduce verbatim as test cases.

## Placeholders to tune (`FallRules`, both `static var`)

- `FALL_DAMAGE_PER_LEVEL = 2` — a 3-terrace drop = 6; canon's "a 20-drop is lethal" property holds (40 ≥ the overkill ceiling).
- `WEIGHT_PER_BONUS_DAMAGE = 10` — the #120 weight hook, live but inert while every authored weight is 0. #120 stays open for push tiers, distance bands, and the weight-authoring pass.

## Embedded-content sweep (the new standing rule, applied)

Nothing to sweep: the new `ResolvedOutcome` fields are runtime-only (never serialized), `Terrain.Kind.VOID` is an append-only enum member, and the two hole tiles in `TestTiles.tres` are authored in this diff — the canonical file IS the change. No committed fixture embeds a copy that could go stale. The rule itself is now written into `CLAUDE.md`.

## Verification

- `tests/law` + `tests/play` + `tests/squad`: **291/291**; `tests/ui` + `tests/elemental` + `tests/ai` + `tests/presentation` mirror file: **340/340**. CI owns the full tree.
- New `tests/law/test_falls.gd` (16 cases): brace (dead stop + mid-flight halt), blow-an-ally-over-a-hole-to-safety, halted-over-the-hole drops in, ending-on-hole removes, scaled fall damage (control-compared, DEF-bypass by construction), a drop that changes the lethality rung, the weight hook, both canon tumble examples, overfly-the-ramp-pays-fall (the airborne model's distinguishing case), the bent path, lip/occupied stops, the water shoreline pin, killed-by-hit never shoves. Plus a play-level end-to-end void removal, and the elevation-delta Law #2 case reshaped to a drop.
- **Falsified both blind spots**: skipping the final lethality predict (fall damage can't change the rung) redded exactly the drop-changes-rung case; deleting the play-session `die()` door (the hand-mirrored twin) redded exactly the play void case. Both reverted, suites green again.

## Feel-check script (on a board with hole tiles + a terrace)

1. Shove an ally across a hole — it should fly over and land safely on the far side.
2. Shove an enemy so its travel ends on the hole — removed, no ghost, the trail alone shows where it went.
3. Shove one off the terrace — the "Fell N!" popup and the damage row should match the preview exactly.
4. Knockback-1 onto the descending ramp chain — free tumble to the bottom, no fall damage; a longer shove over the same ramp pays the drop where it comes down.


Phaazoid:
🤖 Claude says:

Review round landed as two commits.

**`5cc1915` — the verticality rework.**
- *CI red fixed*: `BoardMirror` had no answer for `Terrain.Kind.VOID`, so a hole tile rendered as fallback dirt ground in 3D. It's now the second declared skip with the opposite meaning to NONE — the column clears, so **the pit is the absence of the block** (the neighbours' side faces are the pit walls). Pinned by a paint-a-hole case.
- *The shove slides*: `MovementComponent.slide_along_path` is `move_along_path`'s shove twin — the sprite travels the resolver's own `knockback_path`, facing held the whole way (being moved is not moving; the 3D mirror gates its step-derived flip on it). In 3D the sprite holds its **launch height while airborne** and its height *eases* down at a fall speed, so a cliff shove reads as fly-then-fall and a tumble rolls the ramp. Two new Game-tab knobs under a **Playback** sub-tab: *Shove slide speed* (default 240 px/s, 2× walk) and *Shove fall speed* (default 4 world units/s) — both placeholders for your hands.
- *The arrows stop lying*: `ResolvedOutcome.knockback_landing_index` is the resolver's own flight/tumble split (Law #2 — animation and trail both read it, nothing re-derives). Flown cells draw **in the air at launch height**; the landing contributes a **vertical drop pointer** down to the destination (a removal points one level into the pit). Tumble arrows still lie on their ramps. 2D unchanged — the flat view renders no elevation anywhere.

**`3f80338` — the fixture-saving fix.** Two causes, both real:
- The Unit Editor's Save (and a dev spawn) never called `mark_modified` — that's why `(modified)` never lit.
- Deeper: the verticality fixture is an **authored save**, so every unit is a cast *reference* to its character file, and an authored Update re-referenced them — silently discarding inventory *and* team edits (faction rides the embedded copy, which a reference doesn't write). Per your call this session: **a dev edit breaks the reference** — the edited unit saves as an ordinary snapshot inside the authored save, untouched cast units stay references, and the divergence is sticky across loads. Cost to know: an edited unit stops following later edits to its character file.

**Verification**: presentation + law green (186s), dev + ui + flow green; the class-knob law learned that a static row names its own script home. Falsified the air-trail wire (muted the meta read → exactly the aimed case reds, reverted). The knob-law fix and both new pins run in CI with everything else.

**Feel-check script**: shove down the ramp chain (slide keeps facing) · off the cliff (arrows float, drop pointer, sprite falls at the landing) · across and into the hole (trail crosses in the air; a shove ending on it points into the pit — which now renders as an actual pit, worth a look on its own) · then edit a unit's inventory + team in the Unit Editor, press its Save, watch `(modified)` light, Update, F2 — the edits come back.

— Claude (Fable 5) · 2026-08-20

Issue #423, opened 2026-08-20
Squad ring forms, dissolves, and shatters: feedback on membership changes


Phaazoid:
> [!WARNING]
> **Two premises below were measured FALSE while planning this (2026-08-21) — see [the correction comment](https://github.com/Phaazoid/Godoiosis/issues/423#issuecomment-5371782266).**
>
> 1. **The ring was not a persistent membership marker.** It was produced only by selection/hover paths and destroyed by `clear_selection_icons`. #325 relocated it from head to ground and left the lifecycle alone.
> 2. **So the settle-point half could not work as written.** `execute_orders` clears the channel *before* the pass; both un-previewed ejections run *after* it. There was no ring on screen to shatter.
> 3. **The #358 dependency was wrong** — `OverlayIcon` is its own overlay node; #358's stack rides the unit's `MapSprite`. This was never gated on it.
>
> The persistence this needs shipped as **#435** (PR #436), which owes a dev verdict on whether standing rings stay. The event list below is still the useful part. The rest of the body is left as originally filed.

---

🤖 Claude says:

*(From the 2026-08-20 scratchpad sweep — filed at the dev's request, as recorded.)*

## The ask

Give squad membership changes a **moment** on the ring channel:

- the per-squad ring under a unit's feet **forms** when it joins a squad;
- it **dissolves** when it leaves;
- a **forceful** removal **shatters** it.

## The marker is settled, so this decorates something stable

The ring survived [#325](https://github.com/Phaazoid/Godoiosis/issues/325)'s verdict. The dev's call after playing both styles was a **mix, not a winner**: membership is a per-squad coloured **ring** under each member, leadership stays the **crown** over the head, and the legacy green squares lost outright and were deleted (`SQUAD_MARKER_RINGS` gone, no toggle, no second style).

So this is not decorating a marker that might be replaced — that question is closed. It also sits correctly under [#346](https://github.com/Phaazoid/Godoiosis/issues/346)'s channel rule: ground/underfoot = **what this interaction is about**, above the head = **what this unit is**. Membership changing is an event about the unit's relationship to a squad, and the ring is already the channel that carries it.

## The genuinely useful half: WHICH events

The ask names join and leave, but **the events worth catching are wider than the clicks the player makes**, and that is the part most likely to be missed:

[...]

Issue #358, opened 2026-08-18
Status reads on the sprite itself: wet drip, frost sheen, Crisis — one effect stack


Phaazoid:
*Filed by Claude from the dev's direction in conversation (2026-08-18), not written by the dev directly.*

## The ask

Dev, same conversation that filed #357: *"I think a next step for this would be to have visual effects on the sprite, done using fancy effects. Maybe like a dripping texture for wet, with a blue sheen. If a unit is frozen, they could have a frosty sheen to them, etc."*

And the doctrine that governs it: *"Having access to fancy effects doesn't necessitate using them. Using them in the correct places rather than everywhere makes them have more effect."* Unit status is one of the sanctioned places.

## Scope: one stack, three consumers

A sprite-effect stack on the unit's map sprite, expressing facts about the unit — the same channel rule as #346, one level closer than the head:

- **WET** — dripping texture, blue sheen (dev's words above).
- **CHILLED** — frosty sheen. (The dev said "frozen"; the unit state is `Elemental.State.CHILLED` — terrain `FROZEN` is a different fact with its own form already.)
- **CRISIS** — already canon, dev 2026-08-09 (visual-clarity.md): *"It should be very obvious a unit is in crisis mode"* — the map sprite itself should reflect Crisis. Same mechanism, third consumer, and the loudest of the three by design.

One stack means one answer to layering: a Wet, Chilled unit in Crisis must wear all three legibly. Ordering, blending, and any cap are design questions the build answers once, for every effect.

## Role in the visibility model

Per the dev (2026-08-18): when the player toggles health bars off (#350), sprite effects are what keeps states readable on the board — the icon row (#357) rides the bar; this stack rides the unit itself.

## Constraints

- **Route through the visible sprite.** #44's list already records that `UnitVisuals` effects animate `MapSprite` even while a downed unit shows `DownedSprite` — this stack must not repeat that bug.
- **Both views or a declared asymmetry.** Units render in the flat view too; an effect that exists only in 3D goes on #292's ledger deliberately, not by drift.
- **#325 connection.** Squad membership's experiment chose a ground form; a squad-colored rim riding this stack is that ticket's parked second contender. If it ever activates, the colour channel must not collide with the state sheens.
- Knobs per the tuning rule — these are exactly what the Look tab's presets exist for.

Blocked by #357: the dev sequenced the icon row first ("at the very least" → "a next step").

[...]

PR #436, opened 2026-08-21
Standing squad rings behind a player setting (#435)

### LINKED ISSUE #435: Standing squad rings: a toggle for always-on membership markers (#423 slice 1)
*Filed by Claude while planning [#423](https://github.com/Phaazoid/Godoiosis/issues/423), from the dev's call in conversation (2026-08-21).*

## Why this exists

#423 asks for the squad ring to **form** on join, **dissolve** on leave and **shatter** on a forceful removal. Planning it measured the premise false: **the ring does not persist.** `SQUADMEMBER` icons are produced only by selection and hover paths (`_on_squad_became_active`, `draw_joinable_squads`, `_repaint_squad_plan`, `HoverPresenter`) and destroyed by `clear_selection_icons`. `docs/design/visual-clarity.md` already recorded it about #325 — the build *"RELOCATED that one channel rather than restyling a fill, and the icon lifecycle never moved."*

That kills the half #423 called most valuable. `OrderExecutor.execute_orders` clears the icon channel **before** the pass; both un-previewed ejections (`_process_downed_pending`, `enforce_contact`) run after it, and the turn-start sweep fires when nothing is selected. **At both settle points there was no ring on screen to shatter.**

Dev's call: make the ring persistent first, behind a toggle, and feel always-on against today's behaviour before any animation is built.

## What this ships

- **`PlayerSettings.ALWAYS_SHOW_SQUAD_RINGS`**, default **off**. One enum member + one `DEFS` row; `SettingsScreen` is a pure projection of the table, so no UI work. Sibling of `ALWAYS_SHOW_HEALTH`.
- **`game.refresh_squad_rings()`** — the one sweep. It **rebuilds** the channel rather than adding to it, because `create_unit_icon` only ever adds and a squad shrinking back to one member has to lose its rings. Gated on having squadmates (`Unit.has_squad`'s question), deliberately **not** on `ring_hue`, which is dealt once at the first squadmate and never reset.
- Runs on membership changes (`squad_member_joined`, `squad_created`, `squad_deleted` — `squad_created` covers every leave, since `leave_squad` ends in `create_squad`), at turn start after `enforce_contact()`, at pass settle, and when the settings screen hands the board back.

## The bug persistence exposed

`OverlayIcon` stored the cell it was built on and `OverlayMirror` anchored on that copy. Invisible while markers were rebuilt every repaint — but the instant one outlives a move it sits on the cell its unit walked off, **in both views**. The marker now asks its unit (`OverlayIcon.current_cell`), one answer read by the 2D position and the mirror alike. [#308](https://github.com/Phaazoid/Godoiosis/issues/308)'s law on a second shelf.

## Declared scope decision

Standing rings **stand down for a whole resolution pass** and return on the settled board. A marker sits on its unit's *projected destination*, which mid-pass is a cell the unit has not reached, so a ring left up would jump ahead of the unit rather than travel with it. Riding the animated position is a separate build. If that reads wrong in the feel-test, it is the follow-up.

## Verdict this owes

Same shape as #325's own play-both-then-decide: the toggle exists so the dev can judge always-on against selection-scoped. **When he calls it, the loser is deleted** — either the setting goes away and standing rings become unconditional, or it stays as a preference. #423's animation waits on that answer.

Parity: both views move together, so nothing for [#292](https://github.com/Phaazoid/Godoiosis/issues/292).

Phaazoid:
Closes #435. Prerequisite for #423, which cannot be built as written.

## Why this and not #423

#423 asks the squad ring to **form** on join, **dissolve** on leave, **shatter** on a forceful removal. The thing it proposed to animate **does not persist**: `SQUADMEMBER` icons are produced only by selection and hover paths and destroyed by `clear_selection_icons`. `visual-clarity.md` already said it about #325 — *"the icon lifecycle never moved."*

That kills the half #423 called most valuable. `OrderExecutor.execute_orders` clears the channel at line 60; `_process_downed_pending` and `enforce_contact` run at 95 and 99, and the turn-start sweep fires when nothing is selected. **At both settle points there was no ring on screen to shatter.** Fifth instance of #228's law.

Also worth recording: **#423's declared dependency on #358 was wrong.** `OverlayIcon` is its own overlay `Node2D`; #358's stack rides the unit's `MapSprite`. Different node, different channel — #423 was never gated on it.

Dev's call (2026-08-21): ship the persistence first, behind a toggle, and feel it before any animation.

## What's here

- **`PlayerSettings.ALWAYS_SHOW_SQUAD_RINGS`**, default **off** — so the setting cannot regress the old board by existing. One enum member + one `DEFS` row; `SettingsScreen` is a pure projection, so no UI work.
- **`game.refresh_squad_rings()`**, the one sweep. It **rebuilds** the channel rather than adding to it, because `create_unit_icon` only ever adds and a squad shrinking back to one member has to lose its rings.
- Gated on having squadmates (`Unit.has_squad`'s question), **not** on `ring_hue` — a hue is dealt once at the first squadmate and never reset, so that gate would leave a lone leftover wearing a colour forever. There's a case pinning exactly that.
- `draw_squad_unit_icons` split out of `redraw_squad_unit_icons` so the sweep reuses the "what does a squad wear" answer instead of copying it.
- Triggers: `squad_member_joined`, `squad_created`, `squad_deleted`, turn start after `enforce_contact()`, pass settle, and the settings-screen return. **`squad_created` covers every leave** (`leave_squad` ends in `create_squad`), so the ejection sweeps needed no new signal.

## The bug persistence exposed

`OverlayIcon` stored the cell it was built on; `OverlayMirror` anchored on that copy. It only ever worked because markers were rebuilt constantly — a ring that outlives a move sat on the cell its unit walked off, **in both views**. The marker now asks its unit (`current_cell`), one answer read by the 2D position and the mirror alike. #308's law on a second shelf.

## Declared scope decision — please judge this in the feel-test

**Standing rings stand down for a whole resolution pass** and return on the settled board. A marker sits on its unit's *projected destination*, which mid-pass is a cell the unit has not reached, so a ring left up would jump ahead of the unit rather than travel with it. Riding the animated position is a separate build. **If that reads wrong while you're playing, it's the follow-up.**

## Knobs

The setting is boolean and lives on the existing settings screen (pause → Settings). No feel value is pinned; `SQUAD_RING_ALPHA` is untouched.

## Verification

`presentation` (306 cases) and `core ui squad flow` (588 cases) — all green, 0 orphans. Never the full tree; CI owns it.

Six new cases in `tests/presentation/test_standing_squad_rings.gd`, driven through the real seams (spawn, `join_squad`, `leave_squad`, a real pass) rather than by calling the sweep, because the claim is that membership changes reach the channel at all: off-by-default draws nothing; two members wear rings in **both** views; solo squads wear none; a squad shrunk back to one wears none *and* still has its hue (so the case still covers the trap it was written for); the ring follows a unit that moved; rings come back after a pass settles.

**Falsified:** freezing the cell at creation — the pre-fix behaviour — reds `test_a_standing_ring_follows_the_unit_that_moved` **alone**, with the four cases before it still green. That was the assertion most likely to be written blind.

**Not checked:** these assert on the marker store and the mirror's anchors, never on pixels. Whether standing rings read well on a busy board is exactly the question the toggle exists to let you answer.

## What this owes

Same shape as #325's play-both-then-decide: **when you call it, the loser gets deleted** — either the setting goes away and standing rings become unconditional, or it stays a preference. #423's animation waits on that answer.

### DIFF
--- a/Classes/actions/OrderExecutor.gd
+++ b/Classes/actions/OrderExecutor.gd
@@ -125,6 +125,12 @@ func _end_squad_turn(squad: Squad) -> void:
 	game.refresh_end_turn_button()
 	for member in squad.members:
 		game.overlay_manager.clear_planned_path(member)
+	# LAST, not beside the ejection sweeps above: the clear at the top of this method would wipe an
+	# earlier restore. Standing rings deliberately stand down for the WHOLE pass -- a marker sits on
+	# its unit's projected destination, which during a pass is the cell the unit has not reached
+	# yet, so a ring left up would jump ahead of the unit instead of travelling with it. Riding the
+	# animated position is a separate build; the settled board is where they come back.
+	game.refresh_squad_rings()
 
 # Why a concede happened, for the console: an AI cannot see a red flash, so this is the only place
 # the refusal is visible at all.
--- a/Classes/board/OverlayIcon.gd
+++ b/Classes/board/OverlayIcon.gd
@@ -1,14 +1,17 @@
 extends Node2D
 class_name OverlayIcon
 
-# One marker hung on a UNIT rather than a cell: a Sprite2D plus the cell it was anchored to.
+# One marker hung on a UNIT rather than a cell: a Sprite2D that follows the unit it was built for.
 # Built and pooled by OverlayManager (icons_by_unit), styled by its _style_icon (#325: a ring
-# underfoot or the legacy head square), and mirrored by OverlayMirror._icons into
-# BoardOverlays.Layer.GROUND_ICONS (rings, surface decals) or Layer.ICONS (squares, billboards).
+# underfoot, the crown over the head), and mirrored by OverlayMirror._icons into
+# BoardOverlays.Layer.GROUND_ICONS (rings, surface decals) or Layer.ICONS (crowns, billboards).
 
 @onready var sprite = $Sprite2D
 var icon_type
-var target_cell: Vector2i
+# The unit this marker hangs on, and the grid its cells are measured in. The CELL is deliberately
+# not stored -- see current_cell.
+var unit: Unit
+var grid: TileMapLayer
 
 # The above-the-head channel: what a unit IS, never what the current interaction is about
 # (#346 -- ground markup carries the interaction). CURSOR/TARGET/INVALID lived here and are
@@ -19,7 +22,27 @@ enum IconType {
 	SQUADMEMBER
 }
 
-func setup(texture: Texture2D, cell: Vector2i, type: IconType):
+func setup(texture: Texture2D, marker_unit: Unit, board_grid: TileMapLayer, type: IconType):
 	sprite.texture = texture
-	target_cell = cell
+	unit = marker_unit
+	grid = board_grid
 	icon_type = type
+
+# THE answer to where this marker sits -- read by the 2D position below and by OverlayMirror alike,
+# so the two views cannot drift into two answers. Derived, never copied: a stored cell goes stale
+# silently the moment a marker outlives one repaint (#308), which was invisible while markers lived
+# only as long as a selection and got rebuilt constantly. get_projected_destination already returns
+# the projected cell while a move is queued and movement.cell otherwise, so planning and settled
+# state both read correctly from the one expression.
+func current_cell() -> Vector2i:
+	return unit.get_projected_destination()
+
+func has_unit() -> bool:
+	return is_instance_valid(unit)
+
+# Following the unit is the icon's own job rather than something each redraw path must remember;
+# forgetting is what left a persistent ring behind on the cell its unit walked off.
+func _process(_delta: float) -> void:
+	if not has_unit() or grid == null:
+		return
+	position = grid.map_to_local(current_cell())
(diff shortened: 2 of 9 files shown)


Phaazoid:
**Round 2 — a bug I found by asking what the setting actually looks like in play.**

With rings standing, `OverlayManager.redraw_squad_unit_icons` clears **every** `CROWN`/`SQUADMEMBER` marker and then draws the one squad it was handed — and `HoverPresenter` calls it on every hover-move preview. So **hovering one squad stripped every other squad's standing rings.** Since hovering is what you do between clicks, that is the state the board would have sat in for most of the feel-test — it would have looked like the feature half-working rather than like a bug.

That path was correct for a decade of selection-scoped markers: the only markers that should exist belong to the thing being selected. Standing rings are the first tenant of that channel that has to survive *somebody else's* redraw.

Fix: `standing_squads_source`, a `Callable` injected by game on the `SquadManager.board_source` idiom, so the redraw can restore the standing set without `OverlayManager` learning what a player setting is. **Which squads stand is now one answer** — `game.standing_ring_squads` — read by the sweep and the redraw alike, returning empty while the setting is off so the selection paths stay untouched.

The sweep keeps its own setting guard, and that is **not** redundant with the empty list: while the setting is off the sweep must not touch the channel at all, where clearing it and drawing nothing would wipe whatever the selection had up.

`test_hovering_one_squad_keeps_another_squads_standing_rings` was written **first** and watched fail with exactly its own failure message, the six cases before it green — so it is a real regression pin rather than a description of the fix.

Re-verified: `presentation` 307, `ui squad flow` 557 — all green, 0 orphans.

The general form, now in `visual-clarity.md`: **adding a persistent tenant to a channel means auditing every writer that CLEARS the channel, not just the ones that draw the new thing.**

Issue #431, opened 2026-08-20
Drop rail: one quad in the trail's own plane, aligned with the arrows


Phaazoid:
🤖 Claude says:

**The nitpick (dev, 2026-08-20, reviewing PR #430):** the shove drop pointer is drawn TWICE — two vertical quads crossed on each other, centred at the entry edge of the landing cell. When the incoming trail arrows run horizontal, only the quad in THEIR plane should draw, not both. It also doesn't quite line up with the flat arrows it should connect.

**Where it lives:** `OverlayMirror._append_drop_rails` (built in PR #430, the round-3 polish on the #259 airborne shove). Honest answer to "is this channeling through the arrow drawing system we already have": **half of it.** The rail rides the ordinary `Layer.KNOCKBACK` marker pool and wears the trail's own straight-rail texture (`kb_rail_texture`, stamped in `OverlayManager.show_knockback_preview`) — but its GEOMETRY is hand-built: a vertical basis and entry-edge position computed locally, not derived from the `_anchor_px` + `_apply_marker` conventions the flat trail arrows go through. The misalignment is hiding in those differences — notably `_apply_marker`'s per-layer lift is applied along the marker's own basis.y, which for the vertical rail pushes it HORIZONTALLY off the edge while the flat arrows get it vertically, and the art scale comes off the texture size rather than the trail's own metrics.

**The fix shape:**

1. **One quad, not a cross** — in the vertical plane containing the travel direction, i.e. the plane the incoming arrows draw in. The cross existed because a single vertical quad vanishes edge-on under free orbit; that trade is now ruled the wrong way. If edge-on ever bites in play, the fallback is a camera-facing pick, never a second quad.
2. **Derive, don't hand-tune, the alignment** — the rail's x/z, lift and width should come from the same conventions the flat trail markers use, so its top meets the last air arrow and its bottom meets the arrowhead exactly. Law #4 pressure: two spellings of "where does a trail marker sit" will drift again; extract the shared piece rather than adjusting the copy.

**Verification shape:** extend `tests/presentation/test_overlay_mirror.gd`'s rail cases — exactly ONE non-flat marker per drop, its plane parallel to the travel direction, and its x/z equal to the trail anchor's at the shared edge. The ramp-suppression case (a fall onto a ramp draws no rail) is unaffected.

— Claude (Fable 5) · 2026-08-20

Issue #324, opened 2026-08-16
The 3D fire effect needs a real pass: one static quad, no animation


Phaazoid:
*Filed by Claude from the dev's playtest notes (2026-08-16), not written by the dev directly.*

## The ask

Dev, playtesting v0.39.0: *"the fire effect needs fixing too."*

This is the third time fire has come back in a feel-check — at #215's close it was *"the fire looks not great, a sprite issue more than anything else"* — so it is worth giving it its own ticket instead of routing it to the art pass again.

## What is actually there

A burning cell gets one `MeshInstance3D` holding a **static `QuadMesh`** textured with `res://Art/LookDev/torch_flame.png`, Y-billboarded through the cell centre, plus an `OmniLight3D`. That is the whole effect. Specifically it has:

- **no animation** — no frames, no scroll, no vertex motion, no shader; a flame that never moves reads as a decal of a flame
- **one quad for a whole tile**, at `flame_size` 0.5 x 0.7, so a burning cell is a single small sprite standing in the middle of a 1x1 square rather than a cell that is on fire
- **the look-dev torch texture**, authored for a point light source on a wall, doing duty as terrain fire

The tuning is not the problem — `flame_lift`, `flame_size`, `flame_ground_gap`, `flame_writes_depth`, `flame_light_energy`, `flame_light_range` are all knobs on the Look tab already, and the geometry questions around it have each been fixed on their own merits (#243, #236/#298 depth, the clamp that stops the knob authoring the coplanar bug back). What is missing is that the effect has never been *designed*, only positioned.

## Things to weigh before building

- **Motion is what sells it, and motion is exactly what #217 governs.** Every flicker/strobe has to sit behind the one photosensitivity toggle, with the toggle-off state being a *steady* flame rather than a missing one. Whatever animation this grows must ship through that switch, not beside it.
- **`flame_writes_depth` stays false by default.** #236 turned it on, bought constant z-fighting and was reverted on the dev's call; #298 is the residual (a Y-billboard flame and a Y-billboard unit through one cell centre are literally one plane). A redesign that stops being a single billboard may dissolve #298 for free — worth checking whether it does before treating that as separate work.
- **BLAZE and BURNING share one texture today** (#174's "one Fire texture covers both"). If the effect gains any expressive range, whether the two states start reading differently is a design call, not a free win.

## Verification

[...]

Issue #44, opened 2026-06-18
Visual Clarity Overhaul (umbrella)


Phaazoid:
Umbrella / tracking issue for board-legibility & visual-clarity work, so it all lives in one place.

**Core problem (@Phaazoid, 2026-06-18):** when 5+ units are moving around each other in close space, it's hard to tell who's moving where. Direction still TBD — this issue is the home for working that out.

## 👉 The list is [this comment](https://github.com/Phaazoid/Godoiosis/issues/44#issuecomment-5159428807)

It holds **only** the work that has no issue of its own — anything big enough to be promoted becomes a sub-issue instead, and is then tracked in the panel above rather than repeated as a line item. An item that ships is deleted from that comment, so the list is always what's left.

New ideas: add a comment. It gets folded into the list, and the loose comment goes away.

**Prior art:** the old `visual-clarity-overhaul` branch (squad/movement era) has earlier visual tweaks — mine for ideas; not a live effort.

---
**Files:** `OverlayManager`, `CursorController`, `HoverPresenter`, `UnitVisuals`, `game.gd` · **Source:** in-code TODO triage (`docs/todo-triage.md`, #42) + @Phaazoid direction

*🤖 Issue text authored by Claude (Opus 4.8) at @Phaazoid's direction, 2026-06-18; extended by Claude (Fable 5 / Sonnet 5 / Opus 5) through 2026-07-27; checklist consolidated into a single comment by Claude (Opus 5) at @Phaazoid's direction, 2026-08-02 — the superseded version is in this issue's edit history.*

Phaazoid:
🤖 Claude says:

# The list — everything still open on this umbrella

Rewritten 2026-08-02 at @Phaazoid's request. **This comment is the whole list.** Every item was re-verified against the current code before it went on here; anything that had shipped came off.

Two rules this list follows:

- **Nothing that has its own issue appears here.** The sub-issue panel above is the single record of those — no line item, no issue number, no duplicate checkbox.
- **Done means gone.** An item that ships gets deleted from this comment, not struck through or ticked.

---

## The core problem

- [ ] **Simultaneous-movement legibility.** When 5+ units are moving around each other in close space, it's hard to tell who's going where (@Phaazoid, 2026-06-18). Still the reason this umbrella exists, still needs a design pass before any code.

## Attention & feedback

[...]

PR #439, opened 2026-08-21
Shove trails get their own colour, and their ghost can be highlighted (#44)

Phaazoid:
Builds the shove-trail-colour item filed on [#44](https://github.com/Phaazoid/Godoiosis/issues/44) 2026-08-21, plus a live bug found on the same function while verifying it. They ship together because they are the same few lines of `show_knockback_preview`.

## 1. Shove trails stop looking like move arrows

`show_knockback_preview` drew its trail with `_draw_arrow_trail(path, Color.WHITE)`, and `_arrow_modulate` returns `Color.WHITE` for a valid planned move. **Byte-identical** — so a shove the resolver is *predicting* and a move the player *authored* read exactly the same on the board, which is the opposite of what they mean.

**Your ruling: one hue, "this is a shove."** Not a second state for a lethal one — a shove that kills already announces itself twice (the queue's KILL row, and #431's pointer falling the full void plummet), so the trail doesn't have to carry lethality too.

- `static var KNOCKBACK_MODULATE` on `OverlayManager`, beside `SQUAD_RING_ALPHA` — a game constant, so a `GameKnobs.CLASS_KNOBS` row, not a Look knob.
- **Starting value is pale ice-blue, and it is a starting value.** Picked to be legible *against* the palette rather than merely different from white: `_arrow_modulate` already means red for a refused order and green for a Case-1 trailing member. The slider is on **Game → Colours**; the number is yours.
- **The 3D is free** — verified rather than assumed: `BoardOverlays._apply_marker:324` reads each marker's own `modulate`, and the `LAYERS` spec colour is only consulted for FILL/BRACKET layers, never SPRITE.

**The drop pointer had to change too, and this is the Law #4 half.** `OverlayMirror._append_drop` hardcoded its own `Color.WHITE`. Left alone it would have stayed white while the trail moved — one shove drawn in two colours, which reads as a bug rather than as a knob. It now **copies the modulate of the trail sprite it hangs from**, the way every other flat marker already does. Not a second constant.

**`restyle_knockback_trail()` is not optional.** The 2D sprites bake the tint at draw time and `OverlayMirror` copies *them*, so without a sweep the knob would move nothing until the next shove — #324's born-dead-slider lesson, and the reason `restyle_squad_markers` exists next door. It re-tints **trail sprites only**, forked on parent node (`arrow_icon_overlay`), because `knockback_preview_sprites` also holds the landing ghosts, which wear `PROJECTED_MODULATE` and are not part of the trail. That is the same parent test `OverlayMirror._split_knockback` already forks on.

## 2. A shoved unit's ghost could not be highlighted (live bug, both views)

Found while reading `show_knockback_preview`, not filed anywhere.

It hides the shoved unit's real sprite (`unit.visuals.set_projected(true)`) and appends the ghost to `knockback_preview_sprites` — but `has_projected_unit()` checked `projected_unit_sprites` **only**. So hovering a queue row for an attack that knocks its target back sent `HoverPresenter._highlight_unit` down its `else` branch into `UnitVisuals.set_highlighted`, which wrote `modulate` to a `MapSprite` that `set_projected(true)` had hidden — and `UnitMirror` hides the 3D sprite off the same `projected` flag. **The highlight went nowhere, in the flat view and the diorama alike.**

**Counted from the code before choosing a shape:** `set_projected(true)` has three call sites — `game.gd:853` and `OverlayManager.gd:792` are both the *move* ghost (`projected_unit_sprites`, keyed by unit), `:441` is the knockback ghost (an unkeyed `Array[Node2D]`). `has_projected_unit`/`set_projected_unit_highlighted` have exactly one consumer, `HoverPresenter.gd:252-253`.

**Fix:** the knockback ghost is keyed by unit as it is created, and one private `_ghost_for(unit)` answers across both stores; both public methods go through it.

**The two stores stay separate, and that is declared, not accidental** — their lifetimes genuinely differ (a move ghost dies with `clear_projected_unit(unit)`, a shove's with `clear_knockback_preview()`), so the single answer belongs at the **door**, not in the storage. I deliberately did not add a third answer to "which sprite stands for this unit"; `UnitVisuals.gd:12-17`'s #232 ruling is why.

## Verification

`run_tests.ps1 presentation dev` → **591/591 green**. Full tree is CI's job.

Four new cases in `tests/presentation/test_overlay_mirror.gd`, all asserting **relationships, never the hue** — `KNOCKBACK_MODULATE` is a tuning value and pinning its components would red the moment you drag the slider:

- a shove trail's tint is **not equal to** the planned-move tint, *derived from `_arrow_modulate` itself* rather than typed as a literal, and no knockback marker reaches 3D wearing it
- the drop pointer's modulate **equals** the trail's (the case finds the pointer by its non-flat basis, and asserts non-vacuously that one exists)
- moving the static and calling the sweep re-tints a preview **already on the board**
- a shoved unit's ghost reports as projected and its **ghost** — not the hidden real sprite — is what the highlight moves

**Falsified** (the missing-wire tier): reverting `_ghost_for` to the one-dict form reds `test_a_shoved_units_ghost_can_be_highlighted` on both its assertions — *"the board is standing a ghost in for this unit but will not admit it"* and *"highlighting a shoved unit moved nothing the player can see"* — and **nothing else in the suite**, so the mutant is caught precisely rather than by truncating an earlier case. Restored, re-run green.

**Not falsified, deliberately:** the drop-pointer copy. That assertion sits one hop from the change, which the triage says to skip.

Both suites' **static snapshots were extended** (`test_game_knobs.gd`, `test_overlay_mirror.gd`) — `KNOCKBACK_MODULATE` is process-global, and the third case tunes it, so without the restore it would leak into every later suite in the run.

**Embedded-content sweep:** nothing to do. This adds no field to an embeddable content type — it is a `static var` and a marker tint, so no `.tres` fixture or scenario snapshot can predate it.

## What the suite cannot see — this is the half that needs you

Whether the hue reads as *"this is being done to him"* at play distance, and whether it stays distinguishable from the refused-red and the trailing-green at a glance. That is exactly why it ships with the knob rather than as a number I defended.

## Toolchain note

**`godot --headless --import` on a fresh worktree wipes Dialogic's registries in `project.godot`** (`dch_directory`, `dtl_directory`). Hit on all three worktrees this session; reverted here so this diff is only the ticket. Worth a real fix, since a new worktree needs that import and committing the result would silently break dialog.


Phaazoid:
🤖 **Claude says:**

**Round 2 — the knob could not tint, and the cause was the art (found in play, 2026-08-21).**

Dev set the shove colour to yellow and got **green**. Not a bug in the knob: `modulate` **multiplies**, and the trail art is **cyan**. Measured across the 14 segments — a 5-colour palette, every entry low-red / high-green-blue:

| RGB | count | luminance |
|---|---|---|
| 112,248,248 | 403 | 207 |
| 136,248,248 | 357 | 215 |
| 40,232,248 | 229 | 176 |
| 80,208,200 | 211 | 169 |
| 40,208,248 | 157 | 162 |

Yellow `(1,1,0)` zeroes the blue and leaves `(136,248,0)`. Green, exactly as reported.

**The same fact means today's "white" planned-move arrow has always rendered cyan** — and round 1's ice-blue shove default was landing on cyan art, so it barely separated from a move arrow either. The knob was multiplying against a hue the whole time.

## The fix

**The 14 trail segments are now greyscale.** Luminance (`0.299/0.587/0.114`), then **one global scale** so the brightest palette entry hits **255** — 215→255, 207→246, 176→210, 169→201, 162→193. Five grey levels, original ordering and spacing preserved, so the pixel-art shading survives; verified 1:1 against the old palette's pixel counts, nothing collapsed.

- **Global scale, not per-file** — a per-file normalize would let a corner tile end up brighter than a straight one, since the files carry different subsets of the palette.
- **Peak 255** rather than the old 248 peak: the point of the ticket is *what you dial is what you get*, and at 255 a knob colour reproduces exactly at the brightest pixels.

**Two files in that folder are deliberately excluded, and this is the Law #4 half:**

- **`ERROR.png`** has a second consumer — `GridUtils.ERROR_ICON`, the unknown-terrain fallback in `get_terrain_icon_at_cell`, which is drawn **untinted**. Greyscaling it would turn a red "this should not happen" diagnostic into a white square. It is also not "the base arrow" — it is what the atlas returns when it has no answer.
- **`nomove.png`** is `MoveAction.HOLD_ICON`, a queue-row icon; `draw_path_arrows` never puts it on a trail.

## All three arrow tints are knobs now (dev request)

On cyan art, `_arrow_modulate`'s three literals only *shaded* a colour the PNG carried. On greyscale art they **are** the colour — so leaving them hardcoded would have made the move arrow's appearance newly code-determined with no door to it. They join `KNOCKBACK_MODULATE` on **Game → Colours**:

- **`MOVE_ARROW_MODULATE` — pre-tuned to `Color(0.533, 0.973, 0.973)`**, i.e. `136/255, 248/255, 248/255`: the art's own brightest pixel. A planned move therefore looks **exactly as it did before**, rather than turning white by accident. (Dev asked for this explicitly.) The darker steps are now shaded versions of one hue instead of the original ramp's slight hue shift as it darkened.
- **`INVALID_ARROW_MODULATE`** / **`TRAILING_ARROW_MODULATE`** keep their authored numbers — so **expect them to read brighter and more saturated**, because the cyan used to multiply them down (`136,248,248 × 1,.25,.25` was a muted `136,62,62`; on white it is a clean `255,64,64`). That is the change working, not a regression, and both now have sliders if it is too much.

**No new sweep for these three:** `redraw_planned_paths()` already tears down and rebuilds every arrow through `_arrow_modulate`, so it *is* the re-apply. Only the shove needed a bespoke `restyle_knockback_trail`, because a knockback preview has no rebuild-from-plan path.

## Verification

`run_tests.ps1 law presentation dev` → **693/693 green**.

**New law: `tests/law/test_arrow_art_is_monochrome.gd`.** Re-colour these PNGs and every arrow knob silently lies again, the same way, invisibly — this exact bug reached play without a single test noticing, so it is worth a guard. It asserts the **property** (`r == g == b`, plus a peak of 1.0 so a knob colour can actually be reached) and **never a specific grey** — the shading ramp is art, and pinning its values would be the content razor's own mistake. Textures are reached through `OverlayManager`'s consts rather than re-typed `res://` paths.

**Falsified** — and this one needed it, since a law that reads its subject through `get_image()` could pass vacuously. Restoring one PNG to cyan reds it with the offending pixel named: *"horizontal is not greyscale -- worst pixel (0,5) rgb 0.439/0.973/0.973"*. Restored, re-run green. The first draft asserted per-pixel and produced ~200 identical failures for one bad file; it now reports **one worst offender per texture**, which is what made the falsification output readable.

`test_game_knobs.gd`'s static snapshot grew by three. `test_overlay_mirror.gd`'s did **not** need to — no case there tunes the arrow tints, and adding unused restores would be noise. (Small deviation from the approved plan, which said both; stating it rather than quietly doing less.)

## What the suite still cannot see

Whether the pre-tuned cyan actually matches what you were looking at, and whether the brighter refused-red and trailing-green read right against the new shove colour. That is the launch — and the shove hue is still yours to pick; it now behaves.

— Claude (Fable 5) · 2026-08-21

Issue #418, opened 2026-08-20
Health bars: a third player-settings option for damaged units only


Phaazoid:
🤖 Claude says:

*(From the 2026-08-20 scratchpad sweep — filed at the dev's request, with his scoping call recorded below.)*

## The ask

Show health bars for **damaged units only** — every unit below full HP wears its bar, full-health units do not. The hovered unit keeps its bar and its digits either way.

## Scope — the dev's call, 2026-08-20

**This is a THIRD OPTION in player settings, not a new default.** It joins the existing choice rather than replacing either end of it. So after this ticket the player picks between three:

1. **Hovered only** — today's default, #229's behaviour.
2. **Damaged only** — this ticket. *(new)*
3. **Always on** — #350's behaviour.

Nothing about the shipped default changes; a player who never opens settings sees exactly what they see now.

## The model half is one disjunct

The visibility gate is deliberately **ONE named expression** — `hovered or foretold or always_on` (`docs/design/visual-clarity.md` → *Every bar, if the player says so*). "Damaged only" is another disjunct in that same expression, not a second visibility rule. [#357](https://github.com/Phaazoid/Godoiosis/issues/357)'s state-icon row hangs off the same gate, so keeping it single is load-bearing — a second rule in `UnitMirror` is the bug that ticket already warned about.

**`foretold` must survive whatever shape this takes.** A bar that is up because a queued plan is about to change that unit stays up under every setting. Law #2 says the queue never lies, and [#354](https://github.com/Phaazoid/Godoiosis/issues/354) ruled that a prediction survives to the end of its pass; a preference that could hide what the queue is promising would undo both. This is the one clause the ticket may not touch.

## The real work: `DEFS` has no non-boolean row

`PlayerSettings.DEFS` describes **booleans** (`title` / `desc` / `default`), and `SettingsScreen` is a **pure projection** of it — that projection is the whole point of paying for a table with one row in it ([#350](https://github.com/Phaazoid/Godoiosis/issues/350)). A three-way choice is therefore the **first non-checkbox row the settings table has ever had**, and that is most of this ticket:

- **(a)** Teach `DEFS` a choice kind (an option list plus a default index) and teach the projection to render it. More work now; every future multi-state setting is then a table row, which is the property #350 bought.
- **(b)** Express it as two independent bools. Cheaper, and it makes an illegal state representable ("damaged only" + "always on"), which the single-gate rule above then has to resolve somewhere.

[...]

Issue #314, opened 2026-08-16
Health as discrete 3D tick-blocks that explosively fall away when lost


Phaazoid:
🤖 Claude says:

**Dev idea, 2026-08-15, during [#229](https://github.com/Phaazoid/Godoiosis/issues/229)'s feel-check:** *"In fire emblem, health bars are denoted by a green bar with individual ticks in it per unit of health, that visually drain. That would be cool to do with 3D blocks, that visually fall away, explosively."*

## The idea

Replace #229's continuous fill with a **row of discrete blocks, one per point of HP** — the Fire Emblem read, where you count what is left rather than estimate it — and make losing health a **physical event**: the lost blocks break off the bar and fall away, explosively, rather than the bar simply getting shorter.

## Why it fits this game specifically

**Counting beats estimating here, because there is nothing to estimate.** Law #1 says no randomness: damage is exact and previewable. A continuous bar communicates *roughly how hurt*, which is the right readout for a game where the next hit's damage is a distribution. This game's player wants *exactly how many points stand between this unit and the next rung*, and a tick per point answers that at a glance. It is the readout the ruleset actually earns.

**And the blocks are already the visual language.** The whole HD-2D stack is voxel columns, corner brackets, generated facets — [#263](https://github.com/Phaazoid/Godoiosis/issues/263)/[#264](https://github.com/Phaazoid/Godoiosis/issues/264)/[#274](https://github.com/Phaazoid/Godoiosis/issues/274) spent real effort making the board read as solid blocks. A health bar made of little blocks that shatter belongs to that world in a way a flat quad does not.

## What it would cost, honestly

#229's bar is deliberately cheap: three unshaded quads with no per-frame allocation, shown for one unit at a time. This is a different animal.

- **Geometry scales with max HP.** A unit with 23 HP is 23 blocks. That is fine for one hovered unit and a real question if this ever becomes always-on, or if it lands on top of [#313](https://github.com/Phaazoid/Godoiosis/issues/313)'s predicted bars, where several units show at once.
- **"Falls away explosively" is a simulation, not a state.** #229's bar is a pure function of current HP — told a number, draws it, idempotent, no history. A block that tumbles has velocity and a lifetime, so something must own it after the HP that spawned it is already gone. That is the first real design question: probably a short-lived detached pool that outlives the bar it fell off, never per-block physics bodies.
[...]

Issue #313, opened 2026-08-15
Predicted health above units: a ghost bar for what the queued plan will do


Phaazoid:
🤖 Claude says:

**Dev idea, 2026-08-15, during #229's feel-check:** *"while combat is queued, units with actions queued should keep their health bars over their heads as ghosts, with what their depicted health will be. Then, if it's going to get too low, or they're going to get downed, we can make it flash or something."*

## What this is

[#229](https://github.com/Phaazoid/Godoiosis/issues/229) put a health readout in the volume above a unit, shown **only while the pointer is over it**. This is the other half: a readout that stays up **because a plan is about to happen to that unit**, showing the HP the plan predicts rather than the HP it has.

It is the single most valuable thing that volume can carry, because it turns Law #2 into something you can see. The plan already knows the answer exactly — this is about drawing it.

## Why it is nearly free at the model layer

**The number already exists and is already authoritative.** `PlanResolver` threads a hypothetical through the whole pass, and `ResolvedOutcome` carries `target_hp_after` — the same value `BugReporter` prints raw and the queue panel's ladder reads. Law #1 (no randomness) is what makes it honest: the prediction is not a forecast, it is what will happen. Law #2 says the queue never lies, so a bar showing predicted HP is that law rendered.

So this ticket is **presentation only**. It must NOT compute damage; if it ever needs to, that is the bug.

Where to read it: `SquadManager.resolved_plan_for(squad)` holds the last resolve (`_last_resolved_plan`), which is what the candidate gates already consult. Whether that is the right source for a per-frame reader, or whether the readout should ride the same repaint the queue panel does, is the first thing to settle.

## The questions it raises

1. **Which units show one?** Everyone the plan touches (including enemies your attack will hit — that is most of the value), or only units with queued orders of their own? "Anyone whose predicted HP differs from their current HP" is the obvious rule and probably the right one.
2. **What does "ghost" mean visually?** #229's bar is opaque, black-outlined, bright green over red — deliberately not atmospheric. A ghost bar has to read as *prediction* against that, and the honest options are a second bar, a marker on the same bar (a notch at the predicted level, with the doomed segment in a third colour), or a translucent twin. **The notch is the strongest candidate**: one bar, no crowding, and it shows current AND predicted in one shape.
[...]

Issue #229, opened 2026-08-14
Unit billboard HUD: health, status icons and more in the space above units


Phaazoid:
🤖 Claude says:

**Dev feel-check, 2026-08-14 (stage 4c on `main`, v0.15.0):** *"I like the icons hovering above the units rather than being on them, nice touch. I think we can also fit status icons there, health bars, and more. We should add more to that design space."*

## What exists

4c mirrors the 2D selection icons (crown / squad member / target / cursor / invalid) as **billboards floating above the unit** rather than stamped on its cell — `BoardOverlays.Kind.BILLBOARD`, `Layer.ICONS`, height on the `billboard_lift` knob. That was a mechanical consequence of moving to 3D; the dev's read is that it is *better* than the 2D placement, and worth building on deliberately.

## The design space

The volume above a unit is now real estate the flat board never had. Candidates:

- **Health** — a bar, or pips, or only-when-damaged.
- **Status/element states** — the `UnitStatesBar` / `StateIcons` content the 2D game already renders in the panel (BURNING, FROZEN, CHILLED, COVER…).
- **Will** — the resource whose depletion is the death spiral; currently readable only in the inspect panel.
- **Turn/order state** — has acted, has a queued order, is the active squad's leader.
- **Damage numbers / hit feedback** — see #188, which is exactly this space for one specific event.

## The questions this space raises

1. **How much is too much?** Five stacked billboards over every unit on a busy board is noise, and the diorama read is the thing #176 is buying. Likely wants a policy: always-on (health?), on-hover, on-selection, or on-change-only.
2. **Occlusion and crowding** — billboards over adjacent units overlap at low camera pitch. 4c already staggers co-celled icons by a small height offset; a real HUD needs a stronger answer.
3. **One source, two surfaces?** The 2D game renders states in panels, not over units. If this becomes the 3D presentation of the same facts, it should read the SAME state (Law #4) — the 4c `OverlayMirror` precedent is to poll the 2D's retained state rather than invent a parallel one.
4. **Does the flat 2D game get it too?** Parallel-stacks doctrine says no obligation either way, but worth stating.

## Bins

Sits between #44 (Visual Clarity Overhaul umbrella) and #176 (HD-2D). #188 (no feedback when a unit takes fire damage) is a concrete instance of the same space and should land consistently with whatever policy this settles.

— Claude (Opus 5) · 2026-08-14

Phaazoid:
🤖 Claude says:

[...]

PR #466, opened 2026-08-21
A downed body's readout says so (#322)

### LINKED ISSUE #322: A downed unit's health bar reads the same as a living unit at 1 HP
*Filed by Claude from the dev's playtest notes (2026-08-16), not written by the dev directly.*

## The ask

Dev, playtesting v0.39.0: *"A downed person's healthbar should have some feedback that they are downed, other than just having 1hp, to distinguish it from a not downed person with 1hp."*

#229 gave every mirrored unit a `UnitHealthBar` in the volume above it. A downed unit reads `1/20` on that bar. A living unit clinging on at 1 HP reads `1/20` on the same bar. Those are two completely different board states — one is a body that has to be rescued, the other is a unit that can still act, still counter, and still be finished off — and the readout does not separate them.

## Why the number alone cannot carry it

`DOWNED` is a lifecycle state, not a hit-point value; the 1 HP is an artifact of the ladder, not the fact. The information already exists and is already read elsewhere — `UnitSprite3D.is_downed` swaps the downed art, and `OverlayMirror` marks a downed body in the headless view — so this is a matter of the HUD band asking a question the rest of the stack already answers, not of deriving anything new.

Which also means the wrong fix is to special-case "1 HP means downed" anywhere. It is false the moment a living unit sits at 1, which is precisely the reported confusion.

## What to decide

The form. Some candidates, none picked:

- **The bar itself changes** — greyed out, hatched, or drained to empty with a distinct frame. Reads instantly, but competes with the "at 1 HP" reading it is trying to distinguish from.
- **A state icon beside the bar.** #229's readout already lives in a band designed to carry status icons (`UnitStatesBar` is the 2D precedent), so a downed glyph is the least novel option and generalizes to the rest of the lifecycle.
- **The downed clock as the readout.** A downed unit has `downed_turns_remaining` ticking; showing the turns left before it is lost is strictly more information than "downed", and it is the number that actually drives the player's decision about whether to spend a rescue.

The third is the most interesting and the most opinionated — worth a call rather than a default.

Whatever ships, the 2D has the same gap and should get the same answer, or the asymmetry gets declared (#292).

Related: #229 (the health billboard this extends), #313 (predicted-health ghost bar), #314 (health as tick-blocks), #33 (the downed/rescue lifecycle), #292.

Phaazoid:
Closes #322.

A downed body and a living unit clinging on both rendered `1/20`, because `_go_downed` parks a downed unit at **exactly 1 HP** — the number is an artifact of the ladder, not a fact about the unit. The row above the bar now carries the **downed glyph and the rescue clock beside it**.

## What it does

`UnitMirror._sync_bar` appends `StateIcons.DOWNED` to the row it already builds and hands the turns down; `UnitHealthBar` grows a second `Label3D` placed off the row's measured right edge. Element states first, lifecycle after, icon then count — the order and the shape `hover_info_grid_container` has always drawn downed in.

**Form wasn't really an open fork.** The ticket offered three candidates (grey the bar / a state icon / show the clock), but the hover card already draws icon-**and**-clock and `info_panel` draws `DOWN 3` tipped *"Dies in N turn(s) without rescue"*. A board band inventing a fourth spelling would have been a second answer to a question the game had answered twice.

**No new visibility rule.** The row is a child of the bar and rides `hovered or foretold or always_on`. That's legitimate here specifically, and the reason is in the file so a future ticket doesn't re-litigate it: #325's law says a marker answering *what is this unit* must not ride another marker's gate — and the downed fact **is** carried unconditionally, by the swapped downed sprite. The glyph stops the bar contradicting the art; it isn't the only thing marking the body. A status with no art of its own couldn't make that argument.

**One derived value** drives glyph and count (`unit.downed_turns_remaining if unit.is_downed() else -1`, `> 0` like the hover card, since the clock emits 0 in the instant before the body is lost) — two reads could disagree for a frame.

## Seam cleanup carried along

`Down.png` was preloaded independently by `hover_info_grid_container` and `AttackAction`; a third surface is where that stops being tolerable. `StateIcons` gains a `DOWNED` const — deliberately **not** an `ICONS` entry, since that dictionary is keyed by `Elemental.State` and a lifecycle is not one — and the hover card reads it from there. **`AttackAction`'s DOWN/KILL/MAIM triple stays put**: a predicted lethality *rung* is a different question, and it needs kill and maim art no status row has any use for.

## Knobs

Two new `GameKnobs` rows, Unit HUD group (game constants, not mood — nobody may disagree about them per board):

- **Downed clock size** (`downed_count_height_cells`, default 0.09) — the digits' height in cells, the HP number's unit, deliberately smaller than it.
- **Downed clock inset** (`downed_count_gap_texels`, default 1.0) — gap between the last icon and the digits.

Colour and outline are the HP digits', **not** knobs: two texts on one display should agree by construction rather than by a slider somebody keeps in step.

## Scope left alone, deliberately

- **The bar still reads `1/20` under the glyph.** Turning the track itself into the rescue clock is the ticket's third option and a bigger call — #314's tick-blocks would force it to be re-decided anyway. Say the word if the glyph alone doesn't settle it in play.
- **No 2D half, and it isn't a new gap.** The ticket's *"the 2D has the same gap"* is measured false: the flat view has no over-unit health readout of any kind, so there is no `1/20` there to disambiguate — and it already says downed twice (swapped sprite, hover card icon + count). #229's absent flat-view readout stays the declared #292 asymmetry it always was; #418 says the same thing.
- **No predicted-down glyph.** The alarm pulse already owns *about to go down* (#313).

## Verification

`tests/presentation/test_unit_health_bar.gd` gains four cases, driving the real chain (pointer -> HoverPresenter -> UnitMirror -> bar). Areas run green, not the tree: **presentation 333/333, ui 213/213, dev 291/291**, 0 orphans.

The headline case asserts the actual complaint — a body and a living unit at 1 HP side by side, with the premise (*"they really do read the same HP"*) asserted rather than assumed, so it can't quietly stop testing the bug.

**Falsified, three mutants, one suite file:**

| mutant | result |
|---|---|
| never derive downed (`downed_turns` always -1) | `..._at_one_hp_read_differently` **RED** |
| clock out of the `_drawn` redraw signature | same case RED; with it disabled, `..._follows_the_countdown_down` **RED** |
| count anchored to the bar instead of the row | only `..._rides_the_row_rather_than_the_bar` **RED** |

Worth recording: **the suite stops at its first failure**, so mutant 1 and 2 truncated the cases I was aiming at and I had to disable the earlier case to get a verdict on the later ones. A mutant run that doesn't name which case reddened is not evidence.

**Embedded-content sweep:** nothing to sweep — no field was added to an embeddable content type, and no `.tres` schema changed. The glyph is a code const.

**Not checked:** how it reads in play. Size and inset are guesses behind knobs; the whole point of the ticket is whether a body now reads as a body at a glance.

You are on `feat/322-downed-readout` — launch it.


Phaazoid:
## Round 2 — the clock is the HP digits' size

> "The number is too small - any number needs to be at least as big as the numbers in the healthbar to be readable. Smaller than that is just impossible."

Read as a **legibility floor rather than a preference**, which changes the fix: instead of re-defaulting the knob, `downed_count_height_cells` is **deleted** and the count reads `number_height_cells` directly — joining the colour and outline it already inherited, for the reason those were never knobs either (two texts on one display should agree by construction). Its `GameKnobs` row went with it, replaced by a comment saying why there isn't one.

A dial whose entire lower range you've called impossible to read is worse than no dial. If you ever want the count *bigger* than the HP digits, that's a knob worth adding deliberately — say so and it's one table line. **Downed clock inset** (the gap between glyph and digits) is untouched.

Pinned as a **relationship, not a value** — `downed_count_glyph_height() >= number_glyph_height()` — so retuning "Number size" moves both sides and can never red the suite. That's the tuning razor applied to a rule instead of a number. Falsified: shrinking the count to 0.7x reddens that case and **only** that case.

Also merged `main` (the conflict was `visual-clarity.md`'s staleness marker against #464's — checked that the auto-merged half of that file dropped nothing #464 added, since a clean merge can silently eat an adjacent line).

**Verification:** presentation 334/334, dev 291/291, 0 orphans. Still not checked: how it reads in play — that's the whole question.

You are on `feat/322-downed-readout` — relaunch it.

Issue #105, opened 2026-07-28
Projection: three answers to "where will this unit be when the plan resolves?"


Phaazoid:
🤖 Claude says:

Found by the [#104](https://github.com/Phaazoid/Godoiosis/issues/104) duplicate-seam sweep. Every
divergence below is **confirmed by probe tests**, not inferred from reading; the probes were
temporary and are quoted at the bottom so they can be pasted back as the regression suite.

## The question

*Where will this unit be standing when the current plan resolves?*

## The answers — three of them, plus two draw paths

| # | source | rule | knows about a queued move | honours `is_valid` | knows about knockback |
|---|---|---|---|---|---|
| 1 | `Unit.get_projected_destination()` (`units/Unit.gd:355`) | forward, reads `squad.get_actions()` | yes | **yes** | **yes** |
| 2 | `SquadPlanValidator.projected_cell_for()` (`squads/SquadPlanValidator.gd:39`) | forward, reads a **passed-in** action list | yes | **no** | **no** |
| 3 | `SquadManager.get_projected_unit_from_cell()` (`squads/SquadManager.gd:133`) | **reverse** index, active squad only | yes | yes | **no** |

Plus the board's visuals, which never ask any of the three: `OverlayManager.redraw_projected_units()`
draws the ghost from `planned_move_by_unit[unit].destination`, and `show_knockback_preview()` draws
its own separate ghost from `resolved.knockback_to` with its own hidden-sprite list.

None of the three derives from another. #2 and #3 each re-implement a subset of #1's rule, and each
drops a different part of it.

## Confirmed divergences

**1. An invalid move — the two forward answers disagree outright.** Queue a leader and a squadmate
onto the same cell; `_check_destination_conflicts` invalidates both. Then
`leader.get_projected_destination()` returns the leader's **current** cell (invalid move skipped)
while `projected_cell_for(leader, actions)` returns the **destination** (validity ignored).

This is not academic. `game.draw_squad_leader_range()` centres the drawn cohesion circle on answer
#1; `SquadPlanValidator._check_leader_range()` validates every member against a circle centred on
answer #2. **With an invalid leader move queued, the circle on screen is not the circle members are
being judged against** — the Law #2 shape, in the overlay layer.

**2. A hold-position move silently outranks the knockback projection.** `get_projected_destination`
returns on the first valid MOVE it finds, and `setup_hold_move_actions` gives *every* member of an
activating squad a hold move whose destination is its current cell. So for any unit in the active
[...]

Issue #88, opened 2026-07-24
Group ability actions under a menu category (mirror Weapon Action)


Phaazoid:
Follow-up to the **Weapon Action** menu refactor (2026-07-24): secondary weapon usages — a weapon's extra attacks, plus Rev / Spring Load / Burrow — were collapsed under a single "Weapon Action" menu category instead of each taking its own top-level slot, and Attack was reduced to always firing the weapon's default (main) attack with no submenu.

The same grouping should be applied to **ability-driven actions**, so the top-level menu stays short as more abilities land:

- **Intimidate** (`IntimidateAction`, #61) currently has its own top-level menu entry — move it under an "Ability Action" category (working name, TBD).
- **Rune / transmutation carvings** — the weapons-only Weapon Action refactor deliberately left runes alone (a rune's multiple carvings still open a pick-menu on Attack via `_begin_attack`'s rune branch). Fold those into this Ability Action category so Attack becomes submenu-free for runes too — they're alchemy, not weapon hardware.
- Future abilities (Fortitude / Guard / rally-enhancers, etc., per `docs/design/jobs.md`) get a home for free.

**Shape:** a second submenu category that appears whenever the unit has ≥1 ability-driven option, opening a pick submenu exactly like Weapon Action. Purely a menu-layer grouping — the underlying queued `ActionType`s (INTIMIDATE, etc.) don't change, so the resolver / queue panel / AI stay untouched, mirroring how the Weapon Action refactor kept `ATTACK`/`REV`/`SPRING_LOAD`/`BURROW` as the real order types.

Out of scope of the Weapon Action refactor; filed so it isn't lost.

— Claude (Opus 4.8) · 2026-07-24

Phaazoid:
🤖 Claude says: **BUILT and closing.** Landed 2026-07-29 on top of #90, suite **749/749**, feel-tested in-game.

## Three categories, not two — this issue's plan was wrong on one point

The body proposed folding rune carvings **into** the Ability Action category. @Phaazoid's call reversed that: *"Transmutations should be their own grouping, separate from active ability use. Transmutations stem from Runes which are equippables, abilities are a separate beast entirely."*

So the top level now has:

| Category | Source | Attack fires |
|---|---|---|
| **Weapon Action** | the weapon's non-main attacks + Reload/Rev/Burrow | its authored `main_attack` |
| **Transmutation** | a rune's carvings | the default (first channelable) carving |
[...]

Issue #422, opened 2026-08-20
Player-chosen overlay/reticule colours (needs a colour row in PlayerSettings.DEFS)


Phaazoid:
🤖 Claude says:

*(From the 2026-08-20 scratchpad sweep — filed at the dev's request, and **broadened by him at filing** to the general question underneath; see *The real ask* below.)*

## The surface ask

Let the player choose their own aim / attack reticule colours. Originally the dev's half-joke at [#212](https://github.com/Phaazoid/Godoiosis/issues/212) — *"a color slider to set whatever color they want for their aim/attack reticules, ha"* — kept because it is closer than it sounds:

- the reach fill is already a tunable `OverlayManager.ATTACK_MODULATE` **`static var`**, not a `const`;
- the 3D mirrors that modulate rather than holding a colour of its own, so tuning it moves both stacks by construction;
- since [#350](https://github.com/Phaazoid/Godoiosis/issues/350) the game **has** a settings surface, which did not exist when the joke was made.

Real colour-blindness support would want the whole `BoardOverlays.LAYERS` / `OverlayManager` modulate set rather than the reticule alone — making this a sibling of [#217](https://github.com/Phaazoid/Godoiosis/issues/217)'s photosensitivity toggle, the other accessibility knob whose home is that same page.

## The real ask (dev, 2026-08-20)

> *We want separate ways to keep track of things I'm changing as a dev for default feels, and things that players can change for their own settings page.*

**These are two different questions about the same values, and the project currently has stores for each — but no relationship between them.**

| | who decides | store | surface |
|---|---|---|---|
| **Dev default feel** | the dev, once, for everyone | the declaration that authors it — an `@export` default, a `static var`, a `LookPreset` | Moods / Game / Objects tabs, `KnobSource` writing back to source |
| **Player preference** | each player, for themselves | `user://settings.cfg` via `PlayerSettings` | `SettingsScreen`, a pure projection of `PlayerSettings.DEFS` |

Today a value lives in **one world or the other**, never both. Every overlay colour is dev-only (`GameKnobs.CLASS_KNOBS`, written back into `BoardOverlays.LAYERS` or an `OverlayManager` static); every player setting is a fresh boolean that nothing in the dev tools tunes. So the moment a *player* may override a colour the *dev* also tunes, that value needs to exist on both sides at once — and nothing in the codebase says what happens then.

[...]

PR #468, opened 2026-08-22
The unit action menu becomes a radial (#467)

### LINKED ISSUE #467: Rebuild the unit action menu as a radial around the unit's own sprite
🤖 Claude says:

*(Filed at the dev's request, 2026-08-21. His words: he doesn't like the floating dropdown, "it's getting bloated for a dropdown", make it a radial menu, "maybe with the unit's map sprite in the center, to show what we're actually acting on".)*

> **Every fork was resolved by the dev the same day, before any build.** The design below is settled: a compacting ring of **categories**, expanding **concentrically**, **persistent** until a terminal pick, driven by **angular hover-preview** in which the deepest open ring owns every angle and radius means nothing. The body has been rewritten around those answers twice — the original framing said a radial could not fix the bloat, and that was wrong. One placement is still flagged as open at the end; it does not change the shape.

## The ask

Replace the floating vertical dropdown that opens on a unit click with a **radial menu**, with the acting unit's map sprite at the centre so it is never ambiguous who the menu belongs to.

`MainActionMenu.ACTION_DATA` declares **19 verbs** and the top level can offer fifteen at once, which is the "bloated" complaint. The ring fixes it by nesting: circumference grows with radius, so each ring further out holds more than the one inside it, and no single ring has to hold nineteen.

## What is there now (read off `main`, 2026-08-21)

`ActionMenuController` (`Classes/ui/ActionMenuController.gd`, 96 lines): a `CanvasLayer` at `layer = 10` → a full-rect backdrop `Control` (any click that misses the panel = cancel) → a `PanelContainer` → a `VBoxContainer` of `Button`s, placed at `get_viewport().get_mouse_position()` and clamped `MENU_MARGIN` (6px) inside the view. `MainActionMenu` builds every menu through it — the main menu and all three submenus (Weapon Action, Transmutation, Ability Action) plus the shared attack picker.

Things about it that are deliberate, so a rebuild keeps or breaks them on purpose rather than by accident:

- **It is Control-based rather than a `Popup`, on purpose.** It replaced a `PopupMenu` because an embedded subwindow does not auto-dismiss through the `SubViewportContainer` and needed a manual `_input()` hack (`docs/BACKLOG.md` → *Action menu outside-click*; issue #26). The ring must not go back to a `Popup`.
- **`cancelled` is emitted BEFORE `action_selected` on every pick, and then the controller `queue_free()`s.** That ordering *is* today's contract — it returns the stateless verbs to IDLE and lets the targeting verbs re-set their state afterwards — and it has already produced two bugs (#105, #107), with [#421](https://github.com/Phaazoid/Godoiosis/issues/421) filed on top of it. **This is the one preserved behaviour that does not survive:** see *What persistence costs*.
- **The backdrop turns any click that misses the panel into a cancel.** That concept also does not survive — see *The interaction model*, where there is no "miss".
- **Top-level rows are omit-when-unavailable; submenu rows are listed-and-greyed-with-a-reason** (#166). Both halves survive, applied to different rings — see ruling A.
- Every row carries a hover readout (`Glossary.short`, #135) and every greyed row carries a reason (#166's law: *a menu can only grey what it can explain*).
- Buttons are `FOCUS_NONE`: there is no keyboard navigation today, and `project.godot`'s input map has no gamepad bindings at all.

**Where it draws:** in the 3D build the 2D game *is* the UI surface (#222 stage 4c — the transparent `GameContainer` covers the screen and every Control draws over the diorama). So this menu is already view-agnostic and [#292](https://github.com/Phaazoid/Godoiosis/issues/292) does not apply.

## The count, which is the sizing input

19 ids, gated by `populate()`, with real exclusions (Execute Orders needs queued orders while the four squad verbs need *none* queued; Squad Up needs no squad while Leave/Disband/Group Move need one). Reading the gates, **a squad leader with a weapon and a rune, standing on a capture point beside a downed ally and an enemy, is offered fifteen at once**; the floor is about seven. Under the design below the *inner* ring is five slices or fewer on every unit, and that fifteen is spread across the outer rings.

# The design, as settled

## A. The ring compacts

**Only live verbs are drawn, spread evenly — not a fixed ring with greyed slices.** Dev call: *"there are far too many unit dependent actions. Group move only shows for squad leader, etc."* A fixed ring would be mostly empty wedges on most units.

- **#166's deferred decision stays deferred.** A fixed ring would have brought *"a permanently full main menu is a UX change, not a readout fix"* due, along with extending the reason-first ladder (`*_block_reason` → derived boolean) to every top-level verb. **That work is not in this ticket.** Top-level gates stay booleans and `populate()` is unchanged.
- **Compaction is the MAIN RING's membership policy and nothing else's.** The submenu rule is the opposite one and it is load-bearing: a rune's un-channelable carvings stay *listed and greyed with a reason* (#166) on the outer rings. Compacting a dead carving away is a straight regression, which `tests/ui/test_menu_catalogue_rows.gd` exists to catch.
- Compaction costs positional constancy. **Ruling B buys most of it back** — a five-slice inner ring barely moves between units.

## B. Categories all the way down

**Dev call, having seen both laid out: B2, "better in every scenario".** The inner ring is categories; the verbs live one ring out.

| inner slice | holds |
|---|---|
| **Move** | Move · Group Move |
| **Attack** | the default attack · the weapon's secondaries · Reload / Rev / Burrow · the rune's carvings |
| **Act** | Guard · Rescue · Rally · Capture · Intimidate |
| **Squad** | Squad Up · Join Squad · Leave Squad · Disband Squad |
| **Turn** | Execute Orders · Wait · End Turn · Cancel Actions · Inspect |

**The Attack fold supersedes a previous ruling, deliberately.** `MainActionMenu`'s header records *"Transmutation is deliberately NOT folded under Ability Action even though #88's own text proposed it: runes are equippables and a carving is not ability use (dev call)"*. Flagged before deciding; the dev overturned it in the same breath: *"I was going to get around to grouping attack and transmutation option in the old menu anyways, I just was waiting for a full rework to get on it."* So the #88 separation was a holding position, not a principle. **Update `MainActionMenu`'s header comment when this lands** — leaving it there would leave the file asserting a rule the file no longer follows.

Under ruling A the category ring compacts too: a category with no live children is not drawn. A lone rifleman gets Move / Attack / Act / Squad / Turn; a unit with nothing queued and no squad in reach loses Squad entirely.

## C. Concentric expansion, and the ring persists

**Full outer ring around the whole circle, inner ring still drawn and still live underneath.** A branching sunburst was rejected: its room is bounded by the parent wedge, which is the opposite of the property this buys.

**Amended the same day: a deeper ring is not PAINTED as a full annulus.** Dev call: *"if we try to fill in a full radial circle for every secondary or tertiary menu, it will look ugly. I think we can stylistically build out fewer as we get further in."* Four items stretched into four 90° slabs is a heavy donut; the deeper rings should read lighter and airier.

**This is compatible with ruling E only because the DRAWN SHAPE and the HIT SECTOR are two different things, and that split has to be explicit.** The sectors still tile the full 360° — that is what "the deepest ring owns every angle" means and it is not negotiable — but a wedge need only *paint* part of the sector it owns. Four items still sit at 0° / 90° / 180° / 270°; each simply draws a narrower wedge centred on its own quarter, with air between them. Nothing about hit-testing changes; the ring just stops looking like a pie chart.

**So the aesthetic control is one number per level: what fraction of its sector a wedge paints** (plus, if wanted, the ring's thickness). That is a feel value and therefore [a knob, not a guess](https://github.com/Phaazoid/Godoiosis/issues/253) — it ships as a knob-table line in the same diff, and the HUD-shaped knobs went to `GameKnobs` / `GameTool` in #373.

One hazard this creates is already covered by ruling E's own requirement: pointing into the *air* between two painted wedges still selects the neighbouring item, because the sectors tile. That would be a lie if nothing showed it — but E already requires the current selection to be drawn at all times, so the gap never misleads. Worth noting the two rulings only stay consistent together.

**The reason given is the actual requirement, and it is stronger than "more room":** *"the whole menu sticks around until we've made a final choice, making it easy to choose again if we click wrong without having to start over."* Recoverability, not layout taste.

### What persistence costs (the structural part of the ticket)

Today's behaviour is the exact opposite, and not by accident — `_on_button_pressed` emits `cancelled`, dispatches, then `queue_free()`s **on every pick including a category pick**, and `MainActionMenu._open_menu` mints a fresh controller for the submenu. Picking Weapon Action today destroys the menu you picked it from *and* clears the selection on the way in. Three changes follow:

1. **A pick is one of two kinds, and that becomes declared.** A *category* pick grows an outer ring: no `cancelled`, no teardown, no `clear_selection()`. A *terminal* pick dispatches and tears the whole stack down. **The distinction already exists — implicitly, in which arm of `MainActionMenu.on_pressed`'s `match` you land in.** Mark it in `ACTION_DATA` and the controller reads it instead of the dispatch table implying it.
2. **One controller owns the whole stack of rings**, not one per level. `_open_menu`'s three call sites collapse into "grow a ring on the open menu".
3. **Re-picking at an inner level replaces everything outside it.** At most one ring per level; a different inner slice swaps the outer ring rather than stacking a third.

`clear_selection()` firing on a category pick is the #105 / #107 trap by name — both were state wiped on the way *into* the next mode. Dropping it from the drill-down path is the fix, not a workaround: the selection should not be cleared while the player is still choosing what to do with the selected unit.

## E. The interaction model — angular hover-preview

**This is the part that makes B2 cost nothing, and it is the most portable idea in the ticket.** Dev's design, three pieces:

1. **Hover previews the next ring.** Hovering a category draws its ring in a *not-open-yet* state (a ghost — reduced opacity, or unfilled wedges); a click commits it to open. So the player sees what is inside every category without spending a click, which is the readout property a flat ring would have given and the reason B2's extra depth is not a tax.
2. **Selection is ANGULAR, and the hit area is unbounded.** *"We don't even need the clicks to be on the menu — as long as the mouse is in the radial area of where an option would open up to on the map, they are hovering it."* The pointer's **angle** from the ring's centre picks the slice; distance does not matter. A flick in a direction is a selection, which is why this ports.
3. **Right-click goes back one ring**, so a wrong turn costs nothing and *"we never need to even see the mouse"*.

### What follows from it

- **There is no "miss", so backdrop-cancel is gone as a concept.** Every point on screen lies inside some wedge's infinite sector. The backdrop stays (the menu is modal and must keep eating input), but its job changes from *cancel on miss* to *route every click to the current angular selection*. **The cost, stated plainly: you can no longer click the board to dismiss the menu** — that click would commit whatever direction you were pointing. The replacements are right-click (back / dismiss at the top level) and a **dead zone** around the centre where nothing is selected.
- **The ring must draw the current selection at all times**, since the pointer is routinely nowhere near it. *"We should highlight what's currently highlighted."* This is not a nicety — with an unbounded hit area, the highlight is the only feedback that exists.
- **RADIUS MEANS NOTHING. The deepest open ring owns every angle** (dev call). With two rings live, one angle names both an inner category and an outer child, so something must disambiguate — and the answer is that it does not have to, because shallower rings stay *drawn* as the path you took but stop being hit targets. Right-click is the way back.

  The rejected alternative was radius-picks-the-level (pull in to re-select a category). It loses on both counts: **only the outermost level would get the unbounded hit area** while every shallower one became a thin annulus to land inside — so "distance doesn't matter" would hold for exactly one level and fail for the rest, and the ones it fails for are the ones you back out to. It is also the version that does not port, which was the point of the design: a stick holds an angle indefinitely but not a radius band, and a d-pad has no radius at all. Under this ruling a stick's angle *is* the entire input and magnitude is discarded.

  **Not a closed door:** "pull the pointer back inside the inner band" can be added later as a *mouse-adapter gesture* that emits the same `back()` right-click does. The model has `select(level, index)` and `back()`; that is an adapter learning a gesture, not a change to how selection works.
- **The seam that makes it portable is `(level, index)` with input adapters feeding it** — mouse position → angle → index is *one* adapter. A stick produces an angle directly; a d-pad or the number keys produce an index directly; a touch drag produces an angle. **Build the selection as one authority with adapters, not as hit-testing repeated per input source** (Law #4). This ticket ships the mouse adapter only, but the seam is what stops the gamepad work from being a rewrite — which is the dev's own reason for the design: *"this would make this menu scheme easy to port to other control schemes."*
- **The whole ring tree is snapshotted when the menu opens; nothing re-queries on hover.** Ghost-previewing a category means knowing its contents before it is opened, and those contents come from real queries — `RulesService.guard_candidates(unit, board)` and `adjacent_downed_allies(...)` walk the board, `get_transmutation_choices()` and `get_weapon_secondary_attacks()` walk equipment. Running them per hover-change is both wasteful and *wrong*: a preview that re-queries could disagree with what commits. Build the tree once at open. Safe by construction — the game is modal while the menu is up, so nothing can change underneath it.

## D. The ring is anchored at the cursor

**D1**, today's behaviour: the click that opens the menu *is* a click on the unit, so a cursor-anchored ring opens centred on the unit anyway. The alternative — anchoring on the unit's board position — needs a screen point for a unit, which is a *different question in each view* (canvas transform in 2D, `Camera3D.unproject_position` in HD-2D); `grep unproject_position` over `Classes/` returns **nothing**, so it would be a new seam with two answers and would put this ticket under #292.

Ruling E raises the stakes on this slightly: the centre is now the **origin of the coordinate system the whole interaction uses**, not just where the art sits. It still holds still once opened, so D1 is unaffected — but the edge clamp matters more, since the outermost ring is wider than today's panel. Let outer rings grow inward-biased near an edge rather than moving the anchor after the fact.

## The centre sprite — the seam already exists

`Unit.get_map_sprite_texture()` is already the project's answer to *"which sprite stands for this unit in the UI"* — `BaseAction.get_actor_texture()` calls it, which is what puts the actor's sprite on every action queue row. The ring's centre reads that same accessor and does **not** grow a portrait field (Law #4). The sprite in the ring's centre and the sprite on the queue rows are then the same sprite by construction, which is the point.

The centre is also the natural home for the **current selection's readout** — the `Glossary.short` text every row already carries, plus the block reason on a greyed outer-ring slice. Under ruling E that matters more than it did: the pointer is often far away, so the centre is where "what am I about to do, and why is it greyed" has to be legible. Note the centre is also the dead zone, so *clicking the sprite to Inspect* is not available — Inspect sits in the Turn group instead.

## Existing seams — please don't add a second one

- **`MainActionMenu.populate(unit)` stays the one answer to "what can this unit do".** Ruling A means it needs no change at all; the widget asks it and never re-derives it.
- **`MainActionMenu._entry(name, blocked_reason, detail)`** is the one row builder and already carries the greyed-with-a-reason contract the outer rings still owe. A slice is a row with a different shape; it should still come out of `_entry`.
- **`ACTION_DATA` grows keys rather than gaining a parallel table.** Its declaration order already IS the menu's order; it now also has to say *which group a verb belongs to* (B) and *whether a verb is a category or terminal* (C). All facts about the same list.
- **`UiLayers`** is the declared home of the UI CanvasLayer stacking order, and `ActionMenuController`'s hardcoded `layer = 10` is a fifth magic number that consolidation missed, currently tied with `UiLayers.INVENTORY_POPUP`. No live bug I can see, but the two have no declared relationship. Cheap to fix while the file is open.
- **`Resources/UiTheme.tres` is wired project-wide** (`theme/custom` in `project.godot`). Slices built from real `Button`s inherit it; hand-drawn slices opt out. Note ruling E pushes toward custom drawing — unbounded angular hit-testing is not a thing a `Button`'s own rect can do — so **the theme is opted out of either way and the slice's look becomes this widget's own business.** `ActionQueueRow._draw()` is the precedent.

## What it costs in tests

Most of the coverage is on the list, not the widget, and ruling A leaves the list alone: `test_guard_target_pick.gd`, `test_group_move_gate.gd` and `test_main_action_menu_wait_gate.gd` all assert on `main_action_menu.populate(...)` and are untouched.

What breaks or must be re-pointed:

- **`tests/ui/test_menu_catalogue_rows.gd`** reaches into `controller._button_box.get_children()` and **breaks structurally**. Re-point it at the new widget — do not delete it and do not downgrade it to asserting on the lists. Its own header is the reason: *"a test that stops short of the thing the player looks at is blind to bugs in the thing the player looks at."* It is the only thing pinning #166's catalogue law at the rendered surface, and that surface is now an **outer ring**, which is exactly where compaction must not be applied.
- **`tests/ui/test_move_replan.gd`** finds a `Button` by label text and presses it. Under ruling E the slices are very unlikely to stay `Button`s, so budget for rewriting its two helpers against the new selection API.
- **`test_right_click_cancel.gd`** pins a behaviour that is *changing*, not disappearing: right-click now means back-one-ring, and dismiss only at level 0. It needs both cases.

Three cases that do not exist and should, all of them cheap because the new model is more testable than the old one:

- **Angular selection is a pure function** — given a ring of N and an angle, which index? Assert it directly, including the wrap at 12 o'clock and the dead zone. **Sweep the full 360° and assert every degree outside the dead zone resolves to exactly one item**, which is the falsifiable form of the drawn-shape / hit-sector split: paint the wedges as narrow as the look wants and this stays true, or the split has been done wrong.
- **The unbounded hit area is the headline property**: put the pointer far outside the drawn ring, at a known angle, and assert the correct slice is highlighted and commits.
- **The snapshot rule**: open the menu, mutate the board underneath it, assert the ring did not change. That is the falsifiable form of "nothing re-queries on hover".

Ruling C's recoverability is also a visible consequence and assertable: drive a real category pick, then assert the inner ring is still alive, the unit is still selected, and picking a different inner slice replaced the outer ring rather than stacking one. A test that sets state directly instead of firing the real pick cannot see any of it — the same reason #421 records.

## Still open (does not change the shape)

**Where Intimidate goes.** It is today's only Ability Action, and #414 already promoted Guard out of that category into a verb of its own. I have put it in **Act** as another main action, but the answer probably wants deciding against whatever the next ability-driven verb turns out to be rather than against this one.

*(The look call that sat here is resolved: **a deeper ring is rotated to begin at its parent's angle**, so the group blooms out of where you pointed rather than always starting at 12 o'clock. Dev call — "let's base it from the parent". Compatible with everything above: the sectors still tile 360°, they are just rotated, so level N carries a `start_deg` taken from the slice that opened it.)*

## Notes

- **Gamepad/keyboard bindings are still out of scope** — there are none in `project.godot` at all, so shipping them is a separate ticket. What is *in* scope is not precluding them, and ruling E's `(level, index)` seam is how.
- `docs/SHORTCUTS.md` documents right-click as "dismiss the action menu" flat; that row needs updating for back-one-ring.
- Sits under [#44](https://github.com/Phaazoid/Godoiosis/issues/44) (visual clarity). Canon home is `docs/design/visual-clarity.md` — the *Tooltip / popup legibility* section, where #166's menu rulings live.
- **Sequencing with [#421](https://github.com/Phaazoid/Godoiosis/issues/421)** (auto-open the menu after a move): that ticket's subject is *when* this menu appears, and it lands on the ordering ruling C changes. Do this one first.
- `priority/P1-soon`; not on the Demo milestone — say the word if the dropdown is a hand-it-to-a-stranger problem and it should be.

**Next step is mine:** a plan, not a build — flipped to `agent/claude`. The two open placements above can be settled inside it or on the PR.

— Claude (Opus 5) · 2026-08-21

Phaazoid:
Closes [#467](https://github.com/Phaazoid/Godoiosis/issues/467).

**You are on `feature/467-radial-menu` — launch it.** Click a unit and the dropdown is a ring.

## What it is

Five categories on the inner ring — **Move / Attack / Act / Squad / Turn** — with the verbs one ring out, the unit's own map sprite in the middle, and the whole thing driven by **direction rather than by clicking on the widget**: the pointer's angle from the centre picks the slice however far away it is, hovering a category ghosts what is inside it, left-click commits, right-click backs out one ring. A category pick keeps the menu up, so a wrong turn costs one button rather than a dismiss-and-reopen.

The five rulings behind it are all yours and all recorded in `docs/design/visual-clarity.md` → *The action menu is a RING*, with the reasoning attached so the next person does not have to re-derive them.

## What I want you to feel-check

1. **The proportions.** Wide centre gap, thin bands, as asked. Every number is a knob (below) — I picked defaults, not answers.
2. **Whether Move costing two clicks bothers you.** B2 means every verb is two clicks, including the common ones; hover-preview is what pays for that, and only play can say whether it does. The escape hatch if it does not, and it is small: a category with a single live child could just *be* that child. I did not build it — it makes depth unpredictable, and you did not ask for it.
3. **The readout's position.** The hovered slice's name and reason draw *below* the ring rather than in the centre, because the centre is the sprite's. If the ring opens near a screen edge with the pointer across the map, that text is a long way from your eye. I flagged this in the plan as a look-at-it-and-say item rather than pre-solving it.
4. **Rotate-from-parent.** A child ring is *centred* on the parent's direction rather than starting its leading edge there — a deviation from the words on the issue, taken because starting at the parent's angle leaves a slice boundary directly under the cursor for a hair of jitter to flip across.

## Knobs (Game tab → **Action ring**)

`Centre gap` · `Ring thickness` · `Ring gap` · `Dead zone` · `Wedge fill` · `Wedge fill falloff` · `Preview opacity` · `Slice` · `Slice (pointed at)` · `Slice (unavailable)`

`Wedge fill falloff` is the one that spells your *"build out fewer as we get further in"* — how much less each ring out paints than the one inside it. They are `CLASS_KNOBS` statics rather than `KNOBS` rows because the menu is transient: there is no standing node for a knob to name and nothing to re-apply a change to, which is `SHOVE_SLIDE_SPEED`'s shape exactly.

## Two behaviour losses, both deliberate

- **Clicking the board no longer dismisses the menu.** With an unbounded hit area there is no "miss" — a click anywhere commits the direction you were pointing. Right-click and the dead centre are the replacements.
- **Right-click means collapse one ring first**, and only dismisses at the top. `docs/SHORTCUTS.md` updated for both.

## What broke, and one thing that broke *me*

**`#88`'s separation of Weapon Action and Transmutation is overturned**, per your call. `MainActionMenu`'s header asserted it as settled law, so that paragraph is rewritten rather than left contradicting the file it sits in.

**Re-pointing the suites caught a real regression I had shipped into the branch**: the tooltip-wrapping law (`visual-clarity` #1) went red because I had moved the wrap to draw time, leaving the row holding one form and showing another. It is back at `_entry`, the one row builder.

**And one of the three mutants PASSED, which is the finding.** Emitting `cancelled` on a category pick — the exact #105/#107 trap — did not red my persistence case, because `clear_selection()` does not null `selected_unit` (that is `exit_current_mode`, #149). The case now counts the signal and asserts the board stays in `TILE_SELECTED`; re-run against the same mutant, it reds. The other two (teaching the hit test about the paint fraction; rebuilding the tree on hover) red their own cases first time.

## Tests

- **New `tests/ui/test_radial_geometry.gd`** — pure, no scene. Sweeps all 360° and asserts every degree outside the dead zone resolves to exactly one slice *whatever fraction of it is painted*, which is the falsifiable form of the drawn-shape / hit-sector split.
- **New `tests/ui/test_radial_menu.gd`** — the unbounded hit area, the persistence-and-no-`cancelled` pair, right-click collapsing one ring at a time, and the tree being a snapshot rather than a live re-derivation.
- **Re-pointed** `test_menu_catalogue_rows` (was reaching into the dropdown's `_button_box`; now reads the ring's live level and drills in by *pointing*, since which slice an angle lands on is the thing that can break), `test_move_replan`, and the glossary law, which now asks both tables and refuses a verb whose group has no slice.
- `populate()` is untouched, which is why `test_guard_target_pick` / `test_group_move_gate` / `test_main_action_menu_wait_gate` needed nothing.

Ran locally: **`ui` + `dev` 522/522**, `law` + `runes` 165/165. Full tree is CI's.

**Embedded-content sweep:** nothing here is embeddable content — the new Glossary terms and the two `ACTION_DATA` keys are `const` tables, not authored `.tres`, so no committed fixture can predate them. Checked rather than assumed.

## Notes

- No version bump in the diff; the label carries it (`type/feature`).
- Gamepad/keyboard bindings are still out of scope — there are none in `project.godot` at all. What is in scope is not precluding them, and `(level, index)` with `aim_at` / `commit` / `back` as the vocabulary is how: another input source is another adapter, not a second hit test.
- Still open on the issue and deliberately not decided here: where Intimidate sits. It is in **Act** as another main action.


Phaazoid:
🤖 Claude says:

**Round 2 pushed — all six feel fixes plus both calls.** You are still on `feature/467-radial-menu`.

Move collapses when Group Move is not offered · Inspect is top level · the generic Attack row is gone and every attack lists by its own name · Wait moved to Act · Execute Orders / Cancel Actions / End Turn left the ring · wedges are capped so a lone option stops ballooning · labels ride the curve and shrink to fit · the sprite sits on a disc.

## Three things worth reading before the next play

**Item 2 had a trap and it was load-bearing.** `WeaponInstance.secondary_attacks()` **excludes** the main attack, so the generic Attack row was the only way to reach it — wiping it as literally written would have left a plain sword unable to attack at all. The ring now lists the main attack by name beside the alternatives, which is the same ask said properly. It also makes your naming point visible on screen: a weapon family whose main attack has no `display_name` now draws a **blank slice**.

**One rule, not two.** A category collapses when its ONE child is its own verb — Move holding nothing but Move. Deliberately not "collapse whenever there is one child", because a lone Squad Up names something the word Squad does not, and your item 5 says single-option submenus survive. **"Inspect is top level" is that same rule**, not a special case: a group of one holding the verb of its name.

**Re-pointing the tests found the suites had been lying.** The three that drove Attack emitted the id straight into the controller, which meant they were driving a row `populate()` would never have offered them — their fixtures carry no fireable weapon — and in one case the menu they drove **belonged to a different unit**: it clicked the unit's origin cell while the sprite had already moved to its queued destination. Both were invisible precisely because the direct emit skipped the menu's own subject. They now arm the fixture and click where the sprite is.

## What I decided inside your answers

**Cancel Actions left with the other two.** You said there is no Turn category and only Inspect survives at top level; Execute Orders and Cancel Actions both have HUD doors already, so both went. The one capability genuinely lost is *wipe this unit's whole queue in one press* — that is N presses of the row X now. Three words puts it back.

**The End Turn button needed no disabled state.** `_on_end_turn_button_pressed` already refused while the board is locked, so "always pressable" cost nothing: it is visible always, and the confirm fires on `faction_all_squads_acted` — **the same predicate that decides the flash**, so a flashing button never interrupts and a still one always does.

Two things measured rather than assumed while making it permanent: `MissionStatusPanel` already reserved its corner slot *even while it was hidden*, so nothing reflows; and the queue dock ends at y 490 against the button's y 676, so the old "these two are never up together" argument retired without its conclusion changing.

## Tests

**Locally: `ui` + `law` + `runes` 403/403, `dev` + `flow` + `squad` + `ai` 767/767.** Full tree is CI's.

New knobs on the Game tab → **Action ring**: `Widest wedge`, `Centre disc`, `Centre rim`, `Centre rim width`, joining the ten from round 1.

Three mutants, all confirmed: collapse-on-count-alone, no wedge cap, never-ask-before-ending. The first bit the known trap — it reddened an *earlier* case first, which truncates the rest of the file, so the case I was aiming at never ran until I disabled the ones above it.

## Still yours to judge

- **The label rotation is built my way** (top heads-out, bottom heads-in). Your wording was the other way round; your own *"to be readable"* is what settled it. One-line flip.
- **Nothing has been rendered.** Every number is a starting point.

— Claude (Opus 5) · 2026-08-21

Phaazoid:
🤖 Claude says:

**CI went red on round 2 and is green again** — worth recording what it caught, because my own plan rule would have caught it first.

Renaming `EndTurnButton.set_active` → `set_urgent` left **two callers in `tests/presentation/`**, an area I had not run. *"Every caller/reader of anything being changed, counted from the code"* is the plan requirement and I skipped it on a rename I treated as local — I grepped `Classes/` and `game.gd` and stopped. Both calls existed only to force the button visible, which is unconditional now, so they are deleted rather than renamed.

Afterwards I did the count properly instead of guessing at more areas. Zero references remain to `ENDTURN` / `CANCEL` / `EXECUTE_ORDERS`, the three `show_*_menu` builders, `begin_attack`, or `Term.TURN_ACTIONS`. The last `begin_attack` mention was a **stale comment** on `Unit.can_fire_default_attack`, explaining that gate in terms of a function that no longer exists; it now says what actually gates it.

`presentation` is 334/334 locally, and the full tree passes CI in 4m12s.

The general shape, since it is the second time this ticket has produced one: **local area runs are a filter, not the gate.** Seven areas green still missed a caller in the eighth, which is exactly the split `CLAUDE.md` describes — and the reason a rename deserves a repo-wide grep rather than a per-area run.

— Claude (Opus 5) · 2026-08-22

Phaazoid:
🤖 Claude says:

**Round 3 pushed, CI green.** Still `feature/467-radial-menu`.

## Naming (your 3 and 4, one rule)

The kit slice is named after **what is equipped** — **Weapon** or **Rune**. You floated *Channel* and then answered yourself with *"or maybe we could keep it simple, Rune and Weapon"*, and that is the version that fixes both complaints at once: Reload and Burrow are not attacks, most carvings are not either, so naming the slice for the OBJECT says whose verbs these are and claims nothing about what they do. `Act` → `Action` alongside it. Channel is one word away if you want it, but the pair would then mix a verb with an object.

**That rename re-taught yesterday's lesson within the hour.** Three suites had spelled `"Attack"` themselves and all three reddened the moment the label moved. `MainActionMenu.category_display(group, unit)` is public now as the ONE answer to what a slice is called, and the suites ask it. A label typed twice is a label that goes stale — which is Law #4 wearing a costume.

## Legibility (your 1 and 2)

**The readout is solid, on a panel, and waits.** Both colours fully opaque; the hierarchy between a name and its explanation is **brightness**, because alpha over a live board is exactly what made it painful. The delay is `gui/timers/tooltip_delay_sec` — **the project setting the inspect panel's own tooltips already read** — so *"same amount as in the inspect menu"* stays one number rather than a copied constant, and retuning it moves both surfaces together.

**Previews went 0.32 → 0.8 alpha, and ghost LABELS are no longer dimmed at all.** The label is the entire point of a preview. What still says *not open yet* is position and the absence of a highlight, not faintness.

Five new knobs — `Readout panel`, `Readout border`, `Readout border width`, `Readout name`, `Readout detail` — joining the fourteen already on Game → **Action ring**.

## Filed rather than built

Your hover-delay dev toggle is in `docs/SCRATCHPAD.md`'s inbox with the reasoning attached, so the sweep does not have to re-derive it: it would be a **second writer** of a number the inspect panel already reads, which makes it [#422](https://github.com/Phaazoid/Godoiosis/issues/422)'s dev-defaults-vs-player-settings question rather than a knob-table line — and it is arguably a **player** setting rather than a dev one, since a slow reader and a fast one want different waits.

## Tests

**`ui` + `law` + `runes` + `presentation` 740/740** locally; full tree green on CI in 5m27s.

— Claude (Opus 5) · 2026-08-22

Issue #469, opened 2026-08-22
Different attacks knock the health cubes away differently


Phaazoid:
🤖 **Claude says:**

**Dev ask, 2026-08-21, during [#314](https://github.com/Phaazoid/Godoiosis/issues/314)'s grill:** asked what a hit should do to the health cubes, the answer was *"Let's start with 1, but make an issue to have different attacks do different things to the health bar to knock it away."*

## The ask

#314 ships ONE burst for every hit: each lost cube pops up and away from its socket, tumbles, bounces once off the board and fades. This ticket is the fork — **what knocked the cubes off should shape how they leave.**

Sketches, none of them decided:

- A **heavy blow** (mace, Smash) scatters wide and slow — the cubes go tumbling off in an arc.
- A **pierce** punches straight through: the cubes fly away from the attacker along the shot line rather than fanning.
- **Fire** chars them — they leave dark, or trail, or burn out mid-air instead of fading.
- A **downing** blow might differ from an ordinary one, the way an outright death already detonates the whole grid harder.

## What #314 already built that this rides

- `HealthBlockDebris.burst(positions, facing, color, power)` is the one door, and `power` is already a per-call multiplier — the death detonation is the first caller that passes anything but 1.0.
- Every burst value is a **Game-tab knob** (`block_burst_speed`, `block_burst_spread`, `block_spin_speed`, `block_gravity`, `block_bounce`, `block_lifetime`, `block_death_power`), so "what does a heavy blow look like" can be tuned before it is authored.
- Scatter is **derived from the cube's index, never `randf()`** — deterministic, so a given hit always looks the same. Keep that: it costs nothing and keeps RNG out of presentation.

## The design question to settle FIRST

**Where does the burst shape come from?** Three candidates, and they are not equivalent:

1. **Authored on `AttackData`** — a burst-style field beside `power` and `vertical_rule`. Most expressive, and it is content rather than code. Cost: it is one more thing every authored attack can get wrong, and #314's *embedded-content sweep* rule applies the moment it exists.
2. **Derived from what the attack already says** — `elemental_damage_type` for fire, `knockback > 0` for a heavy shove, the pattern's directionality for a pierce. Nothing new to author. But Law #4's warning is exactly this shape: `knockback > 0` was already once a private predicate for a fact the readiness flags carried (#108), and re-deriving *"is this a heavy blow"* from a field that means something else is how that happens again.
[...]

Issue #188, opened 2026-08-11
No feedback when a unit takes fire damage (shake / flash / damage number)


Phaazoid:
🤖 Claude says:

**Found in play (dev, 2026-08-11, during #144's playtest):** *"there's no feedback when a unit takes fire damage. Probably should give each one a shake, a flash, and a damage number over their head, or something like that."*

## The gap

End-of-turn burn resolves invisibly. `OrderExecutor.apply_burning_tile_damage` (`Classes/actions/OrderExecutor.gd:173`) runs inside `game.end_turn()` and calls `unit.take_damage()` directly — no lunge, no tween, no number. The only evidence is the HP bar having moved by the time you look at it. An attack at least has its lunge animation; the burn has nothing, and it fires at exactly the moment the player's attention is on the turn handoff.

## The ask (dev's spec, latitude on the exact mix)

Per burned unit: a **shake**, a **flash**, and/or a **damage number over the head** — "or something like that."

## Notes for whoever builds it

- **There is no floating-damage-number mechanism anywhere yet** — nothing shows a number for attack damage either. If one is built here, it should be built as the ONE damage-presentation seam (Law #4), because attacks will want it the moment it exists; check `UnitVisuals` first for where hit presentation lives today.
- **A live pulse OWNS `sprite.modulate`** (the 2026-08-01 aiming rule: `UnitVisuals.set_highlighted` yields to it, `reset_visuals` kills it first) — a damage flash that also writes `modulate` must join that ownership rule, not fight it.
- Timing: the burn is applied inside an awaited `end_turn` before `mission_controller.check()` — feedback that awaits (a beat per burned unit?) should route through `Pacing` (one declared table, #118) and must keep its headless escape, or nine suites gain wall clock.
- Related: [#119](https://github.com/Phaazoid/Godoiosis/issues/119) (battle log) is the *reviewable* answer to "what just happened"; this is the *in-the-moment* one. They complement, not duplicate.

Filed under the [#44](https://github.com/Phaazoid/Godoiosis/issues/44) visual-clarity umbrella.

— Claude (Fable 5) · 2026-08-11

Phaazoid:
🤖 Claude says:

Two measurements from [#321](https://github.com/Phaazoid/Godoiosis/issues/321) (merged, v0.40.0) that change this ticket's starting position — both came from re-deriving #321's own body, and this issue's "check `UnitVisuals` first" note is what led to them.

[...]

PR #470, opened 2026-08-22
Health as discrete 3D cubes that fall away when lost (#314)

### LINKED ISSUE #314: Health as discrete 3D tick-blocks that explosively fall away when lost

Phaazoid:
Closes #314.

The health readout is now **one cube per point of HP, ten to a row**, and losing HP knocks the cubes that were standing out of the grid — they pop out, tumble, bounce once off the board and fade. #229's bar is replaced outright.

Every fork below was settled with you in a grill session before a line was written; this body records what was decided and what it cost.

## What it looks like

- A cube is **5 texels including a 1-texel black cage**, and cubes are pitched `block - border` apart so neighbours **share** that cage — a 3-texel coloured core in a black frame. A 20 HP unit's grid is **41 × 9 texels** against the old bar's 26 × 5, so the footprint barely moved. Width was the axis to spend: the crown hangs above and the state-icon row stacks above the grid, so only vertical growth collides with anything.
- **Ten to a row** because that is what makes it a glance rather than a count — one full row plus four reads as 14 without counting. Rows grow upward, the **bottom row fills first**, so losses show along the top where a cube has clearance to leave.
- **A lost cube is red and RECESSED**, not gone. The grid stays a full rectangle so max HP is readable from its shape, and the dent is a second cue beside the colour.
- **The cage is where the 3D read comes from, not lighting** — your words, *"each edge of the cube should be black, corner to corner."* Lighting was ruled out rather than forgotten: #229 established this display must be fog- and light-immune, so a lit gauge would change with the time-of-day preset and wash exactly as its first pass did.

## Structural choices worth ruling on in prose

- **The cubes are OPAQUE, and that is a change.** #229's five quads were coplanar with no depth to separate them, so they were `TRANSPARENCY_ALPHA` purely to buy `render_priority` ordering. Cubes have real depth, so they sort by distance like the solid objects they are and the whole priority ladder they needed is gone. This **partly closes** the open flaw #350 named — overlapping bars interleaving by layer rather than depth. The plate behind and the two labels in front still ride priority, so a far readout's *digits* can still draw over a near one's plate; the cubes themselves cannot.
- **The prediction is TWO materials, not one that swaps colour.** A doomed cube is standing and about to go; a heal-marked one is an empty socket about to be refilled. One material could not tell them apart, so every rendered count would have had to guess the direction — or read depth, which the recess knob is allowed to set to zero.
- **#313's notch is deleted** (`notch_color`, `notch_texels`, `notch_fraction()`). With one cube per point, colouring the exact cubes the plan takes says where it lands more precisely than a mark beside them; both would be one fact twice.
- **`unit_died` is the first signal `UnitMirror` listens to**, and the exception is structural rather than a preference: `die()` emits and `queue_free()`s in the same frame and `reconcile` skips a unit already queued for deletion, so the poll **never observes HP at 0**. Noticing the unit vanish instead would fire on `clear_board`, which frees without dying.
- **The cage texture is generated, not an art file** — black frame, white centre, tinted per role. Black survives the tint and white takes it, so one texture serves green, red and amber alike. That is #325's ring-outline trick reused rather than reinvented.
- **A cube is six explicit quads, not a `BoxMesh`** — `BoxMesh` lays its faces across one shared UV span, so a cage on it would stretch across the box instead of framing each face.

## Three of the ticket's own premises were stale

| #314 said | Measured |
|---|---|
| needs a shared damage EVENT with #188 | `UnitInstance.hp_changed` already exists — and isn't needed either, since the readout is told HP every frame, so the mirror diffs what it already reads. **#188 stays free to build sprite-level feedback on its own terms.** |
| "the flat view's answer is part of this ticket too" | False. The flat view has no over-unit health readout at all; `visual-clarity.md` already declared that, and #418 restates it. No 2D work. |
| (implied) the falling cubes might not be seen | Not a risk. `foretold` reads `order_executor.executing_plan`, live for the whole pass on player **and** AI passes, so a unit taking a hit already wears a readout at default settings. |

## Knobs (all Game tab → Unit HUD)

`hp_block_texels` · `hp_block_border_texels` · `hp_blocks_per_row` · `hp_block_recess_texels` — the grid
`block_burst_speed` · `block_burst_spread` · `block_spin_speed` · `block_gravity` · `block_bounce` · `block_lifetime` — the fall
`block_death_power` — how much harder a death detonates
`block_pop_time` — how long a healed cube takes to rise back out of its dent

Retired: `bar_width_texels`, `bar_height_texels`, `notch_color`, `notch_texels`.

## Verification, honestly

- **`presentation` 341/341** and **`dev` 291/291** green. Full tree is CI's.
- **Falsified with three mutants**, each isolated and each naming the case it reddened:
  - froze the baseline behind the visibility gate → reddened `test_health_lost_while_no_readout_is_up_throws_nothing` **only after I fixed that case** — see below.
  - cut the `unit_died` wire → reddened `test_a_death_detonates_every_cube_that_was_still_standing`, cases 1–3 passing.
  - drained the grid from the bottom → reddened `test_the_cubes_that_leave_come_off_the_top_of_the_grid`, cases 1–5 passing.
- **A mutant PASSED first, and that was the finding.** The baseline case originally never let the readout be up *before* the damage, so `_last_hp` was never seeded and the `.get(id, current)` default silently stood in for it — the case sailed through a mutant that froze the baseline. Fixed by hovering first (the ordinary shape of play: point at a unit, look away, an AI pass hurts it, point back), after which the mutant reds exactly that case.
- **What the suite cannot see: the feel.** Headless does not render, so the debris cases assert pool occupancy, the bounce and the decay — never how any of it looks. Cube size, cage weight, recess depth, burst speed, spin, bounce and the heal pop are all yours to judge, and all knobbed.

## Nothing embeds this

No authored content type gained a field, so the embedded-content sweep has nothing to sweep — stated rather than skipped.

## Canon + spin-off

`docs/design/visual-clarity.md` has a new #314 section, and the stale claims in the #229/#313/#350 sections are annotated in place rather than rewritten.

Spun off: #469 — different attacks knocking the cubes away differently, your ask during the grill, deliberately out of scope here.


Phaazoid:
## Round 2 — your playtest

Pushed. CI green, mergeable. **You are on `feature/314-health-blocks`** with your weapon work (`Carbine.tres`, `Super Gun.tres`) untouched, and `314-health-blocks-clone` is deleted.

### Your tuned knobs were on the working tree — they are kept

The switch refused because `UnitMirror.gd` was dirty: the Game tab's Save writes `@export` defaults straight back into the source, so a playtest with the knob panel open leaves production code modified. **Ten values, all recovered and committed** — outline 1→0, number size 0.13→0.125, burst speed 2.2→2.3, spread 0.8→0.6, spin 9→8.5, gravity 9→7, state icon 8→9, gap 2→1, spacing 1→2.

**One value I deliberately did NOT carry: `hp_block_recess_texels`, which you had at 0.** You zeroed it because the recess looked wrong, then asked for another pass at it — carried forward, the 0 would have hidden the very pass you asked for. It is back at 2.0; one drag returns it to 0 if the new treatment still doesn't land. Everything else is exactly as you left it.

(The stash holding the original is still there as `stash@{0}` if you want to check my transcription: `git stash show -p stash@{0}`. Drop it with `git stash drop stash@{0}` once you are happy.)

### Two of your eight were bugs I shipped

**The numbers.** Not a settings problem at all — the label was placed at the cube's **centre** depth, but a cube reaches a full block toward the camera, and the cubes are opaque and write depth. The text was sitting *inside* a solid and being depth-rejected. Fixed to clear the front face, and pinned as a relationship so it cannot come back.
*(Second, separate gate worth knowing: `unhovered_shows_number` is off by default, so an always-on grid shows no digits until you hover or flip that knob. Both were true at once.)*

**The heal pop.** It borrowed the **recess depth** as its travel distance — about two screen pixels at the shipped depth, and *exactly zero* at the 0 you had set. So the time knob worked perfectly and nothing moved. It has its own `hp_pop_lift_texels` now. My mistake was structural: an animation's amplitude must never be a knob that may legitimately be zero.

### The six feel items

- **A killing hit throws the red cubes too.** `burst` takes a colour per cube, so the whole grid leaves in one staggered sweep with the standing ones green and the lost ones red.
- **Bursts march.** `block_burst_stagger` (0.04s default) delays each cube, and a waiting cube **sits in its socket** rather than hiding — so the grid breaks apart in sequence with no gap running ahead of the cubes.
- **Solid tops.** You were right about the cause: the cage is a texture on all six faces, so a foreshortened top rendered as a squashed row of cells. Only the **front** face wears it now. The debris cube keeps all six, because it tumbles and a blank face would read as a hole punched in it.
- **Held in place by default**, `hp_grid_faces_camera` restores billboarding. Expect grids to go edge-on to a line at some yaws — that is what keeping them in place means, not a fault, and it is the thing to judge.
- **Recess, second pass.** Your screenshot was the useful part: the diagram I drew you had a socket with a dark inner rim, and **the geometry cannot produce that** — a cube pushed back is still a same-sized square head-on, because there is no socket wall to see. It now **shrinks** (so the plate shows around its edges) and **darkens**. Two new knobs: `hp_block_recess_shrink`, `hp_block_recess_shade`. If it still doesn't read, 0 depth loses nothing.
- **Heal animations spun off as #474** — the priest shooting them in, falling from the sky, sprouting from the ground, to try rather than choose between. The pop stays as the fallback. Both #474 and #469 are blocked on the same thing, worth knowing before either is scoped: the readout deliberately knows only that HP *moved*, never what caused it, so an arriving cube has no source to fly from until something threads that through.

### Verification

`presentation` 347/347, `dev` 291/291, CI green on the full tree. Both bug fixes falsified with isolated mutants: putting the label back at the cube's centre reds the digit case alone, and re-coupling the pop to the recess reds the pop case alone.

Still yours to judge, all knobbed: the stagger rhythm, the shrink and shade values, whether held-in-place is better than billboarding, and whether the solid tops read.

Phaazoid:
## Round 3 — the cubes read as cubes again

Pushed, CI green, mergeable. **You are on `feature/314-health-blocks`**, your weapon work untouched.

### The headline: I overshot, and the fix is your own suggestion

You asked to *differentiate* the top of a cube from the rest of the bar. I took the black cage off all five non-front faces, which is a different change — and it killed the thing the cage was there for: *"now they don't look like cubes at all, just a green mass with black painted on."*

The cage is back on **every** face, and the top is told apart by a **darker shade** instead. It works through per-face vertex colours: albedo becomes `colour × texture × vertex colour`, so the black frame survives any shade (zero times anything is zero) and only the coloured core dims. One mesh, one material — a second surface would have doubled the draw calls on every cube of every grid. New knob: **Cube top shade** (0.7).

### The z-fighting was the plate, and it was diagnosable by reading

The plate was never given a depth, so it sat at 0 — and at your recess of **0**, a cube spans 0 to a full block, putting its **back face exactly there too**. Coplanar black against green, which is exactly what you saw from behind. It now derives its depth from the cube depth and sits clear behind the deepest face any cube can present.

### Recess: not-billboarding wins, the shrink stays

You ruled it and I took it literally — depth now **defaults to 0**, and the shrink carries "this socket is empty" on its own. The knob is still there (*"I'll have to play around with it"*), it just starts out of the way.

That default change surfaced something worth flagging: `block_is_proud` reads *depth*, and at a recess of 0 proud and recessed are the same place, so it answers true for everything. Anything asking "which sockets are full" now uses `block_is_filled`, which reads the material and cannot collapse. An accessor whose answer disappears at a legal knob value isn't one a test can lean on.

### The stagger marches from the top right

*"The top right is the start of the healthbar… when we hit the second row, right side again."* Since the grid fills bottom-up and left-to-right, that turned out to be the burst order simply **reversed** — highest socket first, which gives top row right-to-left then the row below, with no second rule to keep in step.

### Filed rather than guessed: #475

Your last two notes are one question — *when* readouts are visible and *for how long* — so they're one ticket rather than a guess made inside a rework round. Worth knowing before you look at it: `Pacing.PLAYER_ACTION` is deliberately **0.0** (your call, 2026-08-10, "the player's own Execute shouldn't dawdle"), which is almost certainly why they flash in and out during your own passes. The ticket lays out the four decisions and the constraints anything there has to respect.

### Verification

`presentation` 349/349, `dev` 291/291, CI green on the full tree. Two mutants, each isolated on its target: reversing the burst order back reds the march case alone, and flattening the top shade to white reds the cage law alone.

Still yours: whether the top shade is the right amount, whether the shrink alone reads as an empty socket, and whether held-in-place is what you want now the backs behave.

Phaazoid:
## Round 4 — the background is gone

Pushed, CI green, mergeable. **You are on `feature/314-health-blocks`**, your weapon work untouched.

### Why the slider couldn't get rid of it

It was the 2D bar's own **outline**. `bar_outline_texels` drew a black quad slightly larger than the flat bar; #314 kept that quad, renamed it the plate, and gave it a second job — something for a lost cube to sink into. At an outline of **0** it is the grid's exact bounding box in black, which is why dragging the slider to zero left the rectangle standing: **zero was never *off*, it was *no margin*.**

Both are deleted — the quad and the knob. Every cube already carries its own black cage, which is what an outline was for.

You told me this in round 2 — *"don't worry about the background, they don't need one, they have outlines"* — and I read *don't worry about it* as *leave it alone*. That one was mine.

The lesson I'd keep: when a feature absorbs an older one's node, that node's knob needs re-justifying or retiring. Inheriting it silently is how a dial ends up structurally unable to say the thing its name promises — and you spent a round dragging one whose whole range was margin.

### One thing you'll see: the L

Max HP is base ± the CON band, so **18 and 22 are real**, and at ten per row that leaves a partial top row. The backing was painting black behind the sockets that *don't exist* — which reads as lost HP, except lost HP is red. So that rectangle was a small lie, and the grid now shows exactly the sockets a unit has: 18 HP is 10 + 8, in an L.

I think that's the honest read and it's what removal means. If you want the rectangle back, say so and the answer is filler sockets, not a background.

### Also moved

The state-icon row measured off the deleted quad's edges. It measures off the grid's own extent now — at your outline of 0 those were the same place, so nothing shifted.

### Verification

`presentation` + `dev` 640/640, CI green.

The law pins the **absence as a relationship** — no quad the readout draws sits behind the cubes' front plane — rather than as *there is no node called the plate*, which the next backing under another name would pass. It also subsumes round 3's z-fight case, since that fight needed something coplanar with a cube's back face and nothing is drawn there any more, so it replaced that case instead of sitting beside it.

Falsified by putting the plate back, and it reds exactly `test_the_readout_draws_NOTHING_behind_the_cubes` — with the usual caveat that a red case truncates the rest of its own suite file, so the sixteen cases after it did not run. The eight before it stayed green, which is what says the mutant is surgical.

One thing worth knowing about that case: an absence law over an empty set passes **without looking at anything**, so it drives a unit carrying a state (the state-row icons are the only quads left) and asserts it actually saw one.

Canon swept: the round-2 bullets round 3 had reversed were still standing in `visual-clarity.md` — my miss last round — so that doc now carries rounds 3 and 4 both.

Phaazoid:
## Round 5 — a heal fills in

Pushed, CI green, mergeable. **You are on `feature/314-health-blocks`**, your weapon work untouched and your four tuned values intact (cube 4.0, debris lifetime 2.35, pop time 0.54, lift 5.0 — committed as yours in `38dde3c` before any of this).

### The order was already decided

The burst leaves **descending** (round 3 — top right first), so the **lowest** socket of a run is the last one knocked out. Reversing that makes it the first one back: **ascending**. That is also the order the grid naturally fills, so a heal grows it exactly the way adding max HP would — one rule, no second convention to keep in step. A cube comes back the way it left, backwards.

### Why it couldn't stagger before

`_pop_phase` was a single tweened 0→1 that every rising cube read, so they shared one clock **by construction** — there was nowhere for a per-cube offset to live. It is `_pop_clock` now: elapsed **seconds**, tweened to a span that outlives the last delay, with each socket subtracting its own. One clock and N delays, never N tweens — the ownership rule survives (a redraw writes the target, never the clock), and killing a half-finished march is still one `kill()` rather than a hunt.

### The knob, and why it is not the burst's

**`hp_pop_stagger`, default 0.08.** Both knobs answer *how long until the next one*, but the burst's races a cube's whole **flight** while this one races a single **rise** — the same number does not mean the same thing on the two. So it is deliberately not a copy of `block_burst_stagger`'s 0.04: at your `block_pop_time` of **0.54**, 0.04 is still one blob, and a knob you cannot see is a knob that shipped dead.

That 0.54 is almost certainly why you noticed at all. At the old 0.18 the simultaneity was over before it registered.

### One call I made rather than asked

Staggering creates a state that did not exist before: a cube **waiting its turn**. It is **green** while it waits, because the material comes from the HP numbers and I left that alone. Holding it red would couple the colour seam to the animation seam and soften `filled_block_count()` — round 3's non-collapsing answer to *which sockets are full* — in the middle of a pop. At 0.08 it is a wave you will not resolve; if you crank the stagger up and small green cubes sitting there reads wrong, that is a one-liner.

### Verification

`presentation` + `dev` 641/641, CI green.

The case reads the **schedule**, not the animation. The burst's twin can ask which *slot* launched first because a slot IS its launch order — a healed cube never leaves its socket, so there is no such handle, and chasing the depths instead would have let two idle frames decide how far a tween had got. `pop_delay_of(index)` is public for that, and it is the **same derivation** `_pop_phase_for` runs on, so what the test reads cannot drift from what plays.

Falsified by reversing the delays: `test_a_heal_FILLS_IN_from_the_lowest_socket_of_its_run` reds alone, all three of its assertions firing, with the seven cases before it green. (The `steam: illegal under the two-knob rune rules` lines in that suite's output are pre-existing noise from the character editor's `before_test` — 22 of them in the green run too.)

Issue #417, opened 2026-08-20
A queued move should still offer Move: re-enter move planning instead of cancelling


Phaazoid:
🤖 Claude says:

*(From the 2026-08-20 scratchpad sweep — filed at the dev's request. Grounding is what the sweep already established; this was not re-researched.)*

## The ask

Clicking a unit whose move is already queued should still show **Move** in its menu. Picking it puts that unit back into move planning — ghost unit, path preview, the normal flow — replacing the queued move rather than requiring a cancel-and-start-over.

## What it is today

The row is **hidden**, not greyed. `MainActionMenu.populate()` gates on `not unit.has_action_type_queued(BaseAction.ActionType.MOVE)`, so once a move is queued the option disappears and the only way to re-plan is to clear the unit's actions.

## Why this is smaller than it looks

**The hard half already exists and is exercised.** By the 2026-08-02 fork (`CLAUDE.md`, `SquadManager` notes), a move is *never* refused for breaking a queued aim — the aim falls into invalid-in-red instead, because a planned heal does not lock its ally in place. So the machinery for "a move changed, re-validate everything downstream" is built, tested, and load-bearing already. Re-entry is plausibly *drop the old move, enter move mode*, and the queue re-validates exactly as it does when any move is re-planned.

## The open half — the gate's OTHER clause

That same `populate()` condition also hides Move when a **main action** is queued (`not unit.has_main_action_queued()`), and that one is load-bearing: move-before-main is a real ordering rule, enforced independently at `MoveAction.actor_can_perform()`. So this ticket has to say which it means:

- **(a)** Only relax the MOVE clause — a unit with a queued main action still cannot re-plan its move. Safe, and probably wrong-feeling for the same reason the original ask exists.
- **(b)** Relax both, and re-plan the move *underneath* a queued main action. Needs a stance on whether the main action survives (its target may no longer be reachable from the new cell) or is dropped with it.

**A greyed row would explain itself.** Since [#166](https://github.com/Phaazoid/Godoiosis/issues/166), a menu row can only be greyed if it can state *why* — `EquippableData.attack_block_reason` is the pattern. That is the shape this wants: show Move, grey it when it genuinely cannot be re-planned, and say the reason, rather than the current silent hide.

## Interaction with #412 — read this before scoping

[...]

Issue #126, opened 2026-08-02
Let a 0-damage shove move a downed unit instead of killing it, and make rescue follow them


Phaazoid:
## The design call

A downed ally should be **pushable**. Gusts of wind, kinetic shoves and similar 0-damage effects should be able to reposition a body — drag someone out of a fire, off a collapsing tile, away from an enemy that is about to finish them. That design space is closed today, and the dev wants it open (2026-08-02).

Today, hitting a downed unit **kills it, at any damage**, and the kill then cancels the shove — so the ally dies where it lies instead of moving.

`Classes/actions/resolution/LethalityRules.gd`, `predict()`:

```gdscript
	if s.lifecycle == Unit.LifecycleState.DOWNED:
		return ResolvedOutcome.Lethality.KILLED
```

That branch runs before any damage comparison — the "downed-attack = kill" fork (fork 3, `docs/design/will-and-death.md`). And a downed unit is a perfectly ordinary victim: `BoardContext.projected_unit_at_cell` has no lifecycle filter, so an attack with `hits_allies` gathers it like anyone else.

`Classes/actions/resolution/PlanResolver.gd`, `_resolve_knockback()`:

```gdscript
	if distance <= 0 or board == null or outcome.lethality == ResolvedOutcome.Lethality.KILLED:
		return   # no shove: no knockback, no board to test cells, or the target is dead (leaving)
```

So the two rules compose into: *any attempt to push a downed unit executes it instead.*

## What has to change

1. **The lethality ladder needs an exemption** so a downed unit can be hit without being finished. Shape is undecided — candidates: a hit dealing 0 damage never finishes a downed unit; or a hit from a non-hostile faction never finishes one; or `knockback > 0 and power == 0` is a distinct "reposition" kind of hit. Whatever is chosen has to stay deterministic and previewable (Law #1/#2), and it must not accidentally make *enemy* downed units unkillable — finishing a downed enemy is legal today and deliberately so (#57, fork 3).

2. **Rescue must target where the ally ENDS UP, not where it started.** Two defects, both currently unreachable *only because* downed units cannot move — opening the design space above is what makes them live:

   **(a) Rescue validation projects the rescuer but reads the target's live cell.** `Classes/squads/SquadPlanValidator.gd`, `_actor_ends_adjacent_to()`:

   ```gdscript
   	var actor_cell := projected_cell_for(action.actor, actions)
[...]

PR #478, opened 2026-08-22
The camera comes back when an order is committed (#471)

### LINKED ISSUE #471: The camera returns to the acting unit when an order is given, but not when the ring is dismissed
🤖 Claude says:

*(Dev, 2026-08-22, out of the [#467](https://github.com/Phaazoid/Godoiosis/issues/467) radial-menu playtest.)*

## The ask

The camera is free while a radial menu is open. If the player has panned away and then **commits** an order — Move, an attack, a target pick — the board work happens around the **selected unit**, so the camera has to come back to it. **Not** when the menu is dismissed without an order: panning away and backing out should leave the view where the player put it.

## The finding that decides the shape of this

**`camera_controller.pan_to(unit)` is NOT the answer, and reaching for it would produce a fix that runs and does nothing visible.** The 3D rig follows the 2D camera **only while `ai_locked`** — `tests/presentation/test_camera_follow.gd`'s own header says so: *"The follow is a MIRROR of the 2D camera (which already knows where the action is), gated on `ai_locked`."* During the **player's** turn the rig is free and driven by the player's own WASD/orbit, and the 2D camera is hidden. So panning the 2D camera on a player action moves a camera nobody is looking at.

That is this codebase's oldest failure shape — correct at both ends, wired to nothing ([#103](https://github.com/Phaazoid/Godoiosis/issues/103)) — so the work is at the **gate**, not at the pan.

**And that gate has deliberate tests around it.** Two cases in `test_camera_follow.gd` are *written to fail loudly if the gate is widened to `_board_locked_for_player()`*, because that predicate also covers MENU, where Mission Select opts out of the modal lock. So widening is not a one-word edit; those two cases are the spec for what the gate may not become, and whatever ships here has to keep them green while adding a **player-initiated** reason for the rig to move.

## Forks worth deciding before building

**A. Snap or pan?** The AI's `pan_to` is a **fixed-duration** glide (`Pacing.AI_SQUAD_PAN`) chosen so a short hop and a cross-map jump read at the same pace. A player who has just clicked an order may want to be *there now* rather than watch a beat. Fixed duration, a shorter constant, or an instant cut are three different feels and only play can pick.

**B. Which picks bring the camera back?** The honest rule is probably *"any pick that needs the board"* — Move, Group Move, the attack modes, and the target-pick verbs (Rescue, Guard, Intimidate). The ones that need nothing on the board — Wait, Inspect, the Squad verbs — arguably should not yank the view. Note the ring already knows the difference in kind (terminal vs category); it does **not** yet know "needs the board", which would be a new fact on `ACTION_DATA` if the answer is per-verb rather than blanket.

**C. Does it return, or does it follow?** `CameraController` has both (`pan_to` hands over to `follow`). A queued move that then executes walks the unit; whether the camera tracks that walk is a separate question from arriving at the unit when the order is given.

## Existing seams

- `CameraController.pan_to(unit, duration)` / `follow(unit)` — the 2D pair, used by `AIController`.
- `CameraRig3D.pose(aim, yaw, distance, board)` and `frame(volume, board)` — how the 3D rig is aimed; `battle3d` already calls both.
- `battle3d`'s `ai_locked` mirror — the thing to fork or widen, with `test_camera_follow.gd` as its guard.
- `game.selected_unit` is the stored selection (#107) — the "which unit" answer already exists and must not be re-derived from a cell.

**What I want from you:** fork **A** (the feel) and **B** (blanket or per-verb). C can follow from those.

— Claude (Opus 5) · 2026-08-22

Phaazoid:
Closes #471.

The action ring does not lock the board, so the player can pan the diorama anywhere while choosing — and then commit an order that plays out around a unit no longer on screen. A **terminal pick** now snaps the view back to the acting unit; **backing out with no order** leaves it exactly where the player put it.

## The finding that changed the shape of this

The issue said the work was at the `ai_locked` gate — widen it so the rig mirrors the 2D camera on a player action. **That was wrong, and building it would have produced a change that runs and does nothing visible** ([#103](https://github.com/Phaazoid/Godoiosis/issues/103)'s shape): `pan_to` moves the hidden 2D camera, which under `HD_2D` nobody is looking at.

`battle3d` already had a **player-initiated** rig recentre that never touches the mirror — SPACE — so this is a **second caller of `_center_rig_on(cell)`**, and the two guard cases in `test_camera_follow.gd` say exactly what they said before. Re-deriving the issue's own premise from the code is what found it ([#228](https://github.com/Phaazoid/Godoiosis/issues/228)'s law).

## The forks, and why none of them needed asking

- **Snap, not glide.** The rig smooths yaw and distance but **not position** — every writer, WASD included, assigns it outright — so a glide is a mechanism to invent, not a duration to pick. The 2D twin's `center_on_position` also sets `lock_manual_input`, refusing the player's own camera on the way in.
- **Every terminal pick**, Inspect included. The dev drew one line — *action versus no action* — and per-verb would have needed a new `ACTION_DATA` fact reaching the synthetic leaves too, on speculation. Narrowing it later is one clause.
- **Return, not follow.** A queued move executing later is a different moment.

## The change

- `game.focus_view_on(unit)` derives the cell **once** (the **projected** destination — already what every gate and pick layer means by "where this unit acts from", [#126](https://github.com/Phaazoid/Godoiosis/issues/126)), writes the 2D camera, and emits `view_focus_requested(cell)`.
- `MainActionMenu.on_pressed` wraps `_dispatch` and calls it **after** — a verb can respend the plan and move where "this unit" is ([#417](https://github.com/Phaazoid/Godoiosis/issues/417)). Hung off `action_selected`, which only a terminal emits: `cancelled` fires on commits too, so the cancel handler structurally cannot tell a pick from a back-out.
- `battle3d._center_on_pointer` generalises into `_center_rig_on(cell)`; the signal is its second caller.
- The 2D write stands down under a 3D host, and that is load-bearing rather than tidiness: `battle3d._update_pointer` snaps that hidden camera per motion to park the hover card, so writing it there would mis-anchor the card and change nothing else.

**2D/3D** ([#292](https://github.com/Phaazoid/Godoiosis/issues/292)): both views answer it. `CORNER` deliberately does not — a debug PiP whose 2D camera the pointer already drags.

## Riding along, by request: the pointer poll fires on world movement

The pick depends on the camera as much as on the mouse, and the early-out only compared the mouse — so panning under a still cursor left the hover bracket on the cell the pointer had **left** until you jiggled the mouse. True of WASD and of SPACE long before this ticket. It compares the camera's whole transform, not the rig's position, because yaw and zoom move the pick too and both lerp for frames after the input.

**One invariant this now leans on, stated because nothing in the code said it:** `_update_pointer` WRITES the hidden 2D camera and `_mirror_camera` READS it, so the two would chase each other in the same frame. They cannot, because `ai_locked` is only ever set in the same block as `AI_TURN` and the poll stands down on any locked board. `test_the_3d_camera_follows_the_ai_camera` was holding `ai_locked` **without** `AI_TURN` — a state the game cannot reach — and now sets both, saying why.

## No knobs

Nothing here is a tuned number: a snap has no value to turn. If the return reads badly in play, the candidates are (a) exempt Inspect, one clause, or (b) give the rig a `_target_position` lerp, which is a real mechanism and its own ticket.

## Verification

- `tests/presentation/test_camera_follow.gd` — 11/11, two new cases (the rig returns; the poll re-picks under a still pointer).
- `tests/ui/test_radial_menu.gd` — 10/10, two new cases (commit returns the camera; dismiss does not).
- `tests/ui` area — 242/242.
- **Falsified, three mutants, each reddening its target case and nothing else:**
  - drop `battle3d`'s `connect` -> only `test_a_committed_order_brings_the_rig_back_to_the_acting_unit` reds. Without this the wire is untested and both ends still look right.
  - move `focus_view_on` into `_on_menu_cancelled` -> only `test_dismissing_the_ring_leaves_the_camera_where_the_player_panned_it` reds. The commit case stays green, since `cancelled` fires on commits too — which is exactly the property that case exists to pin.
  - revert the poll to mouse-only -> only `test_moving_the_camera_re_picks_the_cell_under_a_still_pointer` reds.
- **NOT checked:** how it feels. Whether snapping on Inspect is annoying, and whether the instant cut reads as jarring after a long pan, are play questions.

One behaviour change worth watching for: the bracket now appears under the cursor when the camera moves, including at mission load, without a mouse jiggle.

Content sweep: nothing here adds a field to an embeddable content type, so no `.tres` fixtures were touched.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>


Phaazoid:
🤖 Claude says:

**CI went red on the first push, and what it caught is worth knowing before you review.** Fixed in `06a846a`; `tests/presentation` is 352/352.

My mistake first: I changed `Scenes/Battle3D/battle3d.gd` and then ran one presentation *file* plus the whole `ui` area. `presentation` is the area that loads `battle3d.gd`, it costs ~50s, and it would have caught every bit of this.

**The finding: the poll rider deletes a RESTING STATE that five cases were inheriting for free.** Framing a loaded board is camera movement, so the poll now lands a cell under the cursor before any case body runs — and `_pointer_cell == NO_CELL` stops being where the pointer sits at rest. Nothing asserts that as a rule; five cases just happened to start there.

Each is fixed by **stating** the precondition or by asserting the clear **where it happens** — never by relaxing the claim. The two board-swap cases are the ones worth a look: they now read `_pointer_cell` synchronously after the load funnel, because a frame later a legitimate re-pick would satisfy them *even with the clear deleted*, i.e. the clause would keep passing while losing its teeth.

**CI reported two failures and there were five.** A failing case truncates the rest of its own suite, so each fix revealed the next — the executed count climbed 328 -> 334 -> 352. Worth remembering that a CI failure list is a floor, not a count.

The production change is unchanged from the first push.

— Claude (Opus 5) · 2026-08-22

Phaazoid:
🤖 Claude says:

**Dev report, 2026-08-23** — recentring on a unit in Level_1 and in the verticality fixture put the view above them, *"and it also happens to AI turns in level 1"*, with a filed report showing Thug 7 off the frame.

**Diagnosis: pre-existing since 4d, not caused by this ticket.** `CameraRig3D._aim_at` lifts the opening shot to the top of the whole board volume — right for FRAMING a board, wrong for LOOKING AT something standing on one — and **nothing ever re-derived it**: `_center_rig_on` and `_mirror_camera` both kept `_rig.position.y`. `Level_1.tres` authors `camera_start.aim.y = 5` against a ground surface of `y = 1`, so the look-at point floated four cells in the air. Prolog is flat, which is why it never showed there.

The report's own `centred on (14, 13)` is Thug7's exact cell — the x/z aim was never wrong. The screenshot confirms the direction: screen centre lands on grass in FRONT of the thugs, which is what raising a pitched camera's look-at point does.

**Fixed at both surfaces that read the rule**, per the dev's call to keep it in this PR:
- `_center_rig_on(cell)` takes `BoardSpace.surface_point`'s whole answer — the seam `UnitMirror` places the sprite with (#273), ramps included.
- `_mirror_camera` takes the continuous twin `surface_height_at` (#259): an AI pan is a GLIDE, and a height that stepped at cell boundaries would jolt the diorama mid-beat.

**Two suite cases had pinned the bug AS CORRECT** — mine from yesterday (`"the return moved the rig's height"`, expected value read off the pre-state) and `test_space_recentres_the_diorama_on_the_pointer` from 4d. Both now invert. Worth naming because falsification structurally cannot catch it: a mutant proves a test has teeth for the behaviour it asserts, and both assertions were faithful — the behaviour was wrong.

**Declared residual, not built:** WASD panning still keeps the last recentre's height, so panning from a valley onto a hill re-opens a smaller version of the gap. Making the aim track terrain continuously means teaching `CameraRig3D` about the board, which it deliberately does not know (LookDev shares it and has no board). Its own question.

**Verification:** `presentation` 354/354, `dev`+`flow`+`law` 636/636 — every area that loads `Battle3D.tscn`, checked by grep this time rather than by guess. Two mutants, each reddening only its own surface (revert `_center_rig_on`'s height → the recentre case; revert the mirror's → the AI case), each confirmed past an earlier case that also reds.

— Claude (Opus 5) · 2026-08-23

Issue #245, opened 2026-08-14
Terrain-state icons float over erased cells in 3D (anchored to a top face that isn't there)


Phaazoid:
🤖 Claude says:

Found in play by the dev while feel-checking #231 PR 2. **Not a PR 2 regression** — reproducible on `main` today.

## Symptom (dev's words)

> Frozen tiles get a weird blur over them when they are over nothing (the tile under is empty). This blur isn't over the entire tile, and seems to move a bit when I move the camera, uniform for each tile.

## Cause

`OverlayMirror._terrain()` renders live terrain-state icons (FROZEN / COVER) onto `BoardOverlays.Layer.TERRAIN`, which is `Kind.SPRITE` — **a flat ground quad**, anchored by `_anchor_px()` at `OverlayMirror.TOP` (`= UnitMirror.COLUMN_TOP`, y = 1.0).

y = 1.0 is the height a block's top face sits at on a flat mirror board. On a cell that *has* a block the quad lies flush on that face and reads correctly. On an **erased** cell there is no face — so the quad hovers in mid-air over the hole. That accounts for all three observations:

- **"not over the entire tile"** — a flat quad seen at the camera's pitch, with no surface behind it to read against.
- **"moves a bit when I move the camera"** — parallax. Nothing at that depth anchors it visually.
- **"uniform for each tile"** — every marker uses the same constant anchor height, so every one is wrong identically.

## Repro (no 3D authoring needed)

1. In the flat 2D view, paint a FROZEN tile state onto a cell, then erase that cell's terrain.
2. `F4` into the 3D view.
3. The frost icon floats over the hole.

## The actual question

The render is faithfully mirroring the 2D, which *does* draw a frost icon over an empty cell — 2D simply has no depth for it to look wrong in. So the mirror is obeying "mirror the QUESTION", and the latent gap it exposes is a **rules** one:

**Should a tile state exist on a cell with no ground at all?** A FROZEN cell with nothing to freeze arguably should not be representable.

Three ways out, in rough order of how honest they are:

1. **Rules fix** — forbid a tile state on an empty cell at `TerrainStateManager.apply`, the one deposit seam. Both views agree afterwards and the render bug disappears with no render change. Needs a caller count on that seam first (fix-the-rule law), and it changes 2D behaviour.
2. **Erase clears state** — a narrower version: erasing a cell's terrain also clears any tile state on it. Fixes the reachable path without forbidding anything.
[...]

Issue #481, opened 2026-08-22
Dev-tool saves silently drop a resource's UID


Phaazoid:
🤖 Claude says:

*(Flagged out of [#473](https://github.com/Phaazoid/Godoiosis/issues/473) and deliberately not folded into it — that ticket was the weapon editor's readability, this is every dev tool's save path.)*

## What happened

Editing the Carbine's main attack in the Attack Editor and pressing Update wrote the file correctly **and silently dropped its UID**:

```diff
-[gd_resource type="Resource" script_class="WeaponAttackData" format=3 uid="uid://dbboy21d8bwwo"]
+[gd_resource type="Resource" script_class="WeaponAttackData" format=3]
```

`Resources/Weapons/MainVarieties/Carbine.tres` still names that UID:

```
[ext_resource type="Resource" uid="uid://dbboy21d8bwwo" path="res://Resources/WeaponAttacks/MainAttacks/Carbine.tres" id="2_main"]
```

So the family template now points at a UID no file claims. Godot degrades to the `path=` attribute and loads fine, which is exactly why nobody noticed — the symptom is the `invalid UID: ... using text path instead` warning class, plus a reference that **breaks outright if the target is ever moved or renamed**, since the path fallback is the only thing holding it together.

## Scale, measured on `main` @ v0.79.1

**68 of 92** tracked `.tres` under `Resources/` carry no UID, and the concentration follows the tools exactly:

| folder | lacking a UID | written by |
|---|---|---|
| `Resources/LookPresets/` | 12 / 12 | Moods tab |
| `Resources/Weapons/WeaponVariants/` | 8 / 8 | Item Editor |
| `Resources/TransmutationData/` | 6 / 6 | Attack Editor |
| `Resources/RuneVariants/` | 7 / 8 | Item Editor |
| `Resources/Units/` | 8 / 10 | Character Editor |
| `Resources/WeaponAttacks/` | 8 / 12 | Attack Editor |

**44 files carry `ext_resource` lines that name a UID** — those are the references with something to lose.

Most of these were *born* without a UID, which is harmless in itself. The Carbine is the case that matters, because it shows a **committed, UID-bearing, referenced** resource losing its UID on an ordinary Update.

## Where

`DevWidgets.save_over(resource, path, status_label)` is the single writer every dev-tool save routes through (Attack Editor, Item Editor, Character Editor, the Moods/Game/Objects tabs, ScenarioManager):

```gdscript
[...]

PR #492, opened 2026-08-24
#427 slice 4: the corner-drag tool

### LINKED ISSUE #427: Terrain model migration: per-corner heights, half elevations, and a corner-drag tool (RCT-style ground)
🤖 Claude says: *(filed from the dev's note at the #258 wrap-up, 2026-08-20 — not written by the dev directly)*

**The symptom (dev):** *"when two slopes are next to each other at different angles, we're getting weird cut off space."* Two adjacent ramp cells whose rises point different ways render as two independent wedges, and the seam between them leaves a visible gap/cut-off volume — there is no transitional form to fill it.

**The ask (dev):** *"I'll want blocks that are corner slopes, etc, to help with that"* — richer ground geometry beyond the five-state `Terrain.RampRise` vocabulary (NONE + one cardinal rise per cell).

Two halves, deliberately separable:

1. **Presentation** — the mesh forms. `gen_lookdev_assets.gd` emits one wedge variant per flat tile (#340); a corner slope needs new forms (outer corner, inner corner, possibly a diagonal saddle) emitted per flat tile the same way, and `BoardMirror.ramp_item_for_cell` picking them. Worth studying how the #263 wall forms solved the analogous problem: a facing became a MASK of edges rather than a yaw, and a corner fell out for free as an L. A corner *slope* may want the same shape — which two edges rise — rather than a new enum member per diagonal.
2. **Rules** — only if a corner slope is *traversable* as a corner. `RampRise` cannot say "rises toward two sides", and `RulesService.can_step`'s ramp clauses + `height_step_ok` (melee's half step, #258) + `BoardSpace.lie_on`'s tilt + the tumble rule (#259, when it lands) would all need to learn the new vocabulary. The deferred-list note on 30-degree two-tile slopes in `docs/design/verticality.md` states the same warning for that experiment: extending the ground vocabulary extends every rule that reads it.

A purely-cosmetic corner (renders as a corner, RULES-wise stays a cardinal ramp or a flat cell) would be the cheap first slice if the gap is the only real problem; a traversable corner is a design session. Needs the dev's ruling on which of those this is before any build.

Related: #340 (ramps wear their ground), #263 (wall forms / edge masks — closed), #281 (`surface_transform`), #259 (tumbles will read the ramp vocabulary), #266 (the 2D view will eventually need the same forms as sprites).

— Claude (Fable 5) · 2026-08-20

Phaazoid:
Closes #427.

The Tile Brush gains a fourth paint mode. It drags the **point four cells share** to the level
picker's height, welding them — where the cell brush paints one whole tile and can still author a
deliberate cliff between neighbours.

> *"we would want to keep the version of the tool we have now, but also have a corner mode.
> Different tools for different situations."* — 2026-08-23

Nothing about the existing brush changes. Slice 3 shipped the corner model with nothing able to
author one; this is that gap, closed.

## The two rulings

**Absolute, not relative.** The wheel picks a height and every point the drag touches goes to it,
exactly as the cell brush places a level. `_paint()` re-fires on every mouse-motion event while the
button is held, so a relative `+1` runs away down a stroke where an absolute value drag-paints
idempotently. Right-click pulls the point back down to the board floor — the corner reading of "take
it away", symmetric with paint rather than a second gesture.

**One ring, clamped.** A drag writes the point's corner in all four touching tiles. Anything that
would make a touched tile illegal is clamped, never cascaded outward.

`Terrain.corner_toward` **asks `is_legal_corners`** rather than restating which of its three refusals
it tripped — so saddles, the 45° cap and three-height cells stay refused in one place, the tool
cannot grow a second opinion about any of them, and a rule added there reaches the tool for free.
It walks from the target toward the corner's *current* height, which terminates by construction: the
cell is already legal, so the worst case is the corner not moving.

The consequence is the ruling working as intended rather than a shortfall: **a corner stops rising
once its own tile would break**, so a tall hill takes a few strokes across neighbouring points
instead of one drag that reaches across the board. That is the trade for never having to reason
about how far one drag went.

## Shape

- `Terrain.VERTEX_CORNERS` — the one cell-offset → corner-bit table. `CORNER_COMPONENTS` is its
  bit → `Vector4i`-component twin, the write direction of `corner_mask`'s read; a hand-written
  `match` at a write site would be a quarter of every drag landing on the wrong corner.
- `BoardHeights.set_corner` / `set_vertex` — the per-cell composer and the weld, both writing through
  `set_corners`. The weld takes the ground predicate as a parameter, on `prune_groundless`'s own
  justification: a point dragged at the board edge would otherwise leave heights in cells with no
  tile (#245), and the store cannot reach the grid to ask.
- Corners stay **per-tile**. The store is untouched and this is still not a shared vertex grid — a
  vertex is what the *tool* addresses.
- Undo came free: a stroke is already bracketed and `BoardSnapshot` learned corners in slice 1. There
  is a case for it anyway, precisely because "free" means nothing here was written for it.

## Two traps paid

**The pointer's answer changes halfway across a cell**, so `battle3d._update_pointer` computes the
vertex *above* its cell early-out. #471's law: an early-out is a copy of the render key on the input
side, so it has to compare every input the answer depends on. Below it, the marker sticks to whichever
corner the cursor entered the tile by. Falsified by moving the line down.

**`BrushGhost` gained a declared `Kind`** rather than letting a renderer infer the mode from an unset
`source`. The 2D hide gate reads the kind *before* the source, and that ordering is load-bearing: a
source compare alone matches `null` against a ghost layer nothing has built yet, and paints a tile at
the origin.

Declared approximation: the sub-cell half of the pick is recovered by dropping the same ray onto the
picked cell's *top plane*. Exact on flat ground — which is what a hill is built out of — and off by a
fraction of a cell on a slope at a grazing angle. The marker shows the answer before any click
commits it, and the alternative is a per-form ray/triangle intersection in the hot pointer path for a
dev tool.

## Two things found while wiring it

- **`_climb_row` was never toggled.** It shipped with slice 2 and stayed visible in Zone and State
  mode, greying itself but describing nothing. It joins the row-visibility table here.
- **Z/X/C and the wheel were one predicate**, because the level and rise rows had always appeared
  together. Corner mode shows one and not the other, so `_elevation_brush` (the wheel) and
  `_shape_brush` (the keys) are now two. The rule underneath — *a control only moves a brush you can
  SEE* — is unchanged.

## Knob

`BoardMirror.brush_vertex_ghost_size` (Game tab → Dev chrome, beside the ghost alpha) — the marker
cube's edge as a fraction of a cell. 0.22 is a guess to feel-check, not a decision.

## Verification

Areas: `terrain` 67/67 · `dev` 326/326 · `presentation` 370/370 · `rules` 60/60. Full tree is CI's.

**Falsified, five mutants, each caught by its intended case with its intended message:**

| mutant | caught by |
| --- | --- |
| transpose NE/SW in `VERTEX_CORNERS` | the weld case |
| clamp's predicate always true (returns the raw target) | the legality sweep |
| clamp steps *away* from the tile | the nearest-value case |
| drop the CORNER arm from `_paint()` | the wire case |
| vertex computed *below* the cell early-out | the pointer case |

`test_no_reachable_drag_can_author_illegal_ground` is the exhaustive one — every legal starting
shape × every corner × every target in and beyond range, asserting the result is legal. It is what
makes "structurally unable to author a saddle" true rather than intended.

**One test rewritten rather than extended**, and it is the reason to mention it: `test_height_brush`
had two cases hand-listing "the other paint modes" as `[ZONE, STATE]`. A fourth mode makes a
hand-written list a second answer that goes stale silently — so both now sweep *every* mode and derive
the expectation from the row's own visibility.

**No embedded content touched.** Nothing new was added to an embeddable type; `corner_heights`
already round-trips from slice 1 and no `.tres` shape changed.

## Canon

`verticality.md` gains the corner-tool record and loses the two "until slice 4" gaps; its
`ramp_rise_at` TRANSITIONAL note was stale from slice 3 and is corrected. `SHORTCUTS.md` gains the
mode across the drag/erase/wheel/Z-X-C rows. `CLAUDE.md`'s #427 entry gains the authoring half.
Marker bumped to #491.

## Yours

**You are on `feature/427-corner-tool` — launch it.** Paint a hill, then tell me what the tool is
missing. Your ruling on paint ergonomics was that the friction was *"from not having that mode, so we
can revisit specific pains after that exists and I'm using it"* — so this ships the mode and no
ergonomics beyond it.


Phaazoid:
## Follow-up pushed — markup on corner cells

Three symptoms, one root cause, plus a fourth you hadn't seen.

`BoardSpace.lie_on` promises *"how flat markup lies on a cell"* and answers with a `Transform3D` — a
PLANE. A corner cell's four surface points aren't coplanar, so no transform can keep that promise.
Slice 3 shipped a best-fit plane and my own comment called it *"the average the markup wants"*. It
isn't. Measured on the branch:

| climb | error at every corner |
| --- | --- |
| half (26.6°) | ±0.125 world = **⅛ cell** |
| full (45°) | ±0.25 world = **¼ cell** |

The sign alternates — two corners above the ground, two below — so the quad *crosses* it.
`fill_lift` is 0.02, six times smaller, which is why no lift tweak could have helped. Three of the
four corners sit on the flat half, which is exactly where you saw it.

**The fourth thing, found while measuring and not reported:** `surface_point` — where a unit, a flame
or a prop stands — read the corners' MEAN, while `surface_height_at` read `Terrain.height_at_uv`.
On a corner cell they answered a quarter of the climb apart, so anything standing on one floated.
Slice 3 declared `height_at_uv` *the one surface*; the mean was a second answer to the same question
and is now deleted.

### The fix

**Origin:** `lie_on` reads `height_at_uv`. Exact on every planar form, so no flat cell and no
cardinal ramp moves by so much as a float.

**Basis:** it returns identity on a corner cell rather than a tilt it cannot honour, and the FOLD
moved into the marker's mesh — four vertices at their true heights, split on the diagonal
`height_at_uv` splits on. *The mesh calls the query*, the same law the cap meshes follow, because two
triangulations of a quad meet at all four corners and differ only inside. Planar cells keep the one
shared `PlaneMesh`.

Re-ran the probe: **every legal form now sits exactly on the ground**, at all four corners and at the
centre. Zero, not "small".

### Verification

`presentation` 382/382 · `terrain` 68/68 · `rules` + `dev` 386/386.

Five mutants, each caught by its intended case — restore the mean, force PlaneMesh's fixed SW–NE
diagonal, flat quad everywhere, keep the tilt and double-count the fold, drop the `corners` key.

**The fifth one PASSED first time, and that's the finding.** Deleting the key from
`OverlayMirror._marker` — the wire that carries the fold from the mirror to the renderer — left all
40 cases in that suite green. I'd pinned the renderer's *consumption* of the key and nothing pinned
its *production*: both ends correct, no test on the wire, which is #103's shape. There's now a case
at the mirror end, and it reds.

**You are on `feature/427-corner-tool` — pull and launch it.** Same hill, same three views.

One thing in your working tree that isn't mine: `Scenes/LookDev/lookdev_meshlib.tres` shows as
modified, with Godot having re-added the texture `uid=`s on open. That's the benign direction of the
#481 residual, not damage — commit it or discard it, either is fine.

Issue #102, opened 2026-07-28
Kinetic Mace: firing Blowback permanently removes the unit's Attack option (stale active_attack)


Phaazoid:
🤖 Claude says:

Filed from `docs/SCRATCHPAD.md`. Reported as *"Kinetic Mace — when using blowback, main attacks are disabled."* Root-caused; it is worse than the title suggests and it is **not** mace-specific.

## What actually happens

Once a unit picks **Blowback** even once, the top-level **Attack** entry disappears and never comes back for the rest of the battle. The mace's Weapon Action entry goes with it, so the unit has no attack surface at all — a permanent offensive soft-lock on that unit.

## Root cause

`Unit.can_fire_default_attack()` (`Classes/units/Unit.gd:580`) does not read the default attack:

```gdscript
func can_fire_default_attack() -> bool:
	return is_attack_fireable(get_fired_attack())
```

`get_fired_attack()` (`Unit.gd:507`) returns `active_attack` whenever one is set, falling back to the default only when it is null. So the gate on the Attack menu entry (`MainActionMenu.gd:193`) asks *"is the last thing I aimed with fireable?"* when its own name, its doc comment, and the action it guards all say **default**. Pressing Attack calls `begin_attack()`, which resets `active_attack = null` and fires main — so the gate and the thing it gates are asking different questions.

That mismatch is inert for every other family, because their secondaries stay fireable. The mace is the first weapon whose secondary *disarms itself on use*:

```gdscript
func consume_readiness_for(attack: WeaponAttackData) -> void:
	if _is_blowback(attack):
		charge = maxi(0, charge - 1)
```

## The trap closes

1. Weapon Action → Blowback. `_pick_attack` (`MainActionMenu.gd:283`) sets `active_attack = Blowback`.
2. It fires. `consume_readiness_for` spends the charge → `charge == 0`. `active_attack` is **still Blowback** — resolving an attack never clears the pick.
3. Next turn, `populate()` line 193 gates Attack on `can_fire_default_attack()` → `is_attack_fireable(Blowback)` → `charge >= 1` → **false**. Attack vanishes.
4. `has_weapon_actions()` (`Unit.gd:531`) also returns false — Blowback is the mace's only weapon action and it isn't fireable at 0 charge. Weapon Action vanishes too.
[...]

Issue #77, opened 2026-07-20
Deterministic limb selection: maim rotation + aura-tax tiebreak, beyond a fixed order


Phaazoid:
🤖 Claude says:

**Context.** Split out of [#60](https://github.com/Phaazoid/Godoiosis/issues/60) (the transmutation doctrine catch-up) at the dev's request, 2026-07-20. Item 5 there proposed one narrow fix — the aura-tax tie-break (`UnitInstance._apply_maim_aura_tax`, currently element-enum order) switches to primary affinity once #60's affinity set landed. The dev is pushing that off: not because primary affinity is wrong, but because it's just *one* candidate answer to a bigger, already-flagged, still-open question — and there are other candidates (character "humors," among others) that haven't been explored at all yet. Rather than lock in a stopgap now and redo it later, this issue exists to hold the whole question until it gets a real design pass.

**This isn't new — it's a promotion.** `will-and-death.md` already flags the *maim rotation itself* (which of the four limb slots gets taken, not just the aura tie-break) as "VERY subject to change" (dev, 2026-07-15), with candidates already on the table: primary affinity/humors (story layer), handedness (the `ARM_R`-first convention is arbitrary — handedness isn't modeled), and the punted damage-source-derived idea (melee → arm, terrain/AoE → leg). This issue formalizes that flag into a tracked item and folds the aura-tax tie-break in as the same question asked twice over — both are "given a deterministic trigger, which limb/pool is it" and a shared answer might cover both, or they might legitimately diverge.

## The two mechanisms in scope

1. **`UnitInstance.MAIM_ROTATION`** — which limb slot gets taken when a down can't afford its Will cost. Currently a fixed sequence: weapon arm → off leg → off arm → weapon leg (naturals first, prosthetics last-and-recoverable). Read live via `next_maim_slot()`, previewed on the unit panel ("next at risk," Law #1).
2. **`UnitInstance._apply_maim_aura_tax`** — which elemental aura pool loses a point on every maim. Currently: the highest pool, ties broken by `Elemental.Element` enum declaration order (`FIRE > WATER > EARTH > AIR > AETHER`) — a placeholder with a literal `TODO` marker in the code pointing here.

## Candidates already on the table (not a decision, just the roster so far)

- **Primary affinity** (#60's original item-5 proposal) — whichever element ranks first in the unit's ordered affinity set.
[...]

PR #535, opened 2026-08-26
Recover two scratchpad entries that only existed in a stash

Phaazoid:
`stash@{1}` ("On main: scratchpad-3-entries") held three of your inbox entries that were never applied. Two are still nowhere else in the repo:

- **A bug report**, verbatim: *"In Castle Assault.tres, I made both player squads AI/Rushdown and the Enemy squad AI/HOLD, and the game played itself into a broken state once the enemy leader was the only remaining enemy unit left."* This has never been filed anywhere. Flagged as issue-shaped rather than idea-shaped, with a note that it looks like the #103 shape (a squad turn that cannot terminate) reached by a different route — worth checking before diagnosing from scratch.
- **The limb-loss idea** — direction faced / attacked from, or damage type, influencing which limb goes. Pointed at #77, which already holds that rotation as provisional.

The third entry (Kinetic Mace blowback disabling main attacks) is **already on main** in the Dispersed log, root-caused as #102, so it is deliberately not re-added.

Docs-only, so CI skips it by `paths-ignore`.


Phaazoid:
Closing unbuilt: all three entries in `stash@{1}` were **already swept on 2026-07-28** and I missed it.

- Kinetic Mace blowback → **#102** (Dispersed log, root-caused)
- Castle Assault AI deadlock → **#103**, which is literally this report, filed from the scratchpad quoting the dev with the same board, the same Rushdown/Hold setup and the same last-enemy trigger. Closed 2026-07-30.
- Limb loss by facing / damage type → **#77** + `will-and-death.md`, Dispersed log line 86.

So the stash is a stale pre-sweep copy of the Inbox, not lost work. Re-applying it would have put two solved bugs back in the Inbox and duplicated the audit trail.

My error was grepping the **Inbox wording** ("Castle Assault.tres", "Limb loss choice") against a Dispersed log that had already rewritten each entry into its own one-liner form — grep the question, not the name.

Issue #520, opened 2026-08-25
Battle zoom: the camera director, on the live board


Phaazoid:
Slice 3 of #410 (battle zoom). **Live board, zero new art** — this is where the feature starts to look like the pitch.

## Ask

Point the existing camera rig at slice 1's beat sheet:

- **First attack: side-on profile shot.** Between interactions, a **panning spin** — orbit to the next pair's natural profile angle.
- **One volley = one tracking shot.** An AoE hitting three victims is a single sweep along the line as they are struck in sequence, not three cuts. `AttackAction.volley` already groups them.
- **Knockback follows.** Normal framing keeps no dead space behind a unit; when a gust launches someone the camera pans *with* the flight. `ResolvedOutcome.knockback_path` carries the whole route and `knockback_landing_index` the touchdown.
- **Blown off a cliff:** the camera leaves the scene and follows them all the way down, shuddering with the impact. `fall_levels` and `elevation_delta` distinguish a drop from a void removal (#218 semantics).
- **Lethality-aware direction:** the rung is known *before* playback, so a killing blow can wind up, freeze mid-swing, and dolly in.

## What exists and what does not

`CameraRig3D` is an **orbit** rig — `pose(aim, yaw_degrees, distance, bounds)`, `frame(volume, bounds)`, per-frame smoothing, free yaw with 90-degree detents. Profile shots and the panning spin map onto it directly.

Missing, and in scope here: a **keyframe/segment layer** over `pose()` (CameraPose's header says it was built to extend into keyframes — #234), a **pitch driver** (the Pitch node under CameraRig exists; nothing drives it), and a **shake**.

## Done when

- A pass plays through with directed shots on the live board, driven entirely by the beat sheet.
- Manual camera input is suppressed for the duration and restored after — including on skip.
- The rig's existing contract (detents, pan limits, `rebound`) survives; a law proves the camera returns home.

## Notes

- An `_ai_owned_camera` flag already exists on `battle3d.gd` — check it before inventing a second ownership seam (Law #4).

Phaazoid:
**Notes from playing #519 (dev, 2026-08-26).**

**1. The camera must go to the action, and the action must WAIT for it.** *"Currently, the camera isn't zooming to where action happens. It will need to do that in either version. And action will need to wait for the camera to get to it."* — **in either version** is scope: this is not zoom-only. The `BOARD` profile pans too; the profiles differ in framing and how long a beat holds, not in whether the camera moves at all.

[...]

Issue #484, opened 2026-08-23
Toggling dev mode during an AI turn unlocks the board while the camera stays AI-owned, welding the view to the mouse


Phaazoid:
Reported in play 2026-08-23 (Discord report `2026-08-23_11-56-54`, v0.79.2, `main @ f23ac67`, no note typed). The dev's words:

> Basically, mouse movement got tied to camera position, and moving the mouse in a direction would zoom the camera to its limit in that direction. It might have been caused when entering/exiting dev mode.

The hunch is right, and the mechanism is a Law #4 drift.

## Two answers to one question

"Does the AI own the board right now?" is answered twice, written in one block and hand-synced:

- `game_state == GameState.AI_TURN` (`game.gd:511`)
- `camera_controller.ai_locked` (`game.gd:512`), cleared only after the awaited `take_faction_turn` returns two lines later

`set_dev_mode` resets the first without consulting the second: `exit_current_mode` -> `clear_selection` -> `game_state = _base_state()`, which only ever answers `DEV_MODE` or `IDLE` (`game.gd:688`, `game.gd:723`). F1 reaches this with no lock guard, deliberately -- dev keys are `PROCESS_MODE_ALWAYS` and fire behind locks since #154. Toggling dev mode **off** lands on `IDLE`, which is equally unlocked, so both directions do it. `ScenarioManager.clear_board` is a second path, since it calls `exit_current_mode` too (`ScenarioManager.gd:380`).

## What that opens

The board then reports *unlocked* while the camera still reports *AI-owned*, which is the one combination that lets two gates write to each other:

- `battle3d._update_pointer` snaps the hidden 2D camera to the hovered cell, gated on the board **not** being locked (`battle3d.gd:811`)
- `battle3d._mirror_camera` reads that same 2D camera and writes the 3D rig, gated on `ai_locked` (`battle3d.gd:520`)

Since #471 the pointer poll re-derives on camera movement as well as mouse movement, so each mouse move snaps the 2D camera, the rig follows it, the pick re-runs against the moved camera and lands further in whatever direction the cursor sits off screen-centre. It marches until `pan_limit` clamps -- exactly the reported symptom, and self-sustaining once in that state.

[...]

Issue #410, opened 2026-08-20
Battle zoom (umbrella): cinematic squad-combat replay - the fight rises off the board


Phaazoid:
## Pitch

When squads clash, the board tears itself apart: every tile under the combatants rises out of the world — up, up, off camera — white-out, and the fight fades back in as a floating diorama, the camera swooping from interaction to interaction. Fire Emblem's battle zoom rebuilt for squad combat: **everyone in every participating squad is on stage.** Design pillars, in the dev's words: *"Dynamic and Spectacle are what we're going for"* — *"we should go wild with this."* This should be the highest-spectacle piece in the game; the board's HD-2D work sets the floor it must clear.

This is an **umbrella**: the full vision lives here, sub-tickets get listed below when split off, and a listed child is deleted when done.

## Architecture ruling (settled)

- **Pure replay. The zoom is a mirror, never an authority.** Combat resolves fully before the camera rolls (Law #2); the cinematic consumes the already-computed `ResolvedPlan` (`Classes/actions/resolution/ResolvedPlan.gd`): ordered per-victim `AttackAction`s grouped into volleys, then counters, each carrying a full `ResolvedOutcome` (damage, hp before/after, knockback from/to, states, lethality rung). Nothing is recomputed; skipping the cinematic changes nothing.
- **Cinematic unit = one squad's plan**: moves, then attacks aim-by-aim in queue order, then all counters. A plan can strike any number of opposing squads (long-range AoE, transmutations), and every attacked party counters once — so the diorama holds **all participating squads**, and the field can get large.
- **Natural seam already exists**: `OrderExecutor._execute_action_sequence` awaits each action individually, `OrderExecutor.executing_plan` publishes the pass, and `Pacing` is the beat table (with headless escape). The zoom re-times and re-frames existing playback; it never changes what plays.
- **Display-only randomness is allowed** (staging variants may differ randomly). Law #1 (no gameplay RNG) untouched.
- **3D-view-only by design** — declared #292 asymmetry. The hidden 2D authority plays through as today; the flat view simply doesn't zoom.

## The transition

- Entry: participating tiles rise out of the board, off the top of the frame; screen whites out; fade into the diorama.
- Entry choreography: tiles slam into the diorama one by one **in queue order** — quietly foreshadowing who acts first.
- Exit reverses it: white-out, tiles thud back into their board sockets with a dust shockwave, the aftermath already written on the health bars.

## The diorama (settled: true 3D)

[...]

Issue #47, opened 2026-06-21
Attacks target cells, not units


Phaazoid:
## Summary
Attack targeting is currently unit-centric — `gather_attack_victims` resolves the affected cells to the **units** standing on them, and the attack legality gate requires at least one unit in the blast (`if not victims.is_empty()`). This blocks a core elemental interaction: aiming at terrain. We want the player to throw a fireball at an **empty tile** (e.g. one with only a tree) to start a fire / create a terrain state.

## The change
Make attacks target **cells**, with units (and terrain) being consequences resolved from the targeted cells — for **every** pattern type: directional (forward line / wide), AoE, and Manhattan / point. An attack aimed at a tile with no unit is still a legal order; resolution applies whatever the cells produce (unit damage, terrain ignition, elemental state on the tile, etc.).

This is the natural extension of the resolution pipeline (#5) and the terrain / elemental model (`docs/design/terrain.md`, `docs/design/elemental-system.md`): cells own persistent terrain state; an attack deposits effects onto cells; units on those cells take the unit-facing consequences.

## Touch points (today)
- `game.gd` attack click — the legality gate `if not victims.is_empty()` becomes "is this a legal cell / area to aim at" (allow empty-tile aims).
- `Classes/board/RulesService.gd` `gather_attack_victims` — stays as the unit-resolution step, but is no longer what *gates* the order.
- `Classes/squads/SquadManager.gd` `resolve_plan` — already derives the volley from the aimed cell each pass (#15); extend the resolved plan to carry cell / terrain effects, not only per-unit `AttackAction`s.
- `Classes/actions/PlanResolver.gd` / `ResolvedOutcome` — add a cell-effect channel (terrain ignition / tile state) alongside the existing per-unit damage / state.

## Changeover risk to watch (surfaced while testing #15)
Today you **can** queue a `ForwardLinePattern` attack that lands only on an ally, but you **cannot** with `ForwardWidePattern` — an odd divergence in what each pattern lets you aim at. It stops mattering once targeting is cell-based (you aim at cells regardless of occupants), but it's a latent inconsistency in the current unit-gated path and a likely source of a bug during the switch. Confirm both pattern families take the same aiming path when this lands.

## Not now
[...]

PR #537, opened 2026-08-26
#520: the camera goes to the action, and playback waits for it

Phaazoid:
Diff 1 of #520 (battle-zoom camera director), under umbrella #410. **The camera goes to the action, and the action waits for it** — in both profiles, per the ruling.

Diff 2 (the choreography: profile angles, the panning spin, tracking a volley, knockback and cliff follows, the lethality-aware dolly) is deliberately not here.

## What this is

**The enabling change is a rename, not a feature.** `CameraController.ai_locked` already meant *"something other than the player owns where the camera looks"* — it was only ever **named after its first caller**. It is `playback_locked` now, and `OrderExecutor.execute_orders` claims it too.

- **Saved and restored, never cleared.** An AI turn owns the camera for its whole length and a pass runs *inside* one, so a blind release would unlock the camera mid-turn.
- **`_board_locked_for_player()` is now composed of two named halves** — `playback_owns_board() or menu_is_up()`. Not a duplicate: the whole predicate is built *from* the halves, which finally have different readers.
- **The pan reuses the existing seam.** `CameraController.pan_to(unit, duration)` was already fixed-duration, headless-escaped, and awaited once per squad by `AIController`. `_execute_action_sequence` awaits the same call before a beat's opening action, so a volley pans once and then plays.
- **`BeatSheet.Beat.subject()`** answers who a beat frames: first victim, else the actor — because #47's swing at open ground has no victim and still has to be on screen.

**Why the camera never moved during your own Execute:** `battle3d._mirror_camera` returns early unless the flag is set, and the flag was set solely around AI turns. A `pan_to` moved the hidden 2D camera to no visible effect at all.

## The zoom, per the 2026-08-26 ruling

*"Let the player zoom in and out freely during AI turns and battle, but reset the camera zoom when they start."*

That reversed one of my four reader claims, and the cause was one line: `battle3d` set `_rig.manual_input_enabled` from the board lock, killing **orbit and wheel together** for a whole AI turn. Widening the flag as originally planned would have extended that to battle playback — backwards.

- The rig's wheel block moved **above** the manual gate, behind its own `zoom_input_enabled`, driven by the **menu half alone**. Zoom survives AI turns and playback; a menu still takes it.
- **Claiming the camera resets zoom** to `CameraRig3D.playback_distance` — hung off the edge-detect in `_mirror_camera` that already squared the view up for an enemy phase, so AI turns and battle passes both get it from one line. **Once on entry, never per frame**: a reset is a starting point, not a leash.

Knobs: **Camera travel to the action** (Playback) and **Playback zoom distance** (Camera handling).

## The one regression this caused, and how it was attributed

`tests/presentation/test_predicted_health.gd` → *readouts leave when the pass does* went red — **in its area, never alone**.

| run | result |
|---|---|
| the suite alone | 9/9 green |
| `presentation` on this branch | 1 failure |
| `presentation` on a clean `main` worktree | **386/386 green** |

Two separate facts, each needing its own run: it is **order-dependent**, and it is **mine**. The clean-main probe is what turned "probably mine" into "mine".

**Mechanism, confirmed by mutant rather than argued:** disabling the `pan_to` await took the reproducing pair (`test_input_bridge` + this suite, 53s) from 37-with-a-failure to 39 green. `test_input_bridge` synthesizes clicks, so it leaves the real mouse somewhere on the board; the camera now moves during a player's pass, so that stationary cursor points at a **different cell** when the pass ends — and one of them held a unit.

**The fix isolates the hover leg in that fixture, and that is narrower than it sounds.** The readout gate is `hovered or foretold or always_on`; every case in the file is about the **foretold** leg, and the file already says so (*"The pointer is nowhere — this readout is up because of the PLAN"*). Hover was absent only by accident, because nothing used to move the camera. It now sits beside the identical `PlayerSettings.reset_for_test()` isolation whose own comment calls it "NOT optional", for the same reason: without it these cases pass or fail on what the previous suite left behind.

(My first pass at this dismissed the hover hypothesis on the strength of that same comment — which belongs to a *different test* in the file. Re-reading it is what revived the right answer.)

## Verification

Areas: `presentation squad ui flow dev core` — **1446 cases / 132 suites, 0 failures, 0 orphans.** Full tree is CI's.

**Falsified — four mutants, each killed exactly the case it was aimed at:**

| mutant | case that reddened |
|---|---|
| zoom gate follows the board lock again (the pre-#520 bug) | `test_the_player_keeps_the_zoom_wheel_while_playback_owns_the_camera` |
| `subject()` frames the actor instead of the victim | `test_a_beat_frames_its_victim` |
| the board lock stops reading the flag (#484's exclusion) | `test_the_mirror_gate_and_the_pointer_gate_are_never_open_together` |
| blind release instead of save/restore | `test_a_pass_inside_an_ai_turn_leaves_the_camera_still_owned` |

The #484 exclusion was previously only prose in a comment plus a fixture line; it is a law over the flag now, which is what a rename most needed. And the save/restore turned out cheap to pin after all: **an empty queue still walks every phase**, so a pass can be driven for its ownership side-effect without staging combat.

**Embedded-content sweep:** not applicable. Nothing here adds a field to an embeddable content type — the rig's two new fields are `@export`s on a scene node, not on anything a scenario snapshot copies.

## One design question, for you rather than me

**After a pass, the camera stays parked on the last thing that happened.** That is what made the readout suite's old assumption false, and I think it is right: the camera followed the action and the action ended there.

But **#471 established a "camera return" idiom** elsewhere in this project, so returning the view to wherever you had put it is a coherent alternative. It is a feel call, so it is yours — and it is cheap either way, since the ownership edge that resets zoom is exactly where a return would hang.

## What is NOT verified

**The camera moving is not visible headless** — `pan_to` lands instantly there, deliberately, so nine suites that await `execute_orders` do not pay wall clock. The tests cover ownership, the gates, the subject choice and the schedule; **whether it reads well is yours, by playing.** Specifically worth judging: whether `Camera travel to the action` at 0.35s makes a pass feel deliberate or sluggish, and whether the playback zoom distance opens a fight at a sensible framing.

### DIFF
--- a/Classes/actions/AttackAction.gd
+++ b/Classes/actions/AttackAction.gd
@@ -175,6 +175,11 @@ func clear_preview_sprites():
 
 	preview_sprites.clear()
 
+# The victim. Null for a #47 cell attack, and the base's actor fallback is what covers that:
+# a swing at open ground is framed on the swinger.
+func aimed_at() -> Unit:
+	return target
+
 func get_target_name() -> String:
 	if target != null and is_instance_valid(target) and not target.is_queued_for_deletion():
 		return target.get_unit_name()
--- a/Classes/actions/BaseAction.gd
+++ b/Classes/actions/BaseAction.gd
@@ -76,6 +76,17 @@ func is_main_action() -> bool:
 func actor_can_perform() -> bool:
 	return true
 
+# Who this order is AIMED AT -- the unit it is done TO, rather than the one doing it. The default
+# is the actor, which is the honest answer for a verb that acts on itself (move, rally, reload);
+# Attack/Rescue/Intimidate/Guard override it with the `target` they each already store.
+#
+# Declared per Law #4: those four have held a private `var target: Unit` each, with no shared door,
+# since they were written -- this is the door, not a fifth copy. It exists because BeatSheet has to
+# ask the question of an order whose class it does not know (#520): a beat frames what is being
+# done to whom, and only the order can say who that is.
+func aimed_at() -> Unit:
+	return actor
+
 func get_actor_texture() -> Texture2D:
 	if actor == null or not is_instance_valid(actor):
 		return null
--- a/Classes/actions/GuardAction.gd
+++ b/Classes/actions/GuardAction.gd
@@ -30,6 +30,9 @@ func init(guarding_unit: Unit, warded_unit: Unit) -> void:
 	guard_range = guarding_unit.get_guard_range()
 	action_type = BaseAction.ActionType.GUARD
 
+func aimed_at() -> Unit:
+	return target   # the ward -- what a Guard is about is who it covers
+
 func execute() -> void:
 	begin_execution()
 	if target != null and is_instance_valid(target) and not target.is_queued_for_deletion():
--- a/Classes/actions/IntimidateAction.gd
+++ b/Classes/actions/IntimidateAction.gd
@@ -17,6 +17,9 @@ func init(intimidator: Unit, victim: Unit) -> void:
 	target = victim
 	action_type = BaseAction.ActionType.INTIMIDATE
 
+func aimed_at() -> Unit:
+	return target   # the one losing Will
+
 func execute() -> void:
 	begin_execution()
 	if target != null and is_instance_valid(target):
--- a/Classes/actions/OrderExecutor.gd
+++ b/Classes/actions/OrderExecutor.gd
@@ -87,16 +87,25 @@ func execute_orders(unit):
 	var is_ai: bool = game.ai_controller.is_ai_faction(squad.leader.get_faction())
 	var beat: float = Pacing.base_for(profile, is_ai)
 
+	# Playback owns where the camera looks for the rest of this pass (#520). SAVED and RESTORED, never
+	# cleared: an AI turn claims it for the whole turn and this runs inside one, so a blind release
+	# would unlock the camera mid-turn.
+	var camera_was_locked: bool = game.camera_controller.playback_locked
+	game.camera_controller.set_playback_locked(true)
+
 	await _execute_action_phase_parallel(move_actions)
-	await _execute_action_sequence(plan.attacks, beat, _beat_holds(sheet.volleys(false), profile, is_ai))
+	await _execute_action_sequence(plan.attacks, beat, _beat_holds(sheet.volleys(false), profile, is_ai),
+			_beat_subjects(sheet.volleys(false)))
 	_apply_cell_effects(plan.cell_effects)
 	# The act break, held once between the two montages rather than folded into the first counter --
 	# a turnover the counters then pace on top of, not instead of.
 	await Pacing.beat(self, Pacing.duration_for(sheet.turnover(), profile, is_ai) if sheet.turnover() != null else 0.0)
-	await _execute_action_sequence(plan.counters, beat, _beat_holds(sheet.volleys(true), profile, is_ai))
+	await _execute_action_sequence(plan.counters, beat, _beat_holds(sheet.volleys(true), profile, is_ai),
+			_beat_subjects(sheet.volleys(true)))
 	for type in BaseAction.SIDE_CHANNEL_ORDER:
 		var batch: Array = side_channel.get(type, [])
 		await _execute_action_sequence(batch, beat)
+	game.camera_controller.set_playback_locked(camera_was_locked)
 	# The last await has returned, so the pass is played out: released HERE rather than beside
 	# _end_squad_turn because everything below is synchronous (no frame renders between them) and
 	# the squad-validity bail below skips that call entirely.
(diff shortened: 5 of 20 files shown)


Phaazoid:
🤖 Claude says:

Follow-up commits on your play-check. Two commits: `8aa9159` (the change) and `ad59318` (one test I only found by falsifying).

## Your three notes

**1. The move — confirmed, and it is my regression.** `execute_orders` called the parallel move phase with no pan at all; the `BeatSheet` has built a MOVES beat since #524 and nothing consumed it. Worse, diff 1 made playback own the camera for your *own* pass too — so before it the move was unframed but you could still scroll, and after it the camera was unframed *and* frozen. It now pans to the walk and waits, and `pan_to`'s closing `follow()` carries the camera along beside the walker for free.

The walk opens on the **leader** when the leader is walking, and **hold-position fillers do not count as movers** — almost every real squad move carries a filler per non-moving member, and framing one parks the camera on someone standing still. An all-holds queue earns no beat at all.

**2. The heal — I traced it, and the camera was already going there.** `resolve_plan` expands each reaction through `create_counter_volley`, so a reactive heal is an ordinary counter volley, and the beat's subject is the healed unit. The shipped heal (`heal2.tres`) is a `ManhattanRangePattern` that does not override `get_affected_cells`, so it is single-target — there is no bystander to mis-frame.

What was missing is **emphasis**: `Pacing.hold_for` had clauses for down/kill/maim, Crisis, Iron Will and knockback, and **none for a heal**. `Battle zoom` defaults on, so you were in the cinematic profile where a death holds `0.4 + 0.6 = 1.0s` and a heal held `0.4s` flat — the quietest thing in the pass, arriving last. There is a `Hold: a heal` knob now.

So the fix is a timing one, not a framing one. If you watch it again and the camera genuinely does not go to the healed unit, my trace is wrong somewhere and I want to know.

**3. End Turn button — not a regression, it is #467's own ruling.** `refresh_end_turn_button()` sets only URGENT; nothing sets `visible`. #467 dropped the End Turn row from the action ring, which made this button the sole door to ending a turn, and permanence was the consequence. Spun out as #541 with the three options and the note that the hide-predicate (`_board_locked_for_player()`) is already computed.

## Your fourth note, and what it changed

*"The side channel actions are going to need emphasis as well."* — that turned a two-spot patch into a model fix, because the sheet and execution disagreed about what a coda IS.

`execute_orders` plays the tail **one order at a time, each awaiting its own completion**. `BeatSheet` built **one beat per TYPE**. So three rescues are three moments and the sheet said one — rescuers two and three would have played off-camera even after wiring. One beat per order now, and each side-channel verb has its own hold (a rescue is not a reload).

Two structural bits this needed:

- **`BaseAction.aimed_at()`** — Attack/Rescue/Intimidate/Guard each store a private `var target: Unit` with no shared door, so a beat could not ask "who is this done to" of an order whose class it did not know. The virtual returns theirs; the default is the actor, which is the honest answer for a self-verb. I did not hoist `target` itself — GDScript forbids a subclass redeclaring a parent member, so that means deleting the field from `AttackAction` and rewriting every reader, which does not belong in a feel fix.
- **`Beat.actions` widened to `Array[BaseAction]`**, so one field answers for every kind and `subject()` asks the orders instead of walking `victims`. Same answers for volleys, and it answers for moves and codas for free.

## The finding I would flag even if you skim the rest

**A mutant deleting the side-channel camera wire survived 587 cases** across `presentation` and `squad`. Both ends were freshly, deliberately pinned — the sheet builds the beats, the schedule maps them to subjects — and the argument *between* them was pinned by nothing. That is #103's shape, in brand-new code, written the same hour I re-read the law about it. `ad59318` closes it with a case that parks the camera on another unit first, so "it ended up on the right one" cannot pass by the camera never having moved.

Second, smaller: **the nine new knob rows shipped born dead.** A `CLASS_KNOBS` row needs a hand-written arm in `read_static`/`write_static`, and without it every scene instantiation raised `unknown static` — while the suite reported 21/21, because a `push_error` does not red a case. I caught it in the console. I then checked by mutant that `test_every_class_knob_resolves` *does* catch it, so CI was a real backstop, but only on a full-tree run.

## Verification

- `squad core law dev` — **755 cases, 0 failures, 0 orphans**
- `presentation/test_camera_follow.gd` — 22/22; `presentation squad` together ran 587 green during the mutant pass
- **Six mutants.** Five killed exactly their target case (move pan / `has_heal` / fillers-as-movers / leader-first / coda-per-type). The sixth is the survivor above, now killed by the new case.
- The headless Play API twin needs no change — this adds camera work and pause scheduling, not a phase.
- No embedded content touched: no `.tres` embeds a `BeatSheet` or a `Pacing` value.

## What only you can judge

Whether following the **leader** is the right read when five units move at once, and whether the nine new hold values give each moment the room it wants. All knobs, Game tab → Playback: `Hold: a heal`, `Hold: a rescue`, `Hold: a capture`, `Hold: a rally`, `Hold: an intimidate`, `Hold: a guard arming`, `Hold: a burrow`, `Hold: a reload`, `Hold: a rev`.

Note they only bite where `Drama: zoom on` is above zero — with the zoom **off**, `Drama: zoom off` still ships at 0, so the plain board stays flat by your earlier ruling. If the tail feels flat with the zoom off, that dial is the one to raise rather than the individual holds.

Your tree is on `feat/498-top-bar-plate`, so I have not touched it — `gh pr checkout 537` when you want this.

PR #550, opened 2026-08-26
Battle zoom: the post-turn effect pass, and the camera comes back (#534)

### LINKED ISSUE #534: Battle zoom: a quick pass for post-turn effects
Spun out of #520 (battle zoom camera director) from the dev's own note, 2026-08-26.

## Ask

End-of-turn effects currently resolve **invisibly**. Give them their own quick pass:

> *"We'll probably want a quick post turn effect zoom to show all post turn effects that happen - currently only units standing in fire. I'm thinking a double speed camera zoom to an unit, unit takes fire damage, next, etc."*

So: camera to the affected unit at **double speed**, the damage lands, next unit, done. Quick and rhythmic — this is bookkeeping made legible, not drama.

## What it is today

`OrderExecutor.apply_burning_tile_damage(faction)` is a **synchronous loop** — it walks `terrain_states.burning_cells()`, damages any unit of the acting faction standing on one, then drains `_process_downed_pending()`. No beats, no camera, nothing to watch. `game.end_turn()` awaits it, but there is nothing to await.

So a unit can take fire damage, go down, and be ejected from its squad with **no frame between any of it**.

## Shape

- The affected set is knowable **before** anything plays — it is a pure query over burning cells and the acting faction — so this phase gets a beat list the same way a resolution pass does. Whether that is `BeatSheet`'s second reader or something smaller is a build-time call; what it must not be is a second answer to *"who is about to be hit and what happens to them"*.
- **Reuses #520's camera machinery** — the "move the camera and wait for it" seam, at a shorter duration. It must not grow its own.
- "Currently only units standing in fire" is the point: the phase should be able to hold whatever end-of-turn effects arrive later, not just burn.

## Depends on

**#520** — the camera director. There is no point building this against a camera that cannot go anywhere.

## Notes

- `Pacing`'s headless escape must carry through, as everywhere else: this phase is awaited from `end_turn`, so a literal timer here puts wall clock on every test that ends a turn.
- Its beat lengths and the "double speed" factor are feel values, so they arrive as knobs in the Game tab's Playback group like the rest of the beat table (#519).
- A burn that takes the last unit ends the mission — `game.end_turn` already calls `mission_controller.check()` immediately after, so the pacing must not let the banner land before the damage that caused it is on screen.

Phaazoid:
Closes #534.

Two halves of one act, because the burn phase ends by *releasing* the camera and the return is what that release should do.

## The camera comes back

Your ruling, 2026-08-26: *"the camera should return home after a pass, and after the enemy turn."*

**"Home" was already taken.** `CameraRig3D` has `_home_position`/`_home_yaw_degrees`/`_home_distance`, written by `frame()`/`pose()` and restored by the **R key** — meaning *the opening shot*: the board's fit, or the mission's authored `camera_start`. That is a different question from *where the player was standing*, so this gets its own name (`_borrowed_*`, with `stash_view`/`restore_view`/`drop_stashed_view`) and the two are declared side by side rather than merged.

**One edge serves all three of your cases.** `battle3d._mirror_camera` already edge-detected playback ownership and did the detent/zoom reset on the way *in*; the way *out* was simply missing. A pass claims and releases inside `execute_orders`, an enemy turn holds it across the whole turn, and the new post-turn pass claims it again after that — all through the same flag, so there is no per-caller wiring.

**Invalidation is structural.** `frame()` and `pose()` drop what is borrowed, because anything that redefines the opening shot is a new board — and `ScenarioManager.clear_board` releases the playback lock, which would otherwise fire the restore on a board that no longer exists.

The restore matches R exactly: position assigned (the mirror writes it directly every frame, so there is no target to ease toward), yaw and distance set as targets so `_process` eases them, distance through `set_zoom`'s clamp.

## The post-turn pass (#534)

`apply_burning_tile_damage` was a synchronous loop — a unit could burn, go down and be ejected from its squad with **no frame between any of it**. It now pans to each burning unit at double speed, holds, then lands the damage.

- **One list, walked once**, so there is structurally no second answer to *who is about to be hit*. Deliberately **not** a `BeatSheet`: that consumes a `Squad` and a `ResolvedPlan`, and this phase has neither, so using it would mean faking a plan.
- **A phase with nothing to show claims nothing** — load-bearing rather than an optimisation, since the release is what fires the view return. Claiming on an empty phase would return the camera at the end of every single turn.
- `Pacing.POST_TURN_SCALE` scales the pan *and* the beat together, so "double speed" is a relationship to the values you just tuned rather than a second set of numbers. Game tab → Playback → **Post-turn pass speed**.

## The thing to look at in play

**The return puts back the aim, yaw *and zoom* from before playback borrowed the camera — so a zoom you make during an enemy turn is discarded when the turn ends.** Zoom is the one control you keep during playback (your #520 ruling), so that is the one place "back where you left it" fights you. Leaving zoom alone is a one-line change; I went with the symmetric version because that is what the plan said, but it is worth a judgement in play.

Found by an existing test going red rather than by reasoning: `test_an_ai_turn_squares_the_camera_up` drove a yaw *mid-turn* to prove the realign is an edge and not a per-frame clamp, then released and re-claimed expecting the nearest detent to that yaw. With the return, the release puts the pre-claim yaw back first. Its claim survives — the re-drive moved to *after* the release, which is the honest scenario anyway, since manual input is off during playback and a mid-playback yaw is not something a player can author.

## Verification

- `terrain core dev presentation` — **884 cases, 0 failures, 0 orphans**; no `unknown static` in the console, so the new knob row actually resolves.
- `ui flow` — run separately because `game.end_turn` is in the path.
- **Five mutants, five precise kills:** drop the stash; drop the restore; drop the `drop_stashed_view` from `frame`/`pose`; drop the burn pan; claim the camera on an empty phase.
- The three existing burn cases now `await` the call — they had survived purely on the headless escape, which is too quiet to rely on.
- No embedded content touched; the headless Play API has no burn phase at all, so that stays a documented gap rather than one this widens.
- **Tree-dirty gate** (your rule landed on main mid-ticket, 2026-08-26): this writes no tracked file at any point — no `.tres`, no generated artifact, no settings — so nothing it adds can re-dirty the tree afterwards. Asked at both ends rather than only at commit time.

## Two red herrings, recorded so nobody re-pays them

- **The 2D camera is irrelevant here.** `_update_pointer` snaps it to the hovered cell on every mouse move, to park the hover card through the canvas transform — so it is derived, not a view anyone owns. The return is purely a 3D rig question.
- **An empty pass never fires the ownership edge.** `execute_orders` on an empty queue claims and releases within one call with no frame between (every phase returns immediately and `pan_to`/`Pacing.beat` no-op headless), so `_process` never observes the flag change. Testing anything in `_mirror_camera` needs something that spans frames — a real queued move.


Phaazoid:
## Second pass — your play-check, 2026-08-26

Both notes landed, and tracing the second one turned up a third cause you could not have seen.

### The linger was empty by construction

The loop ran `pan → hold → damage`. The pause was spent on a unit at **full health**, and the camera left for the next one as the cubes burst — exactly what you described. The hit now lands **before** the hold.

That also repairs something #534 asked for and my first cut had backwards: *"the pacing must not let the banner land before the damage that caused it is on screen."* With `damage` last, the final unit's hit landed and the phase returned with no hold at all, straight into `mission_controller.check()`.

### One dial over two questions

`POST_TURN_SCALE` scaled the camera travel and the pause **together** — which is why there was no separate speed and linger to find. And "Post-turn pass speed" is not a phrase anyone would search for; you were looking for fire.

It splits into `ENVIRONMENT_PAN` (0.25) and `ENVIRONMENT_HOLD` (0.45), Game tab → Playback:

- **Camera travel to a burning unit**
- **Hold: a unit burns**

Roughly double the old total, with the weight in the hold since that is now the part carrying the point. #534's own text asked for the plural — *"its beat **lengths** and the double-speed factor"* — and I collapsed them; your note is that reversed.

Neither forks on the battle-zoom profile or on whose turn is ending. The **board** acts in this phase, not a faction, so an enemy's fire and yours are the same event and there is no drama for the zoom to add to bookkeeping. That is a change: the hold used to be longer on an AI turn and longer again with battle zoom on.

`test_the_post_turn_pass_runs_quicker_than_a_blast` is **deleted** rather than rewritten. It pinned "quicker than a blast" as a relationship, and with two free dials that is a feel value you are allowed to disagree with.

### The third one: for most players the phase showed nothing at all

The readout gate is `hovered or foretold or always_on`. During this phase the cursor is not on the unit and there is no plan — so with **"Always show health bars" off, which is the default**, the bar is down. `UnitMirror._settle_health_change` early-returns on a hidden bar, so **not even the cube burst fired**.

You could not have seen it: your `settings.cfg` has `ALWAYS_SHOW_HEALTH=true`.

The phase now publishes the units it is about to hit, and the gate gains that as a **fourth disjunct in the same one expression** — not a second visibility rule, which `UnitMirror` says outright is the bug. It is the *same* reason `foretold` already encodes ("something is about to happen to this unit"), reached from the other direction: this phase has no `ResolvedPlan` to be read out of and must not fake one, so it publishes its own subject list the way `executing_plan` is already published.

Raised for the **whole** phase, not one unit at a time, matching how a plan raises a bar over everyone it will touch — so the units still to come are readable when the camera arrives.

### Verification

- `terrain core dev presentation` and `ui flow` — **1350 cases, 0 failures, 0 orphans**, and no `unknown static` in the console, so both new knob rows resolve.
- **Three new cases.** The pass names every unit it will hit *before it reaches any of them* — probed from inside the running phase through a burn death, the one event that fires mid-pass, and written so it cannot pass vacuously if that death stops happening. It stops naming anyone once it ends. And the readout goes up over exactly the marked unit with the preference **off**, written through the real wire — executor property, `battle3d`'s callable, the mirror's per-frame poll — with only the phase stubbed, because a phase run headless finishes inside one call and there is no frame in the middle of it to look at.
- **Six mutants. Four precise kills, and two that survived.**

| mutant | result |
|---|---|
| put the damage back after the hold | **survived** — see below |
| publish the subject list lazily, one unit at a time | killed: *a unit the phase had not reached yet was left wearing no readout* |
| never clear the list | killed: *the phase left its readouts up after it had finished* |
| drop `marked` from the gate | killed by the mirror case |
| cut `battle3d`'s wire, leaving the gate correct | killed by the same case — so the **wire** is pinned, not just the two ends |
| delete `write_static`'s `ENVIRONMENT_HOLD` arm | **survived** — became the second commit |

**The ordering is not observable headless, and I would rather say so than imply it is tested.** Both `pan_to` and `Pacing.beat` return without suspending under `--headless`, so the hold has no duration to be on either side of and the two orderings leave identical state. Mutant 1 was run to *prove* that rather than assert it. It is verified by reading, by your play, and by a comment carrying your words.

**The second survivor is the interesting one.** Deleting the `ENVIRONMENT_HOLD` write arm left the whole suite green — the row still *reads*, so `test_every_class_knob_resolves` is happy, and the panel draws a slider that moves nothing. That is the third born-dead-knob mode and the one you actually meet. The layer half of `CLASS_KNOBS` was checked for taking a write; the static half was not, and the comment that skips it claimed otherwise — *"the statics are asserted separately: the 2D owns them."* They were not asserted anywhere. `test_every_static_knob_takes_the_write_its_slider_makes` closes it, and every existing static row passes it, so nothing else was quietly dead.

### The thing to look at in play

The hold is now the dial that decides whether the pass is legible, and **0.45s is a first guess**. The two knobs are the answer to that, not the numbers I picked.

Also worth a look: the hold no longer lengthens on an AI turn or with the battle zoom on. If a burn on the enemy's turn now reads too brisk next to the rest of their pass, that fork is one line back.

Issue #203, opened 2026-08-13
HD-2D Stage 0: look-dev scene (the GO/NO-GO gate)


Phaazoid:
🤖 Claude says:

## What this is

Stage 0 of #176 — the **look-dev scene**: a standalone HD-2D diorama (GridMap board with placeholder pixel textures, billboarded sprites, sun + one torch, the full WorldEnvironment/DoF post stack) built to judge whether the style lands for Iosis **before anything else is committed**. It is the GO/NO-GO gate from #176's build order, and it is deliberately not throwaway — it stays as the permanent lighting/grading playground.

## Scope

- `Scenes/LookDev/` — a self-contained scene wired into nothing; run it directly (it never touches `Main.tscn`, `game.gd`, or any game system).
- Generated placeholder tile art (**32 px/tile, flat-lit**) and a MeshLibrary (grass block / stone block / dirt ramp) via `tools/lookdev/gen_lookdev_assets.gd` — regenerate-able, tweakable, no commissioned art.
- Units are the existing `Art/Units/MapSprites/` pixel sprites mounted on `Sprite3D`s (Y-billboard, nearest filter, opaque pre-pass alpha cut, shaded so they receive light and cast shadows). One pixel density everywhere: **32 texels per world unit**, tiles and sprites alike.
- Camera rig honoring the camera contract: perspective, narrow FOV, pitched down, **90°-snap orbit** (Q/E), wheel zoom, WASD pan.
- The post stack: glow with a high HDR threshold, volumetric fog, filmic tonemap + color adjustments, near/far depth of field (`CameraAttributesPractical`), vignette shader.
- Lighting presets on keys 1–4 (day / sunset / night / overcast) to demo the relighting trick, plus per-ingredient toggles (fog / glow / DoF / vignette) so each layer's contribution can be judged in isolation.
- **Renderer switch `mobile` → `forward_plus`** — volumetric fog is Forward+-only. Global, one line in `project.godot`, approved by the dev 2026-08-12.
- Headless coverage under `tests/presentation/`: scene instantiates clean, structural invariants pinned (billboard mode, alpha-cut, nearest filter, shadows enabled, glow/fog/DoF present, board has ≥2 height levels). **No feel-tuned values pinned** — light energy, fog density, DoF distances belong to the tuning session.

## Out of scope

Stages 1–5 (picking seam, units + movement, overlays, UI anchoring, art pass) — filed only if the gate passes, per #176's build order.

## The gate

[...]

PR #563, opened 2026-08-26
Block seams: a side face stops short of the surface, and back faces are culled again

### LINKED ISSUE #559: Hairline dirt seams between ground blocks at axis-aligned camera yaw
*Filed by Claude from a Discord bug report, not written by the dev directly.*

Source: `user://reports/2026-08-26_15-57-23/` in `#feedback-reports` — build **0.96.0**, checkout `feat/534-post-turn-pass @ 0b8e479`, state `AI_TURN`, view **HD_2D, yaw 0 deg, zoom 11.0, centred on (-12, 0)**. The dev's words: *"sometimes, at specific camera angles and zooms, I can see lines going in between our 3d blocks."*

## Symptom

A dark brown hairline runs horizontally across the whole board. In the report's `board.png` it sits at **y = 491** of 720 (about two thirds down), spans **x = 0 to 1279**, and is **exactly one pixel tall**.

It is not confined to one line. In a second axis-aligned report (`2026-08-25_16-54-22`, yaw 0) there are **eight** of them stacked in a regular series — y = 290, 457, 558, 565, 570, 575, 577, 584 — i.e. one per cell row, spacing shrinking with distance exactly as perspective would compress it.

## What it is, measured off the screenshots

- **Constant colour, independent of what is behind it.** The line reads ~`RGB(111, 87, 67)` over grass `(58,226,169)`, over the orange path `(233,145,100)`, and where it crosses the action panel. It is not a modulation of the underlying pixel.
- That colour is **`_dirt_side`** — `gen_lookdev_assets.gd:161` builds the block side texture from base `Color8(118, 86, 58)`, with rows 0 and 1 deliberately darker (`Color8(94,66,44)` / `Color8(102,73,49)`, commented *"darker soil lip under the grass line"*). The measured line sits inside that family and is nowhere near grass or path.
- **The scene is continuous across it.** On the left of the frame a diagonal grass/path edge passes straight through the line without a step (boundary x = 163, 162, 161, 160 at y = 488, 490, 492, 495). So it is drawn *over* a coherent render — not a seam between two renders, not a viewport join.
- **It is a huge outlier, not texture noise.** Scanning every row, the line's darkness delta is 62.6; the next largest anywhere on the board is 14.3, which is the grass fleck pattern.

This rules out shadow-cascade banding, z-fighting between terrain and markup, AO, and depth-precision haze — all of those modulate the colour underneath rather than replacing it with a constant.

## It only happens at axis-aligned yaw

Six reports carrying a camera pose were checked with the same detector (a dirt-coloured hairline sandwiched between *identical* terrain above and below, which excludes genuine cliff faces):

| report | yaw | hairlines found |
|---|---|---|
| 2026-08-26_15-57-23 | **0** | yes — 235/270 sampled columns at y=491 |
| 2026-08-25_16-54-22 | **0** | yes — eight rows |
| 2026-08-23_11-56-54 | **-0** (zoom 33.8) | none resolved at that distance |
| 2026-08-25_23-18-29 | 9 | none |
| 2026-08-23_19-25-57 | 30 | none |
| 2026-08-23_22-59-09 | -50 | none |

Off-axis, the same seams still exist but stagger across many scanlines and read as ordinary pixel noise. At yaw 0/90/180/270 an entire row of cell edges lands on **one** scanline and the noise becomes a clean screen-wide line.

**This is the pose the game snaps to on purpose.** `CameraRig3D.align_to_detent()` rounds yaw to the nearest `yaw_step` (90 deg) and its caller is the enemy phase — *"an enemy phase reads square-on however the player left the camera."* Q/E realign and R reset land on detents too. Today's report was captured during `AI_TURN` at yaw 0: the game drove the camera into the triggering pose itself.

## Likely mechanism

Every ground cell is its own GridMap item, and `_block_mesh` (`gen_lookdev_assets.gd:718`) emits **all six faces** of every block — top, four sides, bottom — with `size = (1.0, ROW_HEIGHT, 1.0)` and no knowledge of neighbours. `BoardMirror._write_column` then fills each column solid from `floor_row` to `top_row`, so the board is a mass of boxes whose interior faces are all still there. `GridMap.cell_size = (1, 0.5, 1)` matches the mesh exactly, which means:

- neighbouring cells' **top faces abut exactly** at the cell border, as two separate coplanar quads from two separate mesh instances, and
- each block's **side face reaches exactly that same top plane** (the side quad spans `up` down to `down`, and `up` is the top face's own plane).

So at every cell border the buried dirt side face is coincident with the edge where two grass top faces meet. A pixel whose centre falls on that shared edge can be won by the dirt face instead of the grass — either as a rasterisation crack between the two abutting quads (their shared vertices are computed through different instance transforms, so they are not guaranteed bit-identical) or as a depth tie right at the edge. Either way the result is one pixel of `_dirt_side` where grass belongs.

Nothing smooths it: **`rendering/anti_aliasing/quality/msaa_3d` is not set in `project.godot`**, so 3D MSAA is disabled and the crack rasterises hard.

Same bug class already bit once here and was fixed by removing the coplanar duplicate — `ef910fd` / #496, *"A corner cap stops drawing ground on its own floor ... That is what stopped an outer corner's flat half being drawn twice into one plane."*

## Candidate fixes (not yet costed — needs a plan before anyone builds)

1. **Stop the side faces reaching the top plane.** Drop the side quads' top edge by a hair so they can never win a pixel at the surface. Cheapest, but needs checking against real cliff faces, where the side is legitimately visible and a gap would show.
2. **Overlap the top faces.** Make the top quad a hair larger than one cell so neighbours overlap instead of abutting. Coplanar overlap of the same material is harmless; costs a small UV stretch.
3. **Stop emitting buried geometry at all.** `_write_column` fills columns solid; only the visible skirt needs sides. The largest change, and the only one that also buys back the wasted geometry.
4. **Turn on MSAA.** Would blend the crack rather than remove it, and has to be weighed against the deliberate pixel-art crispness (`default_texture_filter=0`) and perf. A mitigation, not a fix.

Options 1 and 2 are one-line changes in the generator; both need a regenerated meshlib in the same diff.

## Notes

- The line crosses different tile types (grass and path) on the same row, so it is not a boundary between two terrain kinds.
- The `board.png` screenshot is being sent to the dev directly rather than linked here — Discord CDN attachment URLs carry a signed expiry and would rot in a permanent issue. The original report in `#feedback-reports` stays the source of record.
- Priority is a guess; downgrade freely if this reads as polish rather than demo-blocking. The argument for P1 is that it appears at the camera pose the game itself snaps to every enemy phase, and the Demo milestone's whole job is looking finished to a stranger.

Phaazoid:
Closes #559.

## What was wrong

`_block_mesh` emits all six faces of every ground block with no knowledge of its neighbours, and `_write_column` fills columns solid — so a side quad ran from `up`, the top face's own plane. Neighbouring cells' top faces abut there exactly, which put **four surfaces on one line at every cell border**: both top faces and both blocks' buried sides. A pixel centre landing on that line could be won by a side.

Off-axis those wins stagger across scanlines and read as texture. At an axis-aligned yaw a whole row of borders projects to **one** screen row and the scatter becomes a brown hairline across the board. `CameraRig3D.align_to_detent` snaps yaw to 90 on the enemy phase, so the game reaches the triggering pose by itself — which is the state the report was captured in.

## The fix

Sides now start `BoardSpace.SIDE_RIM` below the surface, and a **rim wearing the TOP material** closes the gap.

The rim is not decoration. Dropping the sides alone would leave a `SIDE_RIM`-tall band of nothing at every real cliff crest, and with back faces culled that reads as a hole through the board. Because the rim wears the ground's own art, the surface still meeting the neighbour at a border is ground rather than the dirt under it — so a residual tie there is invisible instead of brown.

A prop passes `rim` 0: it stands alone on a cell and has no border to tie at.

**The top face keeps its full extent**, which is load-bearing twice — `BoardSpace.CELL_SIZE` has to go on agreeing with the authored `GridMap.cell_size`, and `_form_mesh`'s open shell is safe only while the block below covers the whole footprint.

`SIDE_RIM` lives in `BoardSpace` rather than the generator: "where a side face stops" is a fact about the coordinate convention, and `BoardSpace` already declares itself the one place that arithmetic lives.

## Cull mode

`_mat()` goes back to `CULL_BACK`. It was `DISABLED` from the Stage 0 bootstrap (`6942202`, #203 v0.8.0) — set before the thin PLANE slabs and the open-shell cap existed, so it was never chosen for them.

It is **not** what fixes the seam (the camera-facing buried side is front-facing and still drawn), but it stops shading every buried face in a solid board and halves what competes at each border.

What it costs is that an open shell now reads as a hole rather than as its own inside surface. `_form_mesh` ships one on purpose, and the claim keeping it safe — *"its opening is exactly the footprint the block's top face covers"* — was written while back faces were being drawn anyway. That went from comfortable to load-bearing, so it gets a law instead of a comment.

## Laws

All three measure properties, never `SIDE_RIM`'s value, so it stays free to move.

| law | claim |
|---|---|
| `test_a_side_face_never_reaches_the_surface` | a ground block's side material stops short of its top plane |
| `test_a_ground_block_is_a_closed_shell` | every edge is shared by exactly two triangles — no holes under `CULL_BACK` |
| `test_a_cap_opening_lies_on_the_block_top` | a cap's boundary edges all lie in the block-top plane and inside the cell, **and the shell is still open** |

That last "and" matters: without it the case passes vacuously the moment someone puts the bottom quad back, which is precisely the #496 regression it exists to catch.

## Verification

- `tests/law` **157/157**, `tests/presentation` **397/397**, 0 orphans, exit 0. Area folders only — CI owns the tree.
- **Falsified, committed first.** `SIDE_RIM = 0.0` reds the side-face law on all 313 ground blocks; deleting one rim quad reds the watertight law on all 313; re-adding the cap's bottom quad reds the cap law with its own message. Each revert verified with an empty `git status`.
- **Not verified: the artefact is gone on screen.** The geometric contract is proven and pinned, but confirming the *symptom* needs a render, and a headless run never draws (`frame_post_draw` never fires). Reproducing the report's pose — HD_2D, yaw 0, zoom 11.0, centred on (-12, 0) — and looking is the one check I could not do. Flagging rather than dropping it.

Content sweep: nothing embeds block geometry — the meshlib is regenerated, not hand-authored, and no `.tres` fixture carries a copy of it.

Canon swept (`verticality.md`, `presentation-effects.md`, `CLAUDE.md`), markers bumped to #559.


Phaazoid:
🤖 Claude says:

**CI is red, and it is red on `main` too — same two cases, neither of them mine.**

| | main @ `1425b40` (run [33014130432](https://github.com/Phaazoid/Godoiosis/actions/runs/33014130432)) | this PR (run [33016705304](https://github.com/Phaazoid/Godoiosis/actions/runs/33016705304)) |
|---|---|---|
| cases | 2425 | 2428 |
| failures | 2 | 2 |
| orphans | 0 | 0 |

`1425b40` is this branch's exact base, and the two failing cases are identical. The case count is +3, which is precisely the three laws this PR adds — so all three pass in CI, and nothing here moved either failure.

The two, for the record:

1. **`test_board_lint.gd > test_every_shipped_mission_is_playable`** — `Level_2.tres: 6 ENEMY units on this board, but ENEMY is not computer-controlled`. Authored content, not code: the board never got ENEMY ticked, so it hands the player its own enemies. This is the exact fault `BoardLint._check_ai_factions` exists to catch, doing its job.
2. **`test_dev_tree.gd > test_space_spawns_only_from_the_spawn_page`** — *"SPACE did nothing on the page that owns it — the gate is refusing everything"*, i.e. `spawn_armed()` is refusing on its own page.

Both are out of this ticket's scope and I have not touched them. Flagging rather than folding them in, since one is a content decision and the other is a dev-tools gate that wants its own diff.

**What I could not verify here:** that the hairline is actually gone on screen. A headless run never draws, so the geometry is proven and pinned while the symptom is not. Reproducing the report's pose — HD_2D, yaw 0, zoom 11.0, centred on (-12, 0) — is the check that needs you.

Phaazoid:
🤖 Claude says:

**Green.** Superseding the comment above: `main` was merged in at `5e3f2f1`, which picks up #564's two fixes, and CI run [33023717792](https://github.com/Phaazoid/Godoiosis/actions/runs/33023717792) is **2444/2444, 0 failures, 0 errors, 0 orphans, exit 0**. Both cases named above pass now; the three laws this PR adds are in that count and green.

The merge was clean and checked rather than assumed — we both edited `CLAUDE.md`, so the risk was the adjacent-line one. The PR diff against main now reads `CLAUDE.md | 1 +`, i.e. exactly this ticket's one bullet with #564's canon sweep absorbed from main, and the scope is still only this ticket's 12 files. Nothing on main touched the generator, the tileset or `BoardSpace`, so the committed meshlib is still byte-for-byte what a regeneration produces.

**Still outstanding, and it needs you rather than CI:** whether the hairline is actually gone on screen. A headless run never draws, so the geometry is proven and pinned while the symptom is not. The pose that produced it is HD_2D, yaw 0, zoom 11.0, centred on (-12, 0).

Issue #552, opened 2026-08-26
Water is a solid opaque cube with a static texture: the Tier-1 water pass (animation, shoreline, reflection)


Phaazoid:
🤖 **Claude says:**

*(Filed from the dev's direction, 2026-08-26, after auditing Tier 1 of `docs/design/presentation-effects.md` against the code. Sibling ticket: ambient particles.)*

## The ask

Tier 1: *"animated pixel water with real reflections and foam edges (Octopath 2's showpiece)."* The doc calls it the bar for the tier. What ships is a solid cube.

## What ships today, measured

`Terrain.Kind.WATER` maps to MeshLibrary item 5 (`BoardMirror.KIND_TO_ITEM`), and `tools/lookdev/gen_lookdev_assets.gd:342` builds it:

```gdscript
_add_item(ml, 5, "water_block", _block_mesh(_mat(water_top), _mat(water_top)))
```

Three facts follow, and each is a separate thing to fix:

1. **It is a block, not a surface** — and the *same* material is on the top face and the sides. Every other kind pairs a top with a distinct side (`grass_top`/`dirt_side`, `stone_top`/`stone_side`). Water is the only one whose wall wears its own surface, which is what a cube of water rather than a body of water looks like.
2. **The texture is static and generated once.** `_water_top()` writes a 32×32 flat blue (`Color8(52, 96, 150)`), lightens every 8th row, and scatters **14 two-pixel "wave flecks"** at random. That is the entire animation budget: none.
3. **The material cannot catch light.** `_mat()` sets `roughness = 1.0` and `metallic_specular = 0.2` — fully rough, so there is no specular response at all. The one material property that would make water read as water is dialled to its opposite, uniformly with dirt and stone.

There is no foam, no shoreline, no depth fade, and no reflection of any kind.

## The cost signal worth saying up front

**The project contains exactly one `.gdshader`** — `Scenes/LookDev/vignette.gdshader`. Everything else is `StandardMaterial3D` and generated PNGs. Animated water with a shoreline is the first real shader work in this codebase, and that is the honest size of it: not "author a texture", but "introduce the surface-shader convention, and be the thing every later shader is modelled on".

[...]

Issue #455, opened 2026-08-21
Author more elemental state effects — the board channel, and what justifies the Elemental tab


Phaazoid:
🤖 Claude says:

*(Filed from the dev's direction in conversation, 2026-08-21, alongside [#420](https://github.com/Phaazoid/Godoiosis/issues/420)'s close. His framing: author more elemental state effects **so that the Elemental tab is justified**.)*

## The ask

#420 widened the Game tab's Fire sub-tab into **Elemental**, on the argument that more of this is coming. Today it holds two populations — the fire block (#324) and the cover bump (#326) — and a tab named for a category with two members is a promise, not a category. This ticket is the promise being kept.

## What a terrain state's art costs, and the route it takes

Already answered, twice, and neither answer needs re-deriving:

- **#324 (fire)** — a cell on fire is not one sprite. Several flames seeded from the cell, a generated looping sheet, a real light, all knob-driven.
- **#326 (cover)** — *a terrain STATE whose art draws OBJECTS takes FIRE's route*: it **leaves** `Layer.TERRAIN` rather than forking it, and `OverlayMirror.STANDING_STATES` is the declared list of which states do. That list "earned being a list at exactly two members" — a third member is the natural next step, and the mechanism is already there to take it.

So the cost of a new board effect is mostly authoring plus knobs, not architecture.

## The gaps, concretely

**1. BLAZE and BURNING are visually identical.** #174's *one Fire texture covers both* still stands, and #324 recorded not fixing it as deliberate: *"making the permanent state read fiercer is a design call, not a free win."* The expressive version is already parked — **embers + smoke + haze** (presentation-effects.md Tier 1 and Tier 3). This is the single most-earned item here: two mechanically different states that the board refuses to distinguish.

**2. FROZEN is a flat ground quad, and that was a decision, not an oversight.** #326 states it: *"FROZEN is genuinely flat and stays a ground quad."* Tier 3's *elemental states as materiality* wants the opposite — FROZEN **mirrors**, CHILLED frosts outward with breath-puffs. Revisiting is a genuine design call and should be made deliberately rather than by someone adding frost geometry because fire has some.

[...]

Issue #215, opened 2026-08-13
HD-2D Stage 4a: The Mirror - a real mission as a live diorama


Phaazoid:
🤖 Claude says:

## What this is

Stage 4a of #176, first bite of the dev-chosen **hidden-2D-game-as-authority** architecture (2026-08-13 — the literal Halo-MCC pattern: the 3D battle scene runs the real game invisibly and mirrors it live). 4a is **read-only**: `Battle3D.tscn` loads Prolog through the real load path, sets both factions AI, and the battle plays itself as an HD-2D diorama.

## Scope

- **`Battle3D.tscn` + driver**: hosts the real `Main.tscn` with its 2D view hidden (the sim runs untouched, zero `game.gd` edits — the mirror only reads); look-dev DNA for environment/sun/camera rig; loads the mission via the real funnel; AI-vs-AI so it demos itself.
- **`BoardMirror`**: paints the 3D GridMap from the live 2D grid through the existing kind-at-cell seam (`GridUtils.get_terrain_kind_at_cell`); one declared Kind→meshlib-item table (law-tested complete); the meshlib grows DIRT/MUD/WATER/TREE placeholder blocks. Flat boards — the sim has no elevation yet. **Fire-state cells get flame billboards + real OmniLights** (the torch recipe; Prolog authors 12 BLAZE tiles — the dev's fire-as-light wish arrives early). States refresh each turn (declared v1 approximation).
- **`UnitMirror`**: a per-frame reconcile (no spawn signal exists; `clear_board` frees silently) keyed by instance id (the #149 freed-ref lesson). Position = `unit.position / 16` — **the 2D walk tween is the one animation authority**; walking texture from `movement.moving`, downed from `lifecycle_state`, facing from the frame delta.
- Version → 0.12.0. Look-dev scene untouched.

## Out of scope (the rest of stage 4, each its own PR)

**4b** input (clicks route into the real game — playable in 3D; formalizes the mission-entry seam), **4c** overlays + UI anchoring, **4d** camera pass (the RMB ruling) + the 2D/3D launch chooser + hidden-viewport power savings.

## The gate

Merge = mirrors pinned and falsified. **Closes on the dev's feel-check**: run `Battle3D.tscn` and watch Prolog fight itself in HD-2D — the "how it all looks together" moment.

Part of #176.

— Claude (Fable 5) · 2026-08-13

Phaazoid:
🤖 Claude says:

[...]

PR #569, opened 2026-08-27
#552 slice 1: water reads as water, and shallow reads apart from deep

Phaazoid:
Closes nothing — **[#552](https://github.com/Phaazoid/Godoiosis/issues/552) slice 1 of 3.** A moving, lit, legible water surface; foam/shoreline and reflection are separate and #552 stays open.

## The fork, answered in one line

#552 said the decision it existed to make was *does water stay a `MeshLibrary` item, or become a surface `BoardMirror` builds per cell*, and that the next shader would copy whatever this did. The answer:

> **A cell's own SURFACE is its meshlib item. A MARKER is a per-cell node** ([#324](https://github.com/Phaazoid/Godoiosis/issues/324)'s fire).

So `Scenes/LookDev/water.gdshader` replaces the water items' top material, GridMap batches every water cell into **one draw**, and the waves run off **world XZ** — which is the part that does the work visually: a lake reads as one body of water rather than as N tiles each rippling privately. The issue's worry (*"a meshlib item knows nothing about the cell next to it"*) is true of the *mesh* and not of a shader sampling world position; when foam does need neighbours, that is a board-sized mask as a global uniform, not 87 nodes.

Conventions the next shader copies, all three deliberate: **lit** `shader_type spatial`, never `unshaded` — that is what makes *"the mood presets must be able to grade it"* structural instead of a table row. **Opaque**, so it owes none of the depth-sorting checks `presentation-effects.md:215` catalogues. And it **perturbs `NORMAL` without ever displacing a vertex**, because `BoardSpace.lie_on`, `surface_point`, the hover bracket and every markup quad assume a flat top face at a known Y.

## Shallow vs deep, and where I measured first

Your ask. Before touching anything I cropped the two tiles out of the sheet and counted pixels:

| tile | what the sheet paints |
|---|---|
| `5:6` deep_water | `(77, 155, 230)` in **all 256** pixels |
| `6:6` shallow_water | the **same** colour in 242 of them, plus 11 lighter and 3 white |

Same hue, same value. One of them drowns you. There was nothing there to see, and the hover card had been saying it in words since [#116](https://github.com/Phaazoid/Godoiosis/issues/116).

Two halves now say it, and each answers a different question:

- **Base colour** — an authored `TileData.modulate` per tile (lighter and tealer for shallow, darker for deep). The 2D `TileMapLayer` multiplies it natively and the generator bakes it into the composed atlas, so **one number serves both views** ([#292](https://github.com/Phaazoid/Godoiosis/issues/292)).
- **Behaviour** — the shader's `deep` uniform, which the generator bakes from **the tile's own `walkable` flag**: the flag `RulesService.drowns_in` and the hover card already read. Deep is opaque and slow; shallow is quicker and shows a static mottle of the bed through it. A third water tile authored tomorrow gets its look from its walkability for free.

**Four** water tiles carry the tint, not two: `5:7` and `6:7` are decorated deep variants (a shadow, a weed) that would otherwise have read as *lighter* than the deep tile beside them — a decoration lying about the tile it decorates.

To keep the generator from becoming a second spelling of #109's read, `GridUtils.walkable_of(data)` is the shared tile-level decoder; `BoardContext.is_walkable` delegates its tile half to it and keeps the FROZEN short-circuit that needs a cell. It is a split, not a second answer, and both `CLAUDE.md` and `terrain.md` now say so.

## Your hover constraint

> *"we still have to be able to see tile separation on mouse hover"*

Traced it before designing around it. The 3D hover pointer is a **corner bracket** — a real `MeshInstance3D` volume straddling the cell's top face, not a Decal (4c retired those under your unshaded ruling) — so it does not sample the surface under it and changing water's material structurally cannot make it vanish.

The real risk is legibility, not disappearance: **nothing separates two flat same-kind cells today** (`BoardSpace.SIDE_RIM` is built to make that seam *invisible*, and SSAO cannot see a coplanar boundary), so on a lake the bracket is the only cue — and a thin yellow bracket over a moving, banded, glinting surface is what visual noise eats.

So the shader **draws its own faint cell seam**, and `Cell seam` is a knob with `0` meaning *hand the job back to the bracket alone*. "A lake should look like one body of water" and "I need to see the tiles" pull against each other and only you can say where that line sits — please sweep that one first.

## Also in here

- **Water stopped wearing its own surface down its sides** — #552's premise 1, and the one kind that did it. It now pairs a top with a generated `water_side.png`, as `grass_top`/`dirt_side` have since #215. That also drops the `side_uv` the `_block_mesh` comment warns about, which was a *water* bug in its own words.
- **Nine `GameKnobs` rows**, under a new **Water** group. They reach the shader as **global** shader uniforms rather than material parameters: writing to the meshlib's material would mutate a generated artifact at runtime, and it only misses disk today because `BoardMirror` happens not to be a `@tool` script. That costs one `[shader_globals]` section in `project.godot`, written by `ProjectSettings.save()` rather than by hand.
- **The one premise of #552 that was stale.** The issue says water is MeshLibrary item 5 built from `_water_top()`. That item is only the *declared fallback* — every authored water cell renders a per-tile item wearing the composed **atlas**, so retuning `_water_top()` would have changed nothing anyone sees.

## Flagged, not silently accepted

**A FROZEN water cell will show moving water under a flat ice icon.** `OverlayMirror` mirrors FROZEN as a marker on the TERRAIN layer — [#455](https://github.com/Phaazoid/Godoiosis/issues/455)'s *"genuinely flat"* — and it does not cover the cell. Today that sits over static blue and reads fine; after this it reads as ice on a river. The meshlib route carries no per-cell state, so the proper fix is the same board-sized mask the foam pass needs: one channel saying *frozen*, and the surface holds still and goes pale. Recommending it lands in slice 2 rather than being patched here.

**Unrelated, found on the way, not shipped:** a headless `ProjectSettings.save()` rewrites every input binding's `device` from `0` to `16` — ~12 lines of churn on `project.godot` that has nothing to do with water, and it may not be cosmetic (it is which device `cam_up`/`cam_down`/`cam_left`/`cam_right` match). I reverted it and shipped only the appended `[shader_globals]` block, so it still reproduces on main. Worth its own look.

## Tests

Two suites, 9 cases. Properties and wires, no feel values pinned.

`tests/law/test_water_looks_as_deep_as_it_is.gd` — the `deep` uniform equals `not walkable` for every water item; the shader is on water and **nowhere else**, and on surface 0 only; the shader parses and exposes what the generator sets; the composed atlas carries each tile's authored tint; the three spellings of every global (shader / knob table / `project.godot`) are one set.

`tests/presentation/test_water_knobs.gd` — every knob reaches a uniform of its own name, `_ready` pushes every declared default, every uniform is registered with the renderer. It spies on the push funnel rather than reading back, and says why in the header: measured under `--headless`, the globals **register** but the dummy renderer stores no values, so `global_shader_parameter_get` answers null whatever anyone sets. The one hop no headless test can watch is the funnel's single line into `RenderingServer`, which has no name in it to get wrong.

### Falsification — six mutants, and one of them passed

| mutant | verdict |
|---|---|
| `deep` ignores `walkable` | red — all 7 items of the shallow tile |
| **generator drops the tint** | **PASSED** |
| a knob left out of `_push_all_water` | red |
| a setter pushing its neighbour's name | red |
| shader fails to parse | red |
| shader leaking onto GRASS | red — 28 items |

**The second one is the finding.** The atlas leg compared *mean luminance* of the two composed regions — and the sheet's eleven incidentally lighter pixels in the shallow tile are enough to carry a mean, with or without the tint. The assertion had been riding on the flecks the whole time. A rule about whether a multiply happened has to compare against the multiply, so it is per-pixel against sheet-art × modulate now, with a non-vacuity count; the rewritten version reds at 1024 pixels. Two things it had to learn: the comparison must **clamp** (shallow's modulate is above 1.0 on two channels, so its lightest pixels saturate in an 8-bit atlas), and it reads the **imported** texture, so reverting the PNG alone leaves `.godot/imported/` holding the old one.

Mutants 1, 3 and 4 also each earned a fix — a failure message that named the expected side while reading like a description of the actual one, and two lookups that turned a clean failure into a failure plus a script error.

### One existing test widened, deliberately not relaxed

`test_look_dev.gd::test_tile_materials_are_nearest_filtered_and_textured` cast every surface material to `StandardMaterial3D`. The rule it pins — *a drawn surface names a texture and samples it NEAREST* — is one the water shader satisfies; only the cast stopped covering it. It now asks that rule of either material kind, and holds **every** sampler a shader declares to it, since a second texture sampled bilinear would blur just as visibly.

## Verification

- `law` area green; `presentation` area green.
- Generator run through its full documented cycle (`--textures` → `--import` → `--meshlib`, which refuses once because the atlas changed, then `--import` → `--meshlib`). **Idempotent**: a re-run produces a byte-identical meshlib and no further diff.
- `git diff origin/main...HEAD --stat` is this ticket's files only.
- Two dirty-tree hazards caught and closed: `water_side.png.import` shipped with `detect_3d/compress_to=1`, which silently re-imports on first 3D use — pinned to match its five siblings. And inserting a texture into `_gen_textures` re-rolled the shared seeded RNG and rewrote `tree_top.png` for nothing; it is appended now, with a note saying why.
- `Resources/TestTiles.tres` and `project.godot` were both written by their real writers, never hand-authored, and each diff is only the intended lines.
- **Embedded-content sweep: nothing to sweep, checked rather than skipped.** `modulate` is a `TileData` property on the shared tileset; a scenario snapshot stores atlas coords and references `TestTiles.tres` by uid rather than embedding tile properties, so every committed board picks the tint up on load with no fixture re-authoring.
- Not verified by me: how any of it actually **looks**. That is yours.

## You are on `feat/552-water` — launch it

Nine knobs on the Game tab under **Water**. Worth a sweep in this order: **Cell seam** (the tension above), then **Wave speed** and **Wave scale** together, then **Roughness** and **Specular** against the Moods tab's *Sun elevation*, which is what decides where the highlight falls. **Shallow bed** is the one that carries the shallow-vs-deep read in 3D; the base tint that carries it in both views is the two tiles' `modulate` and is not a knob — say the word and I will move it.


Phaazoid:
## Slice 1b — the two dead knobs, which turned out to be the ice problem

Your feel-check found `Shallow bed` and `Body shade` doing nothing min-to-max, and shallow water reading as ice. Those are one problem, and the shape of it is visible in what you tuned:

| knob | shipped | yours |
|---|---|---|
| Band contrast | 0.35 | **0.8** |
| Ripple | 0.35 | **1.89** |

Contrast, ripple and specular are the only expressive channels slice 1 gave you, and **all three make bright hard highlights** — pale cyan plus gloss plus crisp bands is the recipe for ice. The one knob that would have added *depth* instead was the dead one. You were driving toward legibility with a control set that could only produce more ice.

Your nine values are untouched — they're `47acd62` and they're authored content. Band contrast and Ripple are worth re-sweeping now that there's something under the surface, but that's your call, not an edit I made.

### Why each knob was dead

**`Body shade` was inert by construction.** It ran only on side-facing geometry in **surface 0** — the four *rim* quads — and `BoardSpace.SIDE_RIM` is `0.004`, a rim [#559](https://github.com/Phaazoid/Godoiosis/issues/559) built specifically to be invisible. A block's actual walls are **surface 1**, wearing a `StandardMaterial3D` the shader never touched. Its tip claimed it shaded "a water block's SIDES and its top rim"; the sides half was simply wrong.

**`Shallow bed` was applied but below the perceptual floor, and aimed wrong.** Per-**art-pixel** noise at ±20%, which is roughly one screen pixel at tactical distance, so it averaged to nothing — cranking it multiplied invisible by a bigger number. And it *multiplied the water's own colour*, so it could only make the cyan lighter or darker. There was no sand in it.

### The fix

**The bed is a real layer now, composited UNDER the surface bands.** The old order was surface → bands → multiply mottle over everything, which paints noise across the highlights; that ordering was half the bug. Now: the bed (its own warm colour × a pebble mottle at an authored grain, brightened by caustics) is seen *through* the water, and the bands ride on top additively, because bands are a surface phenomenon.

**Caustics move; the bed does not.** The mottle is hashed off the static art-pixel index so it stays nailed to the board while the light net travels over it. That contrast is the depth cue, and it is the one thing an ice sheet structurally cannot fake — a frozen pane moves all of itself or none of it. Caustic speed *derives* from Wave speed rather than taking a dial of its own; it's the same water moving, and a second dial just invites the two layers to disagree.

**Body shade reaches the real walls.** The shader wears both of a water block's surfaces and forks on the world normal — the up-facing quad is the water's face, the walls and rim are its body. That needed a second sampler (`body_tex`): the two surfaces carry different UVs, so one sampler would have smeared the whole tilesheet down every wall, which is the trap the `_block_mesh` comment records and which was a *water* bug when it last fired. The body samples on a world-derived UV so a wall and a rim agree.

Four knobs replace one: **Bed colour**, **Bed grain** (pebble size in art pixels — `1` is the per-pixel silt that made the original invisible), **Caustics**, **Caustics scale**.

Also fixed on the way: the speckle hash fed `sin()` an argument near 1e5, where float32 carries ~0.01 against a 6.28 period, so neighbouring pixels correlated into blotches. It wraps with `fract` first now.

### On your question about the later slices

**No, and reflection makes it worse.** Foam is a bright edge where water meets land, which ice sheets have too. And reflection is what makes a surface read as ice — that's [#455](https://github.com/Phaazoid/Godoiosis/issues/455)'s own standing point, that SSR is what would make FROZEN read as a literal mirror. Slice 3 pushes the wrong way, so this had to be fixed here.

### Falsification — three mutants, and the third is the finding

| mutant | verdict |
|---|---|
| shader doesn't reach the walls | red — all 4 water blocks, *"1 of its 2 surfaces"* |
| a declared global nothing reads | red — names the knob, its row and its `project.godot` entry |
| **bed composited over the bands, i.e. the original bug** | **passed** |

The first is the assertion slice 1 was missing entirely — it is what would have caught `Body shade` before you did. The third **confirms a limit rather than assuming it**: no headless test can see the compositing order, which is the change that actually fixes the ice read. A new law holds every declared global to being *read*, and its own comment says out loud that it would not have caught either of these two — both were read; one drove invisible geometry, the other was below what the eye resolves. Whether a knob's effect is *visible* is not a question any headless test can ask, and pretending otherwise would be worse than saying so.

`test_only_water_wears_the_water_shader`'s `surface == 0` assertion was **re-justified, not relaxed**: water items wear the shader on both surfaces and nothing else wears it at all.

### Verification

`law` (164), `dev` (374) and `presentation` (407) all green — **`dev` is the one I skipped last time**, which is how a `GameKnobs` group with no `GROUP_TABS` entry reached CI. Generator idempotent; `project.godot` written by `ProjectSettings.save()` with only the `[shader_globals]` hunk kept.

### You are on nothing — `gh pr checkout 569`

Your tree has been moving between other sessions' branches all evening (`feat/520-directed-angle` → `feat/450-guard-link` → `feat/412-413-overwatch`), so I haven't touched it.

Sweep order I'd suggest: **Bed colour** first (warm and desaturated is the whole point — a cool bed under a cool surface is a frozen pane), then **Bed grain**, then **Caustics** and **Caustics scale**, keeping the caustic scale clearly different from Wave scale or the two layers stop reading as separate. Then revisit **Band contrast** and **Ripple** downward.

Phaazoid:
## Next: the dials need to be per water TYPE

Dev, 2026-08-27: *"we need these dials separate for the different water types. Otherwise, I can't tune them separately."* Deferred to the next session — capturing it here so it does not get re-derived.

**He is right and this is my miss.** Shallow's character is currently a fixed RATIO off deep's — `SHALLOW_SCALE = 1.7` and `SHALLOW_SPEED = 1.35`, constants in the shader — and I wrote them as *"the character split, not the pace"* on the reasoning that the knobs set how fast the pair goes while these only say which one is busier. That is a dial you cannot turn. Moving Wave speed moves both types together, so there is no way to make shallow choppy and deep glassy, or the reverse.

**Which knobs actually need splitting**, since it is not all thirteen:

- **Already per-type, no work:** `water_bed`, `water_bed_color`, `water_bed_grain`, `water_caustics`, `water_caustics_scale` are all gated on `(1.0 - deep)` and only ever touch shallow. `water_body_shade` already varies by `deep` through its own `mix`.
- **Need a second dial:** the SURFACE values — `water_wave_speed`, `water_wave_scale`, `water_band_contrast`, `water_ripple`, `water_roughness`, `water_specular`. Those are the six the ratio constants currently derive.
- **Probably stays shared:** `water_seam`, which is about reading the grid rather than about what kind of water it is. Worth asking rather than assuming.

**The constraint that shapes the answer, so it is not rediscovered:** every tuning value is a GLOBAL shader uniform precisely because a per-material one would mean `BoardMirror` writing into `lookdev_meshlib.tres` at runtime — mutating a generated artifact, which is the dirty-tree rule's acceptance half. So *"bake it per material"* is closed off. The shape that survives is **two sets of globals** (`water_shallow_*` / `water_deep_*`) with the shader picking between them on its existing `deep` uniform, and the ratio constants deleted rather than kept as a fallback.

That makes it ~6 new knobs and ~19 rows in the Water group, which is past Fire's sixteen. Whether that wants its own sub-tab, or whether some of the six turn out to be genuinely shared once he can A/B them, is a question for him rather than an assumption for me — the shipped ratio was me guessing at exactly that and guessing wrong.

Not building it tonight; PR stands as-is and is green.

Phaazoid:
## Slice 1c — every dial is per water type

Your ask: *"we need these dials separate for the different water types. Otherwise, I can't tune them separately."* Done, plus the second instance of the same defect that you found for me in the same breath.

**The generalisable form, since it is the tuning rule violated while appearing to follow it: a RATIO BETWEEN TWO AUTHORED THINGS IS ITSELF AN AUTHORED THING.** Shallow's character was a fixed ratio off deep's — wave scale ×1.7, speed ×1.35, body shade ×0.6 — living in the shader as constants, which I justified as *"the character split, not the pace"*. Burying it in a const does not stop it being a feel value; it makes it a feel value with **no surface**.

### What split, and what only got renamed

| | knobs |
|---|---|
| **split into a deep and a shallow dial** | wave speed, wave scale, band contrast, ripple, roughness, specular, **body shade**, **cell seam** |
| **renamed only** — already shallow-only, gated on `1.0 - deep` | bed, bed colour, bed grain, caustics, caustics scale |

**Cell seam splitting was your call over mine and it is the better one** — the bed's mottle already breaks shallow water up while a deep expanse has nothing else going on, so the two genuinely have different grid-legibility needs. That is now written into both tooltips so the next person tuning them knows which way to push.

The bed and caustics carry `shallow_` in their names even though they have no deep twin. They are shallow-only *by nature* rather than by omission — deep water is opaque — but a panel where some rows name their water and others do not is one you have to remember rather than read.

Twenty-one knobs, so **Water took its own sub-tab**; it reads as *Water (deep)* / *Water (shallow)*. The shader picks with `mix(shallow, deep, deep)`, exact because `deep` is baked 0 or 1 and still sensible if a fractional depth is ever authored. The ratio constants are **deleted**, not kept as a fallback — a deleted seam cannot drift back.

### It ships as a visual no-op, deliberately

Each deep dial takes your tuned value; each shallow dial takes **that value times the ratio the shader was already applying**. So `wave_speed` is deep 2.32 / shallow 3.132, `wave_scale` deep 4.4 / shallow 7.48, `body_shade` deep 0.3 / shallow 0.18, and the five with no ratio take the same number on both sides. **The board looks identical the moment it lands — only the panel changes.** Your values are authored content and a split that silently re-tunes them loses your evening.

Your `tuning` commit had never been pushed, so this push carries it too — caustics 0.22 → 1.35 and the net down to 18.5 are in.

### No generator change, no meshlib churn

The per-material uniforms (`atlas`, `body_tex`, `deep`) are untouched; only globals moved. `git status` leaving `lookdev_meshlib.tres` alone is the proof, and it is why the diff is smaller than it sounds.

### Falsification — three mutants, and the third one caught me out

| mutant | verdict |
|---|---|
| shallow reads the deep global | red — via the *read* law: the shallow dial becomes an unconsulted global |
| a knob group with no `GROUP_TABS` entry | red — *"Knob groups with no sub-tab: Water (shallow)"* |
| a uniform naming neither type | red — **but only after isolating it** |

That third one is worth telling you about. It first reddened the **spelling** law, which sits earlier in the same file, and a failing case truncates the rest of its suite — so six of eight cases ran and the law I was actually aiming at **never executed**. Reported naively that would have been a green mutant recorded as a red one. Grepping which case fired is what caught it; isolating past the earlier case then showed the real red.

New law, and it is the regression guard for exactly this ticket: **no water uniform is ambiguous about its type**, and every deep one has a shallow twin. The reverse is deliberately not required — the bed has no deep twin by nature, and a law that overstates its own rule is worse than none.

Also here: `test_every_knob_group_has_a_sub_tab` is the case CI caught me with two slices ago, and **`dev` ran this time** — 375 cases. `law` 181, `presentation` 424.

### You are on `main` — so I can hand you the branch

```
gh pr checkout 569
```

Sweep order: pick one type, get it right, then the other — that is the whole point of the split. On **shallow**, the bed and caustics are what stop it reading as ice, so let those win: Band contrast and Ripple are worth pulling *down* from the 0.8 / 1.86 you set when they were the only expressive channels there were. On **deep**, those two are still the whole show, so they can stay high. Cell seam is where the split should pay off most obviously — deep wants more of it than shallow does.

Issue #567, opened 2026-08-27
Overwatch's triggered shot should interrupt the walk, not play after it


Phaazoid:
Overwatch's damage lands at the crossing moment in the RESOLVE — that is where every number in
[#413](https://github.com/Phaazoid/Godoiosis/issues/413) comes from — but the shot **plays back after
the move phase**, not during it. This is the declared v1 cut, taken with the dev's own words at build
time: *"we want #2, but it can be a separate issue once we get the mechanics working."*

## What it looks like today

`OrderExecutor.execute_orders` runs the parallel move phase to completion, then plays
`plan.watch_shots` in trigger order, then the attacks. A crosser the shot **downs** has already
stopped at the crossing cell (`MoveAction.walked_path()` is the truncated prefix the resolver
stamped), so nothing is *wrong* — the outcome, the HP numbers and where every unit ends up all match
the preview exactly. What trails the fiction is the ORDER OF THE VISUALS: everyone finishes walking,
and only then does the carbine fire at the cell somebody crossed a moment ago.

## What it should look like

The crossing mover halts at the crossing cell, the shot plays, and the walk resumes — the crosser
continuing on unless the shot downed or threw them (both of which already end the walk in the rules).

## Why it was not built with #413

Two reasons, both structural rather than effort:

- **It breaks the parallel move phase apart.** `_execute_action_phase_parallel` starts every move at
  once and awaits them together, which is the whole of "units are still shown moving together"
  ([standing-reactions.md](https://github.com/Phaazoid/Godoiosis/blob/main/docs/design/standing-reactions.md)'s
  own presentation rule for #412). An interrupt needs one mover's walk to be pausable while its
  siblings keep going, which is a different playback shape, not a flag.
- **It lands on [#520](https://github.com/Phaazoid/Godoiosis/issues/520)'s live choreography work.**
  The camera goes to the walk and the walk waits for it; a mid-walk beat is a camera question as much
  as an animation one, and the two arcs would have been rewriting the same function at the same time.

## Where to start

`OrderExecutor` line ~111 is the whole seam — the move phase, then `plan.watch_shots`. Each shot
already carries `AttackAction.triggered_by` (the crosser) and the crossing cell is
`MoveAction.path[resolved_stop_index]` when the walk was halted, so which mover to pause and where is
already in the plan; nothing new has to be resolved. `MoveAction.execute()` walking
`walked_path()` is the other half — an interrupt splits that into walk-to-the-crossing, await the
[...]

Issue #521, opened 2026-08-25
Battle zoom: staging displacement and the tear-out diorama


Phaazoid:
Slice 4 of #410 (battle zoom). Placeholder geometry — flat slabs, no strata.

## Ask

The tear-out and the diorama itself.

- **Entry:** participating tiles rise out of the board, off the top of frame; white-out; fade into the diorama. Tiles slam into the diorama **one by one in queue order**, quietly foreshadowing who acts first.
- **The diorama is a true 3D reconstruction** — board-relative offsets preserved, horizontal *and* vertical. Wide shots keep true scale.
- **Tear-out set = slice 1's touched-cell set**, already computed once from the plan.
- **Exit reverses it:** white-out, tiles thud back into their sockets with a dust shockwave, the aftermath already written on the health readouts.
- **Feels-test knob, both built, behind dev tools:** (a) participants only, bystanders stay behind on the board; (b) bystanders ghosted in for spatial context.

## The seam (dev ruling on #410, 2026-08-25)

**Staging displacement in `BoardSpace`** — one new question, *"is this cell staged, and where?"*, answered where `BoardSpace` is already consulted. Zero displacement is the board. Both mirrors honour it for free. **Not** a separate diorama scene: that would be a second answer to "where does this render", which `BoardSpace.gd`'s own header forbids.

Load-bearing: `UnitMirror` is a **per-frame reconcile** — "poll, don't wire", position recomputed from the unit's 2D position every frame. Moving a unit's node stages nothing; it snaps back next frame. The displacement must be consulted at the placement seam.

## Done when

- A pass tears out, plays in the diorama, and thuds back with board state intact.
- Both feels-test modes are reachable from dev tools and playable.
- With the cinematic off, displacement is provably zero everywhere — a law that reads the placement seam directly, not a surface something rebuilds.

## Notes

- The white-out must read #217's photosensitivity switch.
- Deferred to the art pass: strata on the cut edges (dirt, rock, dangling roots, dripping water), backdrops, per-terrain sets.

Phaazoid:
Closed by #652, with one clause of the Ask deliberately unbuilt: **the dust shockwave on the exit thud**. Cut on your call (2026-08-28 — *motion in, particles out*), because there is no particle system anywhere in the project and `HealthBlockDebris` is the only precedent, so it is new machinery rather than more of the same.

[...]

Issue #450, opened 2026-08-21
Guard's board markup: the defender-to-defendee connector, and whether a QUEUED Guard shows anything


Phaazoid:
[#414](https://github.com/Phaazoid/Godoiosis/issues/414) shipped Guard with a working ward mark (a shield decal underfoot, its own 3D plane, both views). Two pieces of its board presentation were deliberately left out, and both are named in `docs/design/standing-reactions.md`.

## 1. The connector — the design asks for it, the build does not have it

Canon: *"a shield icon under the defendee with a connecting mark from the defender, one clear mark under the restraint doctrine."*

What shipped marks **both** ends of the pair with the same shield, because with one placeholder texture a "connector" would just have been a second identical ring. That says *these two units are linked* but not **which way** — you cannot tell the bodyguard from the warded unit by looking at the board. Today the direction is only readable from the queue row (`"Enemy -> Tank (guarding Alia)"`), which is gone the moment the pass ends — i.e. exactly during the enemy phase, when the information matters most.

Blocked on art more than on code: `OverlayManager._draw_arrow_trail` already draws a tinted trail between two cells and would take a two-cell path, but the path-arrow vocabulary means *movement*, and a ward link is a relationship — reusing it would give one channel two meanings. A drawn link (or an asymmetric pair: shield under the ward, something else under the blocker) is the real answer.

## 2. A QUEUED Guard shows nothing on the board

Currently the ward mark appears only once the Guard has **armed**, which happens at execution. Between queueing the order and pressing Execute there is no board feedback at all — the queue row is the only sign.

Every other queued thing previews: moves get path arrows and a destination ghost, attacks get the reach and aim overlays, a shove gets its trail and landing ghost. A queued Guard is the odd one out.

This surfaced from the dev's own words when picking the sprite — *"the one in the 35th column for the floor to indicate **queued** shielding"* — and was never resolved during the build, so it is an open question rather than a known omission:

- Should a queued-but-unarmed Guard draw a mark at all, or is the queue row enough under the restraint doctrine (*"fancy effects in the correct places, not everywhere"*, [#346](https://github.com/Phaazoid/Godoiosis/issues/346))?
- If yes, does it look **different** from an armed one? Queued and armed are genuinely different states — one is a plan, one is live cover — and drawing them identically would be the board telling the player they are protected when they are not yet.
[...]

Issue #414, opened 2026-08-20
Guard: bodyguard substitution as a main action


Phaazoid:
Ratified in the 2026-08-20 standing-reactions grill session. Full canon: `docs/design/standing-reactions.md` (lands with the session's docs PR) — this body is the build summary, the doc is authoritative. Named **Guard** (not "Defend") to stay distinct from DEF the stat; it absorbs jobs.md's unbuilt "Guard redirects" follow-up. **Provenance: an adaptation of the guarding feature from the GameMaker-era demo** — a beloved survivor of the pre-squad battle system.

## The locked spec

- **A main action: pick a unit in range and become its bodyguard.** Working model (leaning, playtest-subject): Guard is a BASIC action everyone has; what kit grants (weapons like the Aegis-class drill, jobs, armor, carvings — source-agnostic) is the **brace bonus** — bonus DEF applied to the absorbed instance. Bare Guard substitutes at +0.
- **Total victim substitution.** A hit on the ward resolves against the blocker instead — the blocker's DEF/armor + the brace bonus, the ENTIRE payload (damage, elemental states, knockback), resolved from the blocker's own cell. The ward takes nothing. Per-hit resolver hook (Iron Will's shape), previewed with real lethality rungs.
- **The pair is unit-and-range, not cells**: the Guard targets a specific unit and holds while that unit is within the Guard's range of the blocker (range authored on the granting content, default 1). Displacement that keeps them in range keeps the ward (the AoE-wind-blast case); separation beyond range drops it, however it happens. Range is checked at each hit's resolution moment.
- **Catches ANY damaging hit on the ward** — authored attack, counter, overwatch shot, hostile or friendly — with one exception: an already-substituted hit (chains ON HOLD, deliberately not buried — silly edge cases may want a tank conga someday). Heals and pure-utility attacks pass through. Friendly-fire coverage is player agency: the Guard arms at its queue slot, so sequence your own splash before it.
- **"Can't defend while moving" (dev ruling):** no Guard is live during its own faction's move phase (last turn's lapsed at turn start; this turn's arms after moves), so a mover crossing a watch is never covered — Overwatch owns punishing movement, Guard owns the stationary window; the overlap exists only where the enemy manufactures it (a shove during the enemy phase CAN be blocked).
[...]

PR #570, opened 2026-08-27
Overwatch, and the queue order that makes it a decision (#412 + #413)

### LINKED ISSUE #412: Queue order becomes real: moves resolve in queued order, with drag-reorder in the action queue
Ratified in the 2026-08-20 standing-reactions grill session. Canon: `docs/design/standing-reactions.md` (lands with the session's docs PR). This is the structural prerequisite for Overwatch — its trigger clock.

## What

**Move orders resolve in queued order.** Today moves resolve as one unordered parallel set; after this, the resolver walks them in the order they sit in the queue, and the player can **drag-reorder** move rows in the action-queue panel. Units are still *shown* moving together — simultaneity survives as presentation. The order is the clock for mid-move events (an overwatch trigger is the first customer; future on-entry hazards inherit it).

This extends **R6** (resolution-pipeline.md): attacks already resolve sequentially in queued order as the declared skill expression. Moves join them.

## What deliberately does NOT change

- **Whole-plan validation.** Destination conflicts and formation rotations are order-independent questions; `SquadPlanValidator` is untouched. Two squadmates swapping cells stays one legal simultaneous fact even though the resolver walks the moves in order (transient ally co-occupancy during playback is the rotation fiction, already legal — allies never block).
- **Law #2 machinery.** A reorder is a plan edit: it re-runs the synchronous resolve→repaint chain, so derived consequences (overwatch shot annotations) jump rows the moment you drag. Queue/cancel/requeue idempotency holds.
- **Group Move stays one authoring gesture** — one batch, one gate pass, solver emits in member order as the *default*; rows inside a batch are individually reorderable afterward (member order is now mechanically real, no longer just row-stability cosmetics).

## AI note

The AI's queue order becomes load-bearing when its squads cross a standing watch (it feeds the player whichever member its builder queued first). v1 ships naive; ordering smarts ("tank first") are part of the standing-reactions AI pass filed on #117.

## Build map

Demo milestone. Overwatch (see the blocked issue) depends on this; Guard does not.

— Claude (Fable 5) · 2026-08-20, from the grill session

### LINKED ISSUE #413: Overwatch: a standing watch that fires on entry (Carbine identity)
Ratified in the 2026-08-20 standing-reactions grill session. Full canon: `docs/design/standing-reactions.md` (lands with the session's docs PR) — this body is the build summary, the doc is authoritative. **Motive: finally give Carbines a real identity**, and push the player toward predicting the enemy phase (the throughline).

## The locked spec

- **The watch is a pre-aimed shot.** Declaring overwatch = aiming an attack through the normal targeting flow; the watched footprint = that aim's affected cells + facing, frozen at declare time (the `fired_attack` stamp pattern). "Tile or line" falls out of the existing point/directional targeting-mode split. Capability is an authored flag on the `AttackData` **base**, so carbine attacks and transmutation carvings ride one mechanism.
- **Fires once, then spent** — the triggered shot spends one magazine round like any carbine shot. (Fires-at-every-crosser is future keystone-mod space, not v1.)
- **Trigger: an ACTIVE enemy of the watcher enters any watched cell, however it got there** — walking, or displacement (shove combos are intended: arm the watch, knock them into it; the watch arms at its slot in the resolution order, so queue order decides whether the combo connects). Allies never trigger it, but a triggered shot is the attack, full stop — ally splash lands if authored. Arming does not fire on units already standing in the footprint; movement between two watched cells is an entry; stepping out is not. Downed bodies don't trigger ("probably" — AI self-preservation exploration open). Non-movement entries (rescue placement) don't trigger in v1 (the rescue-into-the-line trigger is recorded as future content — dev finds it very funny).
- **Basic shots don't interrupt the move** — damage lands mid-move at the crossing moment and the crosser continues; a crosser the shot DOWNS stops at the crossing cell (lifecycle, not a new rule). **Stopping power is an authored per-attack flag** (default off, mod-grantable; whether it becomes default is a playtest question answered in content). Involuntary displacement ends a move — a knockback overwatch shot is stopping power in disguise. The halted-but-alive unit's remaining queued orders: parked with the stopping-shot content pass.
- **Reactions grammar: counters listen to authored attacks; watches listen to entries.** A triggered shot draws no counter (derived attacks are never counter-bait). Watch-to-watch chains ARE legal — a shot's knockback throwing the crosser into a second watch triggers it (entry is entry). Deterministic pinball, fully previewed.
- **Lifetime: the shared standing-reaction grammar** — arms at its queue slot; absorbs exactly one trigger; if untriggered, lapses when the owner's faction's next turn begins; the owner leaving the anchor cell (shove, rescue, lifecycle) drops it.
- **Queue representation: an annotation on the crossing MOVE's row** ("triggers Overwatch — takes N from Carbine (E)"), a deliberate divergence from the counter section, chosen so drag-reorder feedback lives on the row being dragged. The watcher's own declaration is an ordinary main-action row. Multi-shot crossings stack annotations on the row.
- **Board display: the watch footprint is always visible, both directions**, with its own visual vocabulary — crosshair/flash-shaped threat marks, NOT a filled-square range layer. Must reach both 2D and 3D views (gameplay information; no #292 asymmetry allowed).
- **AI: `NEVER` in every archetype table at v1** — the whole AI side (Sentry watch stance, watch-aware pathing, crossing-order smarts) is deferred to the bigger AI pass, filed on #117.

## Dependencies

**Blocked by the queue-order issue** (its trigger clock). Demo milestone.

— Claude (Fable 5) · 2026-08-20, from the grill session

Phaazoid:
Closes #412. Closes #413.

Both on one branch, decided up front: **#412's clock half has no consumer without Overwatch, and Overwatch is what defines the shape the walk has to answer.** Building the walk blind would have been a coin flip on the one seam everything else hangs off.

## What you can do now

Pick a carbine, choose **Overwatch** from its own ring slice, pick the attack to watch with, and aim it the ordinary way — same reach, same facing, same overlay — and it holds fire. The aimed cells stay marked on the board for **both sides**, and the first active enemy who **enters** one takes that attack where they stand. Fires once, then spent; lapses when your next turn begins; drops the moment the watcher leaves the cell it aimed from.

And on your side of it: the queue panel's **MOVE section is draggable**, so which of your units crosses first is a row you set. Drag it and the annotation on that row — *"triggers Rook's watch — takes 12 from Shot"* — moves with it.

## The three things worth reviewing

**1. Seeding every mover at its ORIGIN is the whole of #412.** `resolve_plan` grows a move phase ahead of the attack walk, and before any mover walks, all of them are pinned to where they started. That is what makes the ordering real rather than a list order: a shot fired while mover #1 crosses gathers its victims with movers #2 and #3 still standing at their origins. It threads through `PlanResolver.projected_position(unit, hypo)`, the incumbent seam, not a second projection. **When the block ends every uninterrupted mover is back at its destination**, which is the regression guarantee for everything downstream — pinned by its own case.

**2. Watches listen to ENTRIES, and there are exactly two.** A step of a queued walk, and the landing of a shove. One rule, one implementation, which is what makes the shove combo and the watch-to-watch chain the *same* mechanism instead of two features that have to be kept in step.

**3. Counter-immunity is structural.** The shots live in `plan.watch_shots`; `calculate_reactions_for_squad` reads `plan.attacks`. "A derived attack is never counter-bait" is true because the list it would have to be in is a different list.

## Three things the ticket said that the code disagreed with

Re-derived before planning (#228), and it changed the build:

- **The drag already existed** — press, live re-parent, release, emit — gated to attack rows alone. #412 needed the gate opened, not new machinery. Nothing tested the incumbent either; it does now.
- **The clock already existed** — `resolve_plan`'s walk over `action_queue` is what #414's Guard arms on. Moves were simply invisible to it.
- **`GroupMoveSolver` already emitted in member order**, so "member order becomes the default row order" needed no code, only a consumer.

## Rulings I made — argue with any of them

- **`can_overwatch` is stock on the Carbine's Shot AND mod-grantable**, your call at build time. The mod side composes exactly like `hits_allies_override` (OFF beats ON, so fitting order can never change what a weapon can do).
- **A halted walk defers to a shove.** `Unit.projected_cell`'s *"a real move beats the shove: you walked out from under it"* gains *"...unless the pass stopped you"*.
- **`OVERWATCH` sits immediately before `GUARD` in `SIDE_CHANNEL_ORDER`**, for GUARD's own reason.
- **A QUEUED watch draws no footprint yet** — only an armed one does. That follows #414 exactly (a queued Guard draws no ward marker either) and the queue row is the preview. Say the word if you want the preview on the board instead.
- **Overwatch is a WEAPON ACTION, one row, no submenu** -- *your call, 2026-08-26, and it replaced mine.* I shipped it as its own top-level ring slice; you said *"Overwatch should be a weapon action, not it's own top rung on the menu"* and then *"The Carbine will get 1 Overwatch action as a special weapon action. That's it. There will be no sub menu."* It now sits in the kit slice beside Reload, named for the VERB, and clicking it picks the attack. The top-level category, its menu id, its `ACTION_DATA` row and its `populate()` gate are all gone, so the GATE moved onto `Unit.has_weapon_actions()` and follows that rule: a watch you could fire right now opens the slice, one you cannot still lists greyed with its own reason. The LABEL rule is yours too -- since a mod-granted watch *"will have their own special distinguishing names"*, each is its own flat row under its own name and no label can collide with its firing twin.

## The two declared cuts

- **Playback: shots play after the move phase, not mid-walk.** Every number is the crossing moment's and a downed crosser has already stopped at the crossing cell — only the order of the visuals trails the fiction. Filed as #567, as you asked, because it breaks the parallel move phase apart and lands on #520's live choreography. The camera does go to the crossing, so the shot is not off-screen.
- **The board mark is PLACEHOLDER art** — the board cursor's brackets, tinted, at 16px (one cell — #414's lesson). Colour and size are GameKnobs rows with a sweep, so they're not born-dead sliders. The crosshair cut is your pick; a knob can't fake a sprite.

## Sweeps and hygiene

- **Embedded content:** nothing embeds `AttackData`. Every committed scenario references its weapon TEMPLATE as an `ext_resource`, so the canonical attack file is what every board reads and no fixture is stale. `ScenarioUnitEntry`'s five new fields are all at their defaults for every committed board, so no `.tres` changed.
- **One file did change, and it is a pre-existing defect this diff fixes:** saving `Carbine.tres` through the real writer moved `scaling_blend` to the position Godot actually serializes it at. The committed file had been hand-authored at a position the engine does not use — #111's shape — so your first editor save would have handed back a modified file you never edited. Verified idempotent (second pass byte-identical).
- **Canon swept:** `standing-reactions.md` (both issues BUILT + every ruling the build cost), `resolution-pipeline.md` R6 + *Movement order*, `weapons.md`, `weapon-mod-ideas.md` #26, `visual-clarity.md`, `grill-queue.md` (item 17's warning fired — the direction question WAS answered by the build, with the doc's own maximal ruling), `philosophy.md`, `squad-system.md`, `CLAUDE.md`, `BACKLOG.md`. Markers bumped.

## Verification

Area suites only, per the standing rule. `law` + `squad` + `elemental` + `dev` + `flow` + `ui` + `presentation` + `play` green.

**Falsified with nine mutants, nine kills, no survivors** (the last three are the menu rework's):

| mutant | what reddened |
| --- | --- |
| `seed_at_origin` no-op | *a later mover is still at its origin when an earlier shot lands* |
| walk the path from index 0 | *stepping out of a watched cell is not an entry* |
| never mark the watch spent | *a watch absorbs exactly one trigger* |
| never stamp the halt | *a crosser the shot downs stops at the crossing cell* |
| shots also go into `plan.attacks` | *a triggered shot draws no counter* |
| walk the moves in reverse queue order | *the first mover in queue order eats the shot* |
| point the Overwatch row at an ordinary attack aim | *the row opened an ordinary shot's aim, not a watch* |
| drop the watch clause from `has_weapon_actions` | *a watchable attack opens the weapon row* |
| label the row after its attack instead of the verb | *the kit slice drew no Overwatch row* -- `_append_unique` eats it |

The second one is the finding worth naming: I wrote the *stepping out* case **before** running the mutant, on the suspicion that nothing covered "the origin is not an entry" — and it was right. Without that case the mutant passes clean.

**What is NOT covered by a test, so you know where to look:** how the mark actually reads on the board — its colour, its size, and whether a footprint under a squad ring or a frost icon is legible. Its 2D z-index and its 3D layer sort are both reasoned rather than measured, and nothing pins a 2D marker's z. That is the half only playing answers.

## What is still owed

The AI side of all three standing reactions (#117, unchanged), the real mark art, and #567.

---

**You are NOT on this branch** — your tree is on `feat/520-directed-angle`, and the rule says an occupied tree stays put. To play it: `git switch feat/412-413-overwatch`. The worktree I built in is already released.

---

**Merged main twice, and the second one had a real collision (2026-08-26).**

#521's tear-out landed first: the only conflict was `OrderExecutor`'s phase order, where both sides insert after the walk. Kept both, **watch shots before the tear-out** -- a watch fires at a crossing CELL and #521's stage set holds only what MAIN ACTIONS touch, so lifting the ground first would play the shot over a hole. #521's own rule settles it: the board is where you move, and a watch shot is the tail of moving.

Then #566 landed, and it is the overlap I flagged above -- three conflicts, two of them trivial keep-both (a `GameKnobs` sweep arm, the canon marker) and **one that was not**: #450's `GUARD_LINK` and #413's `WATCH_ICONS` had each taken `BoardOverlays` **sort 9**, both reasoning "the first slot free above the ward shield." A layer is a plane and a warded pair standing in a watched line is an ordinary board state, so they cannot share it. **The watch mark moved to 10**, because `GUARD_LINK`'s 9 encodes a stated relationship to the shield's 8 while the watch mark only cares that it is not *in* the pair.

**The 2D cannot follow, and I have declared that rather than left it as drift.** There is no integer between the arrow band (3) and the units (4), so shield, link and watch mark all sit in that band and tree order decides -- and `ArrowIconOverlay` is a later sibling than `IconOverlay`, so on a cell carrying both, **the link draws on top in 2D and the watch draws on top in 3D**. Two 16px marks on one cell; #450's own comment already accepted a shared band for the shield. Stated at the source (`OverlayManager._rebuild_watch_sprites`) and on #292's terms. Fixing it means re-banding 2D markup, which is a bigger call than this PR should make.

Verified after the second merge: `presentation` 421 green, `dev`+`law`+`ui` 804 green.


Phaazoid:
🤖 Claude says:

**Heads up on merge order — this overlaps [#566](https://github.com/Phaazoid/Godoiosis/pull/566) (#450's guard markup) in seven files.** Both are `mergeable: clean` right now because each is measured against `2dff4f7`; whichever lands second will conflict. Every conflict is a **keep-both**, not a real disagreement — the two touch adjacent rows of the same tables:

| file | what collides | resolution |
| --- | --- | --- |
| `Classes/presentation/BoardOverlays.gd` | both append to `Layer` + the `LAYERS` table | keep both entries; my `WATCH_ICONS` must stay at a sort no other layer on a shared cell uses (9 today, above the ward's 8) |
| `Classes/dev/GameKnobs.gd` | both add `CLASS_KNOBS` rows and their read/write/sweep arms | keep both rows and all six arms |
| `Classes/presentation/OverlayMirror.gd` | both extend `_icons` | keep both marker lists |
| `Classes/board/OverlayManager.gd` | #450 reworks the ward draw, I add the watch draw beside it | keep both; they share no function |
| `Classes/actions/OrderExecutor.gd`, `game.gd` | adjacent `refresh_*_markers` call sites | keep both lines |
| `docs/design/standing-reactions.md` | both edit the banner + canon marker, different sections | keep both sections; take the HIGHER canon marker |

The one to actually read rather than accept blindly is **`BoardOverlays.gd`'s sort**. A layer is a plane, so a marker that can share a CELL with another marker must not share its sort — and a warded unit standing in a watched line is an ordinary board state, not a corner case. If #566's final shape moves the ward off 8, mine wants re-checking against whatever it lands on.

Happy to do the merge on whichever goes second — say the word and which one you want first.

Phaazoid:
🤖 Claude says:

**This PR's CI is red for a reason that has nothing to do with it — don't read the X as mine.**

The failure is one case, `test_queueing_a_move_drags_the_link_along_with_the_shield`, and it is `main`'s: your `c70c2e9 "tuning"` moved `GUARD_LINK_HEAD_INSET` 0.5 → 0.4 and that case asserted the arrowhead at `from.lerp(to, 0.5)`, which is the inset restated. `main` is red on its own for it. A PR's CI tests the merge of its head with the base tip, so this branch inherits it.

Confirmed by reproduction rather than by reading: the suite is green in isolation on this branch, and applying your 0.4 to it locally reds exactly that case with exactly that message. **2502 cases ran, 1 failed** — every part of this ticket is green.

Fixed by [#574](https://github.com/Phaazoid/Godoiosis/pull/574) (another session got there first; I closed my duplicate). Once that lands, re-run this and it goes green with no change here.

One thing worth knowing while it stands: a failing case truncates the rest of its own suite file, so that one failure was also hiding the **six cases after it** in `test_guard_lifecycle_wires.gd`. All eleven pass under #574's fix.

Issue #579, opened 2026-08-27
The shallow lakebed is one flat colour times hash noise, and it now carries the whole shallow read


Phaazoid:
🤖 **Claude says:**

*(Filed from the dev's own tuning sweep on [#552](https://github.com/Phaazoid/Godoiosis/issues/552) slice 1c, 2026-08-27. Sibling: deep water has no depth layer.)*

## The observation

Slice 1b added a lakebed under shallow water — a warm colour with pebble grain, and caustics travelling over it — to stop shallow water reading as ice. It worked. The dev's sweep then took `water_shallow_bed` to **1.0, its maximum**, and cut `water_shallow_caustics` from 1.35 to **0.37**.

That is close to the inverse of how the slice was built. The reasoning at the time was that the moving light net is the depth cue, because *"a frozen pane moves all of itself or none of it, so this is the one thing ice structurally cannot fake"* — so caustics shipped loud over a faint bed. His values say the opposite: **it is seeing a bottom at all that kills the ice read**, and the net on top of a strong bed is closer to noise.

## Why that makes this worth a ticket

The layer now carrying the entire shallow-water read is the least authored code in the shader:

```glsl
vec3 bed = water_shallow_bed_color.rgb * mix(0.72, 1.28, speckle(grain));
```

**One flat colour times hash noise.** `water_shallow_bed_color` is a single `Color(0.60, 0.52, 0.38)`, and `speckle()` is a per-pixel hash. There is no structure to it — no sand against rock, no variation at any scale larger than the grain, and no relationship to what the tile above it actually is.

That was defensible when the bed was a faint tint under a busy caustic net. It is not obviously defensible now that it is what shallow water *is*.

## Candidate directions

Not a decision, just what is on the table:

1. **Let the tile's own art read through.** The water tiles carry authored `TileData.modulate` (slice 1) and the ground atlas has real terrain art in it. A bed derived from the *neighbouring ground* would make a lake in a rocky map look different from one in a grassy map, for free, and would tie the bed to authored content rather than to a constant.
2. **Two-tone: sand against rock.** A second bed colour plus a low-frequency mask, so the bottom has patches rather than uniform speckle. One extra colour knob and one scale knob.
3. **Grain at more than one scale.** Cheapest of the three — the existing `speckle()` at two frequencies. Probably the smallest real improvement too.

## Constraints that already apply

[...]

Issue #578, opened 2026-08-27
Deep water has no depth layer, and the shallow/deep colour step needs blending


Phaazoid:
🤖 **Claude says:**

*(Filed from the dev's own tuning sweep on [#552](https://github.com/Phaazoid/Godoiosis/issues/552) slice 1c, 2026-08-27. Sibling: the shallow bed's structure.)*

## The observation

After the per-type dial split shipped, the dev swept both water types and committed his values (`bb31215`). Reading them back, deep water came out the **thinner** of the two reads:

| layer | shallow | deep |
|---|---|---|
| bed (what is under the water) | yes, at **1.0** — his maximum | none, by design: deep water is opaque |
| caustics | yes | none, same reason |
| moving bands | yes | yes |
| ripple (normal bend) | 1.86 | **1.32** — he pulled it down |

So deep water is now **a flat tinted texture plus moving stripes**, and that is the whole of it. Shallow water gained two layers over slices 1b/1c; deep gained none, and then lost some of the one channel it had.

## Why this matters, and why it is not a bug

Nothing here is broken and none of it is a mis-tune. Deep water being opaque is correct, and the bed being shallow-only is a deliberate call recorded in `BoardMirror.gd`'s own comments: *"they are what you see THROUGH shallow water, and deep water is opaque."*

The gap is that **nothing in the shader says *bottomless***. Shallow water reads as shallow because you can see its bottom. Deep water currently reads as *shallow water with the bottom switched off* rather than as depth. The one cue it could carry — vertical travel in the surface normal — is the cue the dev turned down, presumably because at slice 1's settings it read as glare rather than swell.

## What might say it

Not a decision, just the candidates, cheapest first:

1. **Distance from shore.** Water darkening as it leaves the land is the single most legible depth cue in a top-down view, and it costs nothing new: [#552](https://github.com/Phaazoid/Godoiosis/issues/552) slice 2 builds a board-sized water mask for the shoreline, and `filter_linear` on it yields a distance ramp for free. **This issue is best picked up after that slice lands**, since the data will already be there.
2. **Slow, large swells.** A second wave octave far slower and wider than the current bands — a big body of water moves on a longer wavelength than a puddle, and that difference is readable without any brightness.
[...]

Issue #285, opened 2026-08-15
Paint elevation in the 3D view, with a ghost voxel showing where the next click lands


Phaazoid:
🤖 Claude says:

**The successor #231 named and nobody filed** — the same gap that became #273. #231 (closed) gave the dev brush 3D input and a half-transparent ghost of the real block; what it never gave it was **height**. A click in the 3D view still paints on the flat plane.

Dev, 2026-08-15, straight after the #273 feel-check: *"the next obvious step is being able to paint this while in 3D, with ghost voxels showing me what I'm going to paint."*

Now unblocked: #273 makes the board actually *show* elevation, so painting at height in 3D has something to paint onto.

## The decision this needs first — what does a click MEAN?

The ghost is only honest once this is settled, so it is a grill, not a build:

1. **Stack on top** — a click adds a level to the column you clicked, voxel-editor style. Reads naturally for building up; awkward for carving down.
2. **Paint at the brush's level** — the 2D model (#260's wheel sets a level; Z/C set the ramp rise) simply previewed in 3D. Consistent with the existing tool and needs no new concept, but the number lives in a panel rather than under the cursor.
3. **Paint on the FACE you clicked** — Minecraft's rule. `BoardPicker`'s ray already walks columns and knows which cell it hit; whether it can cheaply report which *face* is the open question.

These are genuinely different authoring feels, not implementations of one idea.

## What already exists (verified 2026-08-15, don't re-derive)

- `DevController.cell_source` — the injectable brush-cell source #231 added; the 3D host supplies the picked cell and paint/erase/ghost all inherit it from one place.
- `BoardMirror.show_brush_ghost(cell, ghost_layer, heights)` — the 3D block ghost, height-aware as of #273.
- `BoardPicker.column_tops_from(board)` derives column heights from the GridMap itself, and `battle3d._refresh_tops()` already runs after every sync.
- `BoardSpace.of_cell(cell, level)` / `surface_point(cell, heights)` — the coordinate seams.

## The one known blocker

`DevController.brush_ghost_cell()` **returns `NO_CELL` for every paint mode except TERRAIN** — deliberately, because the ghost is a *tile* preview and elevation places no tile. An elevation ghost in 3D has to get past that gate, and the honest fix is probably to make the ghost answer "what would this click produce" per mode rather than widening a TERRAIN-shaped predicate.

## Done when

[...]

PR #580, opened 2026-08-27
#552 slices 2 + 2b: the shoreline, and the board field under it

### LINKED ISSUE #552: Water is a solid opaque cube with a static texture: the Tier-1 water pass (animation, shoreline, reflection)

Phaazoid:
## #552 slice 2 — the shoreline, and the board mask under it

Foam where water meets land. But the deliverable that matters is underneath it.

### The debt this slice pays

#552 argued against keeping water a MeshLibrary item, and the reason it gave was exactly this slice: *"a foam edge has to know its neighbours, and a meshlib item knows nothing about the cell next to it."* You took the meshlib route anyway. This is where that comes due, and the payment is a **board-sized water mask** — one texel per cell, white where that cell is water, pushed as a global `sampler2D` beside the tuning globals.

**The elegant part is free, and it is why this is a small slice.** Texel centres land on *cell* centres, so `filter_linear` hands over a distance-to-shore ramp with no distance field baked and no neighbour walk: 1.0 well inside the water, exactly 0.5 on the shared edge with land, hardware doing the interpolation. `repeat_disable` clamps, so water running off the board meets its own value there and does not foam against the void. The shader turns a world position into a texel with one `vec4` rect, because `BoardSpace.CELL_SIZE` is 1.0 and a 2D cell's `y` **is** the world `z`.

It rebuilds from **both** sync doors, since only one of them is on the path a brush edit takes.

### Foam

Four knobs, per type per slice 1c's rule, on the Water tab you already have. **The foam colour's alpha is its strength** — a surf hue and how much of it there is are one decision, not two, and the knob panel's colour widget has a real A channel so it is a live dial rather than a hidden one.

Two choices worth your disagreement:

- **It composites AFTER the cell seam**, on purpose. The seam draws every cell boundary, and a grid line running through the surf is what a shore must not have.
- **The band breathes** on the wave field the shader already computes, scaled by the foam's own width — derived rather than given its own dial, for `CAUSTIC_SPEED`'s reason. A static band is a painted outline, not surf. If you want the breathing separately tunable, that is one row.

**Defaults are a guess to react to, and deliberately not matched:** deep 0.3 wide at alpha 0.7, shallow 0.5 at alpha 0.5. A shallow shore laps so it gets the wider softer band; deep water meeting a wall stops dead and gets a narrow hard one. Given your sweep moved away from bright channels everywhere, I would not be surprised if both alphas want to come down.

**Declared 3D-only** (#292, with #285 as the model): the flat view's water is a tile *pick*, and a shoreline is a sub-tile gradient no tile pick can express. Stated in the shader rather than in a tracker.

### What it cost the tests, and one finding that invalidated my own

**The type-ambiguity law from 1c had to grow a category.** The mask names no water type and never can. So `BOARD_GLOBALS` is a declared exemption whose membership is asserted in **both** directions — a hole a future knob could fall into unnamed is that law quietly deleted, while a named list with fixed membership is a category. Mutant 4 below is what proves the difference.

Separately, **three cases each parsed `global uniform` lines for themselves** with the same one-liner, and the sampler's hint suffix (`: filter_linear, repeat_disable`) broke all three at once. They are on one helper now. That was forced by this change, not tidying.

**And the finding, which is worth more than the feature:** headless, `ImageTexture.get_image()` answers the **first** image the texture was ever given and never changes again — a second `set_image`, even at a *different size*, still reads back the original. Measured, not assumed.

That matters because **two of my mask cases were passing for the wrong reason.** Each ran on a fresh mirror, so each read its own first write. The case that actually mattered — *the board changed, did the mask follow* — could not see its own subject, and failed for a reason I initially misdiagnosed as `ImageTexture.update()` being the culprit. The assertions now watch the `Image` at the push funnel, which is the layer where our code makes a decision; the `Image` → texture → uniform hop is engine plumbing with no name to get wrong.

### Falsification — four mutants, two needed isolating

| mutant | verdict |
|---|---|
| the rebuild is dropped from `sync_cells` | red — *"after sync_cells(): cell (6, 3) is WATER on the board but the mask reads it as land"* |
| the mask rect's origin shifts one cell | red — **but only after isolating** |
| a global naming neither type | red — **but only after isolating** |
| `BOARD_GLOBALS` claims a member the shader lacks | red — *"the board-data exemption is declared as [...] and the shader's is [...]"* |

**Both isolations were the 1c trap again, and both times the run itself no-opped.** The rect mutant reddened the *content* case first and only 4 of 7 cases ran; the untyped-global mutant reddened the *spelling* case first and only 6 of 8 ran. In both, the case I was aiming at never executed, and reported naively each would have been a green mutant recorded as a red one. Grepping which case fired is what caught it.

### A neighbour I did NOT touch, because you should decide

**`LookDev.tscn` renders water at `project.godot`'s declared defaults, not yours** — it loads the meshlib but has no `BoardMirror`, and nothing else pushes the globals. So the playground the docs call *"the standing playground where any of it gets prototyped"* is showing shallow caustics at 1.35 where your board runs 0.37, and bed at 0.39 where yours is 1.0.

This predates slice 2 — it has been true since slice 1 — and it is the same shape as any two-spellings-of-one-default problem: `[shader_globals]` values are written once and nothing updates them when *Save to source* rewrites the `@export`. Harmless in the game, because `_push_all_water` overrides at `_ready`. Not harmless in the scene you'd prototype the next water change in. Say the word and I will file it.

### Verification

`law` / `dev` / `presentation` run locally; CI owns the tree. `lookdev_meshlib.tres` is untouched — no generator change and no regeneration, so the diff is smaller than it sounds — and `project.godot` took a clean 24-line hunk with zero churn, which is #576's input-device fix earning itself.

**The blind spot, stated as usual:** no headless test can see shader output. Every law here stops at the push. That foam appears *in the right place on screen* is your check, and the UV derivation it rests on is two lines I can only reason about.

### What this unblocks

The mask is the shared asset three open items wanted: **FROZEN-over-moving-water** (which `presentation-effects.md` already names as needing exactly this — it now needs only a channel), the distance-from-shore idea in #578, and reflection if it ever happens.

---

**Closes #552.** The dev called the ticket here (2026-08-27); the remainder is filed rather than parked — colour blending and deep water's missing layer in #578, the lakebed's structure in #579, FROZEN-over-moving-water in #584, land↔land transitions in #581, and reflection stays out of scope under #455. See the closing ruling on #552 for the full accounting.


Phaazoid:
## Slice 2b — the mask becomes a field (folded into this PR)

Your two complaints turned out to be **the same shape of problem**: a value that could only ever be 0 or 1 where the eye wanted a gradient. Both are fixed by the mask slice 2 shipped, promoted from a bitmap to a small field.

### 1. Rough corners were a GRADIENT artefact, not an art one

Thresholding a *binary* mask through `filter_linear` gives a foam band whose width follows the **interpolation gradient** — steep across a straight edge, shallow near a corner. So an L-shaped patch grew a fat mushy corner while its straight runs stayed crisp. The band wasn't drawn wrong; it was drawn *wider*, exactly where you noticed.

R is now a **signed distance to the shore**, built by a two-sweep chamfer transform in `_rebuild_water_mask` (CPU, once per terrain edit, ~30 lines). A real distance field has a **constant gradient, so the band has a constant width** — that is the whole fix.

**It ships as a visual no-op on straight shores**, deliberately: the encoding puts 0.5 exactly on the cell boundary, so the 0.5 crossing lands where the bitmap put it. Corners are the only place the numbers differ. Your foam knobs also keep their meaning — the shader decodes back into *cells* before thresholding, so width still reads in half-cells.

Second, smaller thing in the same area: **the foam now samples at the art pixel's centre**, so the shoreline stair-steps like the bands and the bed. It was the one smooth anti-aliased curve on a surface quantised to 32 pixels a cell, which is the other half of why corners read soft.

**The honest limit:** two water cells touching only *diagonally* still meet at a point. That is geometry, not encoding, and no amount of field makes a point into an edge.

### 2. The stark depth seam was structural — there was no code path that could blend

`deep` was a **per-material** uniform, baked 0 or 1 by the generator from the tile's `walkable` flag. Two materials, one per type. So all eleven `mix(shallow, deep, deep)` sites snapped at the cell edge and the bed popped from present to absent in one pixel. Nothing could have been tuned to fix that.

G is now deepness in the same mask, the per-material uniform is **deleted**, and **the two water materials collapse to one**. That collapse is the point rather than a side effect: a bake sitting beside a mask channel is two answers to *how deep is this water*, which is the duplicate-seam shape that fails late once both have callers.

At a cell centre G is exactly 0 or 1, so **away from a type seam this reads identically to the baked version** — your tuned values are untouched. At a seam, all eight knob pairs glide across and the bed fades out *into* the deep.

**The subtlety that makes it work at all:** land texels take the deepness of the water beside them (one 8-neighbour dilate, deeper wins a tie). Without it, bilinear from a deep cell's centre reaches into its land neighbours, so a land 0 would drag every cliff rim toward shallow and fade the bed in against the wall. That has its own law, asserted as a *pair* — "land next to water reads deep" would pass on a dilate that just wrote 1 everywhere.

**Residual, stated rather than left to be discovered:** the two tiles' authored `modulate` still steps at the cell edge. That's tile art and no shader can unbake it per fragment. My guess is that once everything around it blends it reads fine; if it doesn't, the follow-up is a depth-tint knob driven by the same channel, and I've deliberately not pre-built it.

### `walkable_of`'s justification got SHARPER, and both docs were wrong

`GridUtils.walkable_of` exists as a split from `BoardContext.is_walkable`, and `terrain.md` and `CLAUDE.md` both justified it as *"the meshlib generator asks the same thing of a TILESET, which has no board"*. **The generator no longer reads it at all** — `BoardMirror`'s mask does.

The reason survives and improves: the render needs the tile **TYPE's** flag precisely because a **FROZEN water cell is walkable while still being deep water to look at**, which `is_walkable` structurally cannot say. Both docs corrected.

### One design note worth your eye

The decode scale (`water_board_shore_range`) is **pushed, not spelled in both languages** — a const in GDScript and another in GLSL is one edit away from two different answers with nothing to catch it. It joins `BOARD_GLOBALS`, which is the closed exemption list slice 2 built. **Widening that list deliberately is the mechanism working as designed rather than around it**, and the law asserts membership in both directions, so this couldn't have happened silently.

### Falsification — three mutants, one truncation caught

| mutant | verdict |
|---|---|
| the distance field degraded back to binary | red — *"a cell two in from the shore reads 1.000000 and one touching it reads 1.000000 — the mask is still a BITMAP"* |
| the dilate dropped | red — *"the land cell beside DEEP water reads shallow — every cliff rim will drag its own water toward shallow"* |
| the generator bakes depth per material again (regenerated to prove it) | red — *"water surfaces wear more than one material"* |

The first truncated the case after it, so I checked case counts rather than trusting the exit code — 8 of 9 ran there, 9 of 9 on the second.

### Verification

`law` and `presentation` locally; `dev` deliberately skipped and the reason written down — no `GameKnobs` row moved, no new knob. I checked that nothing in `tests/` actually opens a docs file before deciding the doc edits carry no test dependency.

The meshlib was regenerated by the tool, never edited: 3 ShaderMaterials → 2, zero `shader_parameter/deep`.

**The blind spot, as always:** no headless test sees shader output. Every law here stops at the mask. **That corners actually look rounded and the depth seam actually reads blended is your check** — ideally on the same L-shaped patch and the same shallow/deep boundary you judged rough.

---

### Why this is in #580 rather than its own PR

Your call, and the right one: you had already played slice 2 and told me what was wrong with it, so there was no reason left for it to land on its own — and you'd otherwise have been feel-checking corners I was about to change.

**One thing that happened while this was in flight, worth knowing.** Your `content authoring` commit on main added slice 2's six `[shader_globals]` entries to `project.godot` — the editor writes them when it loads a project whose shader declares them, so running the branch in the editor and committing picked them up. That made this PR conflict, since both sides added the same block. Resolved by **regenerating rather than hand-merging hunks**: took the superset, ran `ProjectSettings.save()`, and `git diff` came back empty — which is the proof that the resolution is exactly what the writer emits rather than something I stitched together. `project.godot` is down to a **+4 diff** now, since main already carries the rest.

Issue #519, opened 2026-08-25
Battle zoom: dramatic timing profiles - one beat table, CINEMATIC and BOARD


Phaazoid:
Slice 2 of #410 (battle zoom). **Board-only, no diorama, no art.**

## Ask

Make beat duration a first-class, *shared* thing — and fix the plain board playback in the same stroke.

Per the dev's ruling on #410 (2026-08-25): the cinematic toggle picks a **profile**, never "timings vs. no timings". One table maps a beat's **dramatic weight** to a duration; `CINEMATIC` and `BOARD` are two profiles of that one table.

- A `KILLED` blow gets wind-up and hold. An Iron Will survival gets the held breath — hit lands, unit still standing. Chip damage stays fast. The counter phase gets its act break.
- The `BOARD` profile is where *"the simpler battle show, it is lacking right now"* gets fixed. `Pacing`'s flat constants are the degenerate case it replaces — note `PLAYER_ACTION := 0.0`, i.e. no gap at all on the player's own Execute.
- Ships the `PlayerSettings` row (one enum member plus one `DEFS` entry; `SettingsScreen` builds the row itself) and **hold-to-skip**.

## Why the shared table is the whole point

Building durations inside the cinematic and extracting them later is a second answer to *"how long does this beat take"* arriving once both paths have callers — Law #4's late-failure shape. The abstraction is the deliverable, not a follow-up.

## Where it lands

`OrderExecutor.execute_orders` currently spends **one** float across the entire pass: `_execute_action_sequence(plan.attacks, beat)`, then `_apply_cell_effects`, then `_execute_action_sequence(plan.counters, beat)`. The beat sheet (slice 1) owns those durations instead.

Skip is trivially safe — combat is fully resolved before playback, so nothing resolves during it.

## Done when

- Both profiles drive the same table; a test proves a `KILLED` beat and a chip beat differ *under each profile*, measured in a unit the feel values cannot move.
- Toggling the setting off leaves a board pass visibly better paced than today's, not one that reverts to flat.
- Hold-to-skip returns to the board with the pass fully applied.

## Notes

- The white flash on skip must read #217's photosensitivity switch — already a `PlayerSettings` tenant.
- Do not pin feel-tuned durations in a test (project law); pin the *fork*, not one side of it.

Phaazoid:
**Rulings (dev, 2026-08-26).**

[...]

Issue #118, opened 2026-07-30
AI turns resolve too fast to follow: a beat between actions during execution


Phaazoid:
🤖 Claude says:

Filed from `docs/SCRATCHPAD.md` (sweep 2026-07-29). @Phaazoid's words:

> During AI turns, the actions the AI takes need to be more clear. Each action can be queued quickly,
> but there needs to be more of a pause between each action, so the player watching can understand
> what is going on.

Sibling of the battle-log half of the same note, filed as its own issue because this one needs no UI
design at all — the log does.

## Where the beat goes

`OrderExecutor._execute_action_sequence` advances the instant an action reports
`execution_complete`, so attacks and counters resolve back-to-back at animation speed. There is
already one await between actions (`_offer_pending_crisis`), which is the seam a beat belongs beside:

- **between sequential actions** — attacks, then counters, then each side-channel batch;
- **after the parallel move phase** — moves all animate at once by design, but the pass then jumps
  straight into the first attack;
- **not** inside `_execute_action_parallel`'s completion poll, which is a frame-waiter, not a pace.

`AIController` already has one deliberate pause per squad (`await camera_controller.pan_to`), so the
shape and the precedent both exist.

## The two forks worth a decision before a walkthrough

1. **Does the player's own turn get the same beat?** The gap is identical when you hit Execute —
   nothing about `_execute_action_sequence` is AI-specific. Leaning: **one tunable, two values**
   (a longer AI beat, a shorter or zero player beat), since the player already knows what they
   queued and the AI's plan is the thing being read for the first time.
2. **Fixed duration, or per-action-kind?** A move that animates across four tiles already reads;
   an attack that resolves in one frame does not. Leaning: one constant, tuned by feel, because a
   per-kind table is a second thing to keep in sync with the registry for no measured gain.

## Constraint: this must not land on the test suite

`OrderExecutor.execute_orders` is awaited directly by tests, and the suite already runs 800+ cases
at 0.06–0.26 s each (see `tests/README.md` and the timing note in `CLAUDE.md`). A hard-coded
`get_tree().create_timer(0.4)` per action would add real wall clock to every test that resolves a
plan. So the beat needs to be a **tunable that headless and test contexts can set to 0** — a
constant on `OrderExecutor` a test can zero in `before_test`, not a literal at the await site.
`play/play_session.gd` is a deliberate mirror of this path (it resolves synchronously, no awaits,
no prompts) and should stay at 0 unconditionally.

## Law check

[...]

PR #583, opened 2026-08-27
#520 diff 2b slice 2: the linger, and a Playback page you can read

Phaazoid:
**Stacked on #577** (`feat/520-camera-glide`) per your call, so this build has the camera glide AND the linger together. They merge bottom-up: #577 first, then this.

Your three findings. Two of them turned out to be the same bug.

## Nothing ever waited after an action

Every pause in the game landed *before* its action — `_execute_action_sequence`'s own comment said so, and it was right when it was written. Then health readouts gained debris. `AttackAction.execute` does await its lunge, its block, the knockback slide and the plummet — but the cube burst is thrown by `UnitMirror`'s own per-frame HP poll, flies on `HealthBlockDebris`'s own clock, and **nothing anywhere holds a reference to it.** So the last counter killed someone and the pass cut away mid-explosion.

`linger_for` now mirrors `hold_for` clause for clause, with two deliberate differences:

- **Flat**, per your call — no profile, not scaled by drama. A hold is anticipation; a linger is matched to an animation that runs in real time either way.
- **One outcome rung, not five.** What makes a beat longer to *watch* is how much debris it threw, and only a death empties the grid. A shove and a plummet are already awaited inside `execute`.

It fires after each beat's **last** surviving member, where the hold sits on its first — that is "one blast is one moment however many it hits". Keying both ends on the first member puts the pause between the first hit and the second.

## The ladder had no floor

`HOLD_ATTACK`, seeding `hold_for`'s volley branch instead of `0.0`. It was worse than a missing control: a hit that only did damage earned zero, so the commonest beat in the game took the bare base beat — and every other rung was a number with no zero point to be read against.

It is deliberately **not** in `coda_hold`, which answers for side-channel verbs and must go on reporting ATTACK as undeclared. Attack is an entry of the panel's Actions list; in the code its numbers live in the volley branch.

## The page

Six sections on one tab — The profile / Actions / Outcomes / Camera travel / The tear-out / Motion. No new machinery: `GROUP_TABS` already maps several groups onto one tab and `_add_heading` already fires per group.

- **A battle-zoom toggle at the top writes the real `PlayerSettings` value** — the first dev tool to write one — and doubles as the profile filter. So you always tune the mode you are watching, and a slider that would move nothing is not in front of you. That is what those `zoom off / zoom on` pairs really were: one dial measured under each of two live modes, with both branches of the fork shown at once.
- **An Actions dropdown** shows one verb's Hold and Linger at a time, attack included. A verb that grows a third dial costs no page.
- **The base beat stays in The profile**, per your ruling — it is the floor every other slider executes against, so it is tied to no action. It is the ragged cell: zoom-off forks again on whose pass it is, zoom-on does not.

## Flagged for your play-check

- **Zoom-off is now slower than it was.** The linger is flat, so it applies fully in both profiles, and zoom-off had been bit-for-bit the pre-#519 pacing. That was the plan's stated riskiest assumption. If the plain board reads as sluggish, the fork is one line plus a second column of knobs — and the toggle now makes judging it a two-second A/B.
- **The move phase does not linger**, deliberately: nothing bursts on arrival, and the parallel phase has no per-action tail to hang it from. One row if you want it.
- **A watch shot that downs someone lingers as a plain hit** — declared cut. Watch shots have no beat to read a rung off, and #567 is rewriting that playback anyway.
- Defaults are guesses: `HOLD_ATTACK` 0.25, `LINGER_ATTACK` 0.45, `LINGER_DOWN` 1.0. All on the page.

## Verification

Green locally across eight areas: `core law dev squad` 843/843, `presentation ai flow ui` 1014/1014. CI has the full tree.

**Mutants: six run, five killed at the case they aimed at.**

- linger keyed on the volley's first member -> "the linger sits on the volley's OPENING action"
- `hold_for` seeded from 0.0 again -> "the ladder has no floor"
- the filter *skips building* instead of hiding -> "has no row in the panel at all -- the filter is not building it"
- `coda_linger` loses its -1.0 sentinel -> caught by the case, while the registry law stayed **green** (0 passes `>= 0`), which is why the two are not redundant
- a linger row loses its `action` tag -> "BURROW has no LINGER row tagged to it on the Game tab"

**The sixth survived, and I said in the plan that it would: deleting the `await` on the linger leaves 86 cases across four suites green.** `Pacing.beat` returns without awaiting headlessly — the escape that keeps every resolve-pass test off the wall clock — so the pause is structurally invisible to the suite. The call site says so in as many words, the same declaration `clear_guard_preview` carries three phases up. What is pinned is the table and the schedule; **the pause itself is yours to judge by playing.**

**One test rewritten, not patched.** `test_a_real_felling_blow_reaches_the_beat_table_as_one` pinned "a scratch earns exactly 0" — the rule `HOLD_ATTACK` changes. Its fear was real (its own message: "every beat now reads as a big moment"), so it now asserts the relationship: a kill outlasts a scratch, and a scratch is no longer bare.

**One hand-copy derived away.** `test_beat_timing`'s Pacing snapshot was an 18-entry hand list and this ticket would have made it 30 — the shape #450 already fixed in `test_game_knobs`, where a missed entry poisons every later case in the RUN. It now walks the table through `read_static`/`write_static`, with a guard case proving the derivation covers something.

**Swept two measured figures that had drifted** and are about exactly what this diff changes: `execute_orders` is awaited by 34 call sites across 12 suites (the docs said nine, from #118), and `Pacing` is forty statics (they said five). Both in `CLAUDE.md`, `visual-clarity.md` and `test_pacing.gd`'s header.

**No embedded content** carries any of these types, so the `.tres` staleness class does not apply here.

Canon: `docs/design/visual-clarity.md` -> *The camera STAYS*. Its #118 section now records that "the beat lands BEFORE each action, never after" was repealed, annotated rather than rewritten since the rest of that bullet still stands.


Phaazoid:
Superseded by #587 — this PR was closed automatically when #577's merge deleted the branch it was based on (`feat/520-camera-glide`), rather than being retargeted to `main`. No work was lost; #587 is the same commits rebased onto main.

Issue #582, opened 2026-08-27
A cell painted more than 1 unit below flat can never be picked again: the hole plane occludes it from in front


Phaazoid:
🤖 Claude says:

**A terrain cell painted more than ~1 unit below flat can never be picked again.** Not hard to hit, not recoverable with the tile brush, and the erase gesture cannot undo it either — every dev gesture funnels through the same pick, so the cell is simultaneously unpaintable, unerasable and unselectable.

Reported by the dev while authoring `Scenarios/missions/Level_2`, with two in-game reports 5 seconds apart: `user://reports/2026-08-27_11-32-21/` (the cell alone) and `.../11-32-26/` (after painting a ring of neighbours around it to try to reach it). Build `0.103.1`, `main @ bb31215`.

> due to how the tile brush selects at depth, I am actually unable to paint, delete, or even select a spot on the board. I think this is caused from me authoring low blocks.

He is right, and it is worse than one spot: at the nominal camera pitch **nothing 2 or more units below flat is ever pickable.**

## The board in the reports

`board.tres` in both: a 12×20 slab at height 0 (`x 0..11`, `y 0..19`), and exactly one cell off it —

```
corner_heights = Dictionary[Vector2i, Vector4i]({
Vector2i(-2, 10): Vector4i(-3, -3, -3, -3)
})
```

**−3 units = 1.5 levels below flat.** The second report adds six cells at height 0 — `(-3,9) (-3,10) (-3,11) (-2,9) (-2,11) (-1,10)` — the ring he painted around it. Nothing else changed. It did not help, and the measurement below shows it structurally cannot.

## The mechanism

`BoardPicker.pick_cell` walks columns in ray order and stops at the first one the ray dips into. `_top_row` decides what each column offers:

```gdscript
static func _top_row(column: Vector2i, tops: Dictionary[Vector2i, int], plane: Rect2i) -> int:
	if tops.has(column):
		return tops[column]
	if plane.has_point(column):
		return BoardSpace.FLAT_TOP_ROW      # 2 -> surface at y = 1.0
	return NO_COLUMN
```

That middle branch is #231's hole plane, and it is **opaque and flat at height 0** across the whole board plus an 8-cell apron. `battle3d._paint_plane()` returns `_board_rect.grow(paint_apron_cells)`, and `_board_rect` already contains the sunken cell — so a sunken cell's neighbours are *always* inside the plane. There is no board layout that gets you out of this.

[...]

PR #595, opened 2026-08-27
The player can tilt the camera (#586)

### LINKED ISSUE #586: The player can tilt the camera up and down
🤖 Claude says:

> **RE-SCOPED 2026-08-27 by dev ruling — this is a PLAYER control now, not a dev affordance.** The two earlier framings (a dev tilt; then a park-and-play-check) are both superseded. History kept in the comments.

**The player can tilt the camera up and down by dragging, the same gesture that already orbits it.** Grown out of [#582](https://github.com/Phaazoid/Godoiosis/issues/582), where a one-cell well proved unreachable at the board's fixed −40°.

## The ask

> I think it's better to just let the player in general tilt the camera up and down too. We have a reset button, and Q/E act as soft resets as well. I don't see the harm in letting the player drag up/down too.

The orbit drag already consumed `relative.x` for yaw and threw `relative.y` on the floor. That axis becomes the tilt.

## Why it can't just write the Pitch node

Board pitch is authored as `CameraRig/Pitch.rotation_degrees:x` and addressed **directly** by a Look knob (Moods → *Camera framing* → **Board pitch**, −85° to −10°). Once the player can tilt, `_process` would write that property every frame — and a knob naming a property the game overwrites per frame is exactly the born-dead slider `tests/dev/test_moods_tool.gd` exists to refuse.

So the node's rotation becomes **derived**, the shape [#520](https://github.com/Phaazoid/Godoiosis/issues/520) already drew for position:

- **`board_pitch_degrees`** — the authored baseline, and the Look knob's new target. Snaps rather than eases, because pitch is part of the camera basis: a rig still lerping toward a freshly applied mood would unproject at one angle and pick at another.
- **`_pitch_degrees`** — the live angle, the rig's own float, eased in `_process` as a third channel beside yaw and distance.

## Rulings

- **R levels the tilt; Q/E does not** (dev, 2026-08-27 — see comment). Tilt-and-turn is the gesture the tilt exists for.
- **The clamp band is a Game knob, not a mood.** A mission may not re-teach the controls. Shipped at −80/−20 as a starting guess.
- **One sensitivity.** The drag reuses `orbit_sensitivity` rather than inventing a second rate.
- **The fit is not re-solved mid-drag.** That staleness already exists and the Moods tooltip declares it; re-framing the board under someone's hand is worse, and tilting steeper narrows the ground covered, so the stale ceiling errs permissive.

## Measurements that motivated it

A one-cell well 3 units deep with ground around it, 69120 rays, share of aim points resolving to the sunken cell (25% would be the whole cell):

| pitch | share |
|---|---|
| 25–45° | 0.0% |
| 50° | 0.08% |
| 55° | 0.7% |
| 60° | 3.6% |
| 70° | 9.9% |

A usable look into a one-cell well needs roughly **70°+** — inside the shipped band.

## Note on the art

**Sprites are billboards.** At 75–85° they are seen from nearly overhead, which is not the angle HD-2D art is drawn for. The −80° steep limit is a knob precisely so the answer can move once it has been played.

Built in [#595](https://github.com/Phaazoid/Godoiosis/pull/595).

Phaazoid:
Closes #586 — reshaped from the dev tilt I originally filed into the player control you asked for:

> I think it's better to just let the player in general tilt the camera up and down too. We have a reset button, and Q/E act as soft resets as well. I don't see the harm in letting the player drag up/down too.

The orbit drag already read `relative.x` for yaw and threw `relative.y` away. It's the tilt now.

## The shape, and why it can't be simpler

Pitch was authored as `CameraRig/Pitch.rotation_degrees:x`, addressed **directly** by a Look knob. Once the player can tilt, `_process` writes that property every frame — and a knob naming a property the game writes back per frame is the slider that moves and silently reverts, the one failure `tests/dev/test_moods_tool.gd` exists to refuse.

So the node's rotation became **derived**, the shape #520 drew for `position`:

- **`board_pitch_degrees`** on the rig — the authored baseline, and the Look knob's new target. Its setter re-seeds the live angle and **snaps** it (`frame()`'s "hold_at, not glide_to" reasoning: pitch is part of the camera basis, so a rig still lerping toward a newly applied mood unprojects at one angle and picks at another). A mood arriving is a board arriving.
- **`_pitch_degrees`** — the live angle, the rig's own float, eased in `_process` as a third channel beside yaw and distance.

**R levels the tilt**, to the *board's* angle rather than a `_home_` of pitch's own — the mood already says what the opening angle is, so a second copy could only drift. **Q/E does not**, and neither does the playback square-up: tilting to read a pit and then turning to see its other side is the gesture the tilt exists for (your call). The camera return carries it back with the yaw and the zoom.

How far a hand may go is **two Game-tab knobs** (`Tilt limit: shallow` / `Tilt limit: steep`), not a mood — a mission may not re-teach the controls. Both are feel values; the shipped −80/−20 is a starting guess, and the drag reuses `orbit_sensitivity` rather than inventing a second rate.

**13 preset files** re-keyed `CameraRig/Pitch|rotation_degrees:x` → `CameraRig|board_pitch_degrees`, values untouched. Swept and stated per the embedded-content rule; a miss would be loud (`apply_preset` reports unresolved keys, and `test_look_presets.gd` is a content law over the shipped set).

## The bug this shipped with, and what is NOT known about it

The area run went red on **two health-readout suites** — on a branch touching nothing but the camera, one of them reporting a wrong **cube count**. Attributed before diagnosing: fails alone as well as in company, and a clean-`main` worktree runs the same two suites 36/36 green. Mine.

Measured cleanly: reading the live angle **back off the Pitch node** reds them; the rig's own float greens them. Verified both ways. The read-back round-trips through the basis, which returns `-39.999992` for an authored `-40` — the number every LookPreset records — so the ease chases a target it cannot express.

**The mechanism beyond that is unconfirmed, and the comment in the code says so.** The obvious story is `battle3d._poll_pointer`, which skips its work only while `_camera.global_transform` compares equal — but I wrote two cases to pin that (a settling transform, and a synthetic pointer surviving) and **both passed against the very mutant they were aimed at**, in two different scenes. Both were deleted rather than left advertising coverage that isn't there; the guard is the two health-readout suites, named at the source.

## Verification

`tests/presentation` + `tests/dev` — **837/837, 0 failures, 0 orphans** before the rebase.

**After rebasing onto `d017a51` the area is red, and none of it is this branch.** Five suites fail: `test_board_lint`, `test_input_bridge` (two cases), `test_mod_content`, `test_mod_fitting`. Attributed the same way as above — a detached worktree at `origin/main`, the same five suites, produces **the identical verdict**: 1 error, 5 failures, matching case for case and statistic for statistic. Diffing the two runs' failure output gives no differences at all.

The cause is content, not code. `356a2e7 content authoring` deleted `Resources/WeaponMods/Super Scope.tres` (renamed to `Line Sniper.tres`) and `Resources/Weapons/WeaponVariants/brrr.tres` — but `Super Gun.tres` and `Noemie.tres` still carry `ext_resource` lines pointing at both deleted files. Those two fail to load, `Level_1.tres` fails with them, and four suites fall over downstream. Filed separately; which weapon those references should point at now is a content call, not mine to guess.

`test_look_presets.gd` — the one suite the 13-file re-key could plausibly have broken — is **4/4 green on both sides**.

Falsified, one line at a time, each killed by the case named for it:

1. clamp widened to nothing → `test_the_tilt_is_clamped_to_its_own_band`
2. R stops levelling → `test_r_levels_the_tilt_back_to_the_boards_own_angle`
3. Q levels the pitch → `test_a_realign_keeps_the_tilt` (*"a realign is yaw only, or you cannot turn while looking down"*)
4. authored pitch stops re-seeding → `test_the_authored_pitch_snaps_rather_than_easing_in`

**A correction to my own plan:** I claimed the Moods no-placebo law would cover the re-keyed knob without new cases. It does not — measured, 26/26 green with the setter gutted. That law catches a property the game *overwrites*; it is blind to a knob wired to a plain `@export` that nothing reads. Since the fix for a per-frame-written knob **is** to move it onto an export, the move that satisfies one failure mode walks into the other. Mutant 4's kill is my own case instead.

## Feel

The suite can't judge this one. Drag rate, direction (grab-the-world: dragging down tilts down) and where the clamps sit are all things only playing decides — `orbit_sensitivity` and the two tilt-limit knobs are where the answers live.


Phaazoid:
🤖 Claude says:

**Two notes on this branch.**

**1. CI here will be red, and none of it is this diff.** It is [#596](https://github.com/Phaazoid/Godoiosis/issues/596) — `356a2e7 content authoring` deleted `WeaponMods/Super Scope.tres` and `WeaponVariants/brrr.tres` without repointing the four files that reference them, so `Noemie.tres` fails to load and `Level_1.tres` fails with it. `main` is already red at `eb1d04f` for the same reason, before this branch existed. Attributed by running the same five suites on a detached worktree at `origin/main`: identical verdict, case for case, and diffing the two runs' failure output gives no differences at all. Fixing it is a content call, so it is filed rather than patched.

**2. Pushed a second commit — the canon sweep I had missed** before opening this. Worth stating plainly because I opened the PR without it:

- `docs/SHORTCUTS.md` still described the right-drag as **orbit only**. That is the readout going stale against the control — precisely the fault I had updated the in-game help strip to avoid, missed in the doc that documents controls for real. Q/E and R now carry the ruling too.
- `docs/design/verticality.md` asserted a **fixed 40° pitch twice**, and one was load-bearing: the #582 verdict that a one-cell well is *honestly* occluded — *"it needs 56° to see the floor"* — and #586 hands the player exactly that. Said so, and said why it does **not** retire the brush-plane answer: the brush needs no gesture and works at whatever angle you are already holding, so it stays the fix and the tilt is a slower second road.
- The trail-quad passage leans on the same constant to argue one quad never goes edge-on. A vertical quad viewed from near-overhead approaches edge-on by construction, so the steep clamp bears on it — **flagged as conditional and explicitly unmeasured** rather than rewritten, since I have not measured it.

`verticality.md`'s own `Canon checked through` marker is **deliberately not bumped**: I re-verified its pitch claims, not the whole doc, and a marker claiming more than was checked is worse than one sitting behind.

Issue #590, opened 2026-08-27
An authored can_overwatch attack never offers its watch: the row is deduped away by its own firing twin


Phaazoid:
An attack authored with `can_overwatch` in the Attack Editor never offers its watch. The row is
built correctly and then thrown away, silently, by the menu's own duplicate-name guard.

## Repro

1. Attack Editor: new weapon attack, tick **Can overwatch**. (Test fixture: `Line Snipe`, a
   `ForwardLinePattern` of length 5.)
2. Grant it to a weapon — a mod's `granted_attacks`, or a template's `extra_attacks`.
3. Equip that weapon and open the kit slice.

**Expected:** a row that takes you into overwatch targeting.
**Actual:** exactly one row under that attack's name, and it *fires* the attack.

## Why

`ATTACK`, `WEAPON_ACTION` and `TRANSMUTATION` all sit in `Group.ATTACK_GROUP`
(`Classes/ui/MainActionMenu.gd:113-115`), so the whole kit slice is **one bucket** and every child
passes through `_append_unique` (`MainActionMenu.gd:206`), which drops any node whose `name` is
already present. First listing wins.

A watchable non-main attack is offered by two different builders inside `_weapon_children`, under
the same name:

| # | row | built by | outcome |
|---|-----|----------|---------|
| 1 | `Shot` | ATTACK expander (`effective_main`) | kept |
| 2 | `Line Snipe` | `get_weapon_secondary_attacks()` — **fires** it | kept |
| 3 | `Overwatch` | `_overwatch_rows()` — watches with `Shot` | kept |
| 4 | `Line Snipe` | `_overwatch_rows()` — **watches** with `Line Snipe` | **dropped** |
| 5 | `Reload` | self-verb | kept |

Row 4 is fully constructed — `can_overwatch` is read, `Unit.attack_can_overwatch` returns true, the
leaf and its `_pick_watch` callable are made — and then discarded because row 2 already claimed the
name. The label rule is `MainActionMenu.gd:307`:

```gdscript
var label: String = "Overwatch" if atk == unit.get_default_attack() else atk.display_name
```

A non-main watchable attack is labelled with the exact name its own fire row already uses.

The carbine's stock overwatch is unaffected because its watchable attack **is** the main: `Shot`
lives in the ATTACK expander, its watch row is special-cased to the literal label `Overwatch`, and
nothing collides. That is the whole of the difference between "the hardcoded one works" and "the
authored one does not".

## Already ruled

[...]

PR #600, opened 2026-08-28
#592: draw the watch annotation, instead of writing it into a hidden label

### LINKED ISSUE #592: A move that walks into a watch shows no annotation in the action queue
A unit walking through a watched cell takes the shot, and the action queue never says so.

> Also, when a unit moves through an overwatch tile right now, the overwatch hitting them doesn't
> show up in the action queue, so that needs an issue as well.
>
> — dev, 2026-08-27

## This is spec'd, and the machinery exists

`docs/design/standing-reactions.md` (queue representation): the annotation belongs on the crossing
move's own row — *"triggers Overwatch — takes 12 from Carbine (E)"* — stacking when one route
crosses several watches. A deliberate divergence from how counters display, *because
[#412](https://github.com/Phaazoid/Godoiosis/issues/412)'s whole payoff is dragging a move row and
watching who eats the shot change.*

It is built. Whoever picks this up should start from "find where it stops", not "build it":

- `ActionQueueDisplayEntry._watch_notes` (`Classes/ui/queue/ActionQueueDisplayEntry.gd:96`) writes
  exactly those strings, in all three victim cases.
- The MOVE section is the only one that passes `plan` into `_add_section` (line 61) — every other
  section takes the `null` default, so only move rows can carry notes. That matches the spec.
- `plan.watches` is gathered from **every** unit on the board, both factions
  (`SquadManager.gd:614-619`), so a player mover crossing an enemy watch is in scope.
- `PlanResolver` stamps `shot.triggered_by = entrant` (line 247) and appends to `plan.watch_shots`.

## Where it can break

The filter is one line (`ActionQueueDisplayEntry.gd:101`):

```gdscript
if shot.triggered_by != action.actor or shot.resolved == null or shot.resolved.skipped:
```

Candidates, in the order worth checking:

1. **`resolved` is null or skipped** on the derived shot at the moment the panel reads the plan —
   the panel and the resolve may not be looking at the same pass.
2. **The crosser has no MOVE row.** A unit *shoved* into a watch triggers it (intended design — the
   mace combo), but it never authored a move, so `triggered_by` matches no row's actor and the note
   has nowhere to land. Structural, not a wiring slip.
3. **The enemy-turn case.** If the observation was an AI unit crossing the player's watch, the
   panel is showing a different plan (or none). That is the pure-feedback half; the
   player-plans-into-an-enemy-watch half is the one that costs the player a decision.

## Priority note

Filed P2. **It should rise to P1 if case 3 is not the one that is broken** — cases 1 and 2 mean the
player cannot see a shot they are about to walk into while they still have the ability to re-order
around it, which is the lever #412 exists to give them. Case 3 alone is feedback after the fact.

Related: [#413](https://github.com/Phaazoid/Godoiosis/issues/413),
[#567](https://github.com/Phaazoid/Godoiosis/issues/567) (the shot plays after the walk instead of
interrupting it — same moment, different surface).

Phaazoid:
Closes [#592](https://github.com/Phaazoid/Godoiosis/issues/592).

A queued walk that crosses an armed watch takes the shot, and the queue never said so. **Every hop
was correct**: the resolver fires the watch, `build_for` makes the note, the panel hands it to the
row — and the last hop appended it to `ActionQueueRow`'s `description_label`, which has been
`visible = false` in `ActionQueueRow.tscn` since `572fa2e`, the panel's first version. The row is
icon-only by design, so #413 wrote its feedback where nobody can see it, and it has been invisible
for the mechanic's entire life.

The dev's bug report is what made it tractable, and the split in it is the tell: `report.md`
projects the **entries**, not the label, so the report showed the note while the screen did not.

```
**MOVE**
- Soldier10 moves to (-2, -7)
  - triggers Soldier9's watch — takes 14 from Overwatch
```

## The fix

Each row now sits in a `VBoxContainer` **column** inside its wrapper; an annotation becomes its own
visible `Label` under the row. Every other row stays icon-only; the crossing row grows one line.

**Wording compacted**, because the note is drawn now and the panel is ~128px: the old
`"triggers X's watch — takes N from Y"` wrapped to four lines per row. The **attack's own name** is
what went — since #590 a watch attack is watch-only, so naming the watcher already says which watch
fired, while who eats it and for how much is the thing the player re-orders around. One wording, not
two, because `report.md` is a projection of what the panel draws.

| case | note |
|---|---|
| crosser hit | `-14 from Soldier9's watch` |
| ally splashed | `Soldier7: -14 from Soldier9's watch` |
| nobody left in the footprint | `Soldier9's watch fires` |

Size and tint are `GameKnobs.CLASS_KNOBS` rows on the Game tab — the mission clock's shape, for its
structural reason (`KNOBS` resolves against the Battle3D host and cannot reach 2D UI), with the
same re-apply, since the panel repaints only when the plan changes and nothing is changing while a
slider is being dragged.

## What the column broke, and how it was caught

Three fixed-depth node walks assumed `wrapper → row`. **One was production**: the drag logic mapped
a wrapper to its row (`_row_in`) and a row to its wrapper by counting hops, so with a column between
them **dragging a queue row silently stopped reordering anything** — every cast still returned
something, so it failed as "the queue is not draggable any more" rather than as an error. Both now
search rather than count: `_row_in` down, `_wrapper_of` up. `tests/ui/test_queue_row_drag.gd` caught
it, which is the argument for that suite existing.

## Verification

`ui squad dev` — **873/873**, 0 orphans. Full tree is CI's.

**Falsified**, and the result is the point of the whole session. Mutant: `line.visible = false` on
the new annotation label — i.e. #592 in its original shape.

- `tests/ui/test_watch_note_reaches_the_panel.gd` **reds**: *"nothing VISIBLE on the panel mentions
  the watch… Visible text: ["MOVE"]"*
- `tests/squad/test_watch_queue_annotation.gd` stays **green**

That split is exactly why this shipped. My first version of these suites asserted on
`description_label.text`, which was correctly populated the whole time — they passed against the
live bug in the real game scene, through the real declare→execute→turn-handoff path. The assertion
that has teeth is **visibility walked up the whole ancestor chain**, not presence.

## Canon

`standing-reactions.md`'s queue-representation bullet carried the original wording and said nothing
about where the note is drawn; it now records the compact phrasing, the visible label, and why the
note was invisible for #413's whole life. `CLAUDE.md`'s debt list had #592 as open. `weapons.md`
describes the design choice (an annotation on the crossing row), which is still true — untouched.

## Note for review

`DescriptionLabel` is left in the scene, still hidden. It has been dead since v1 and removing it is
a separate question — `setup()` still writes the row's description into it, and the hover readout
may want that text.


Phaazoid:
🤖 Claude says:

Reworked after the feel-check. The first version drew the note as its own label under the move row, and it was near-invisible against the board — legible only where it happened to cross the panel's black edge.

> those colors clash. They need to stand out... I'd like them to mirror the general action queue setup by just being another action queue row with the firing unit, the weapon icon, and the unit getting hit.

That is better and it is **less** code. A watch shot IS an `AttackAction`, so it already answers every question a row asks — `get_actor_texture` (the watcher), `get_action_icon` (the attack, wearing its lethality rung), `get_target_texture` (the crosser) — and a real row inherits the panel's own styling, so the contrast problem dissolves rather than being tuned to.

`build_for` now emits each shot as its own ACTION entry at `indent_level 1`, right after the move that took it, and the panel draws it through the same `_make_row(list, action, 1, false)` an expanded volley's hits already use — matching the AoE style, as asked. A shot that splashes draws one row per victim, exactly as a volley does.

It is not an order anybody gave, so `AttackAction.is_reorderable` is false for a watch shot and the row's X is inert — the rule counters and hold-position fillers already follow.

**Deleted with it:** the annotation Label, `ActionQueueDisplayEntry.annotations`, `ActionQueueRow.add_annotations`, the two `ANNOTATION_*` statics and their GameKnobs rows, and the wrapper column those needed — which takes the drag-path repairs with them, since the node structure is back to `wrapper -> row`. `report.md` prints the nested row through the ACTION arm it already had, indented by `indent_level`.

**Verified:** `ui squad dev` 875/875 before the rebase; the four affected suites re-run green on the replayed history. Falsified by building the row and hiding it — the real-scene suite reds with *"no VISIBLE row on the panel stands for the watch shot"*, and the data-layer suite stays green, which is the same split that let the original bug through.

Note on history: I rebased onto the new main and then replayed onto the pushed commits instead of force-pushing, since you have this branch checked out — `git pull` will fast-forward cleanly.

Issue #622, opened 2026-08-28
A shared test fixture that resets through the game's own door, and a law that keeps it honest


c3potheds:
🤖 Claude says:

Follow-up to #619. That ticket says *what* is slow; this one is about the design that would make the fix safe, and about a payoff bigger than the wall clock: a shared-scene fixture resets through `clear_board`/`apply_scenario`, so it would continuously audit the determinism contract those doors encode — the same one [#70](https://github.com/Phaazoid/Godoiosis/issues/70) will need when mission-to-mission transitions ship.

## Measured: where a fixture rebuild goes

Probe against `Battle3D.tscn`, six rounds, steady state **214.5 ms** per rebuild (local machine — treat absolutes as relative, the shape is what transfers):

| phase | ms | share |
|---|---|---|
| `load()` the `.tscn` | 0.0 | **0%** — Godot caches it; the 1.32s is paid once per PROCESS |
| `instantiate()` | 2.6 | 1.2% |
| **`add_child` + one frame (`_ready` cascade)** | **125.8** | **58.7%** |
| `load_mission(PROLOG)` | 60.6 | 28.3% |
| `free()` | 25.4 | 11.8% |

The intuitive suspect — reading the scene off disk — is exactly zero. The cost is the whole `Battle3D` tree *entering* the tree: the 3D world plus the entire 2D game inside a SubViewport, per case. `Battle3D._ready` is almost purely wiring (node lookups, signal connections, `attach_3d_host`, pushing refs into the mirrors); nothing in it depends on which mission is loaded, which is what makes it a share candidate at all.

Collapsing ~760 rebuilds to one per suite is worth roughly **330s, ~58% of CI** — and unlike sharding it also shortens the local run the dev actually sits through.

## The reset door already exists, and is lossless

`apply_scenario` **begins with `clear_board()`**, and `capture_scenario`/`apply_scenario` is a complete in-memory round trip that #87 built precisely so a board could survive one without touching disk — battle-scoped state included (weapon readiness, element states, lifecycle, Crisis flags, stat-effect countdowns).

So the per-case reset is not something to invent:

```gdscript
func before():        # once per suite
    build Battle3D; load the mission; _pristine = scenario_manager.capture_scenario("fixture")
func before_test():   # per case
    scenario_manager.apply_scenario(_pristine)
```

## The leak surface is much smaller than it looks

131 `static var` across `Classes/` + `Scenes/`, but they are three different things and only one is dangerous:

[...]

Issue #619, opened 2026-08-28
CI wall clock is 4.6x in twelve days and approaching the 15-minute timeout


c3potheds:
🤖 Claude says:

CI wall clock has grown **4.6x in twelve days** while the suite grew 1.8x, and `tests.yml` sets `timeout-minutes: 15`. The slowest run in the last 25 was **11.5 min**. On the current trajectory CI starts hard-failing on the timeout within a week or two — and it will present as a flaky red on whichever PR happens to be in flight, not as a performance problem.

## The trend

| date | cases | CI job | s/case |
|---|---|---|---|
| 2026-07-27 | 610 | 34s | ~0.04 |
| 2026-08-15 | 1493 | 127s (the figure in CLAUDE.md) | 0.085 |
| 2026-08-22 | 2102 | ~5.0 min | ~0.14 |
| 2026-08-27 | 2668 | ~9.8 min | **0.219** |

The 127s was almost certainly correct when measured — it sits on the trend line. It is stale, not wrong. (GitHub only retains runs back to 2026-08-21, so the first two rows are the recorded figures rather than ones I could re-measure.)

## It is not that tests got slower — the MIX changed

Per-suite wall clock, parsed from the run-33148201725 log timestamps (267 suites, 572s, which matches gdUnit4's own `9min 31s` and the step's 585s):

| area | cases | wall | share | s/case |
|---|---|---|---|---|
| **tests/presentation** | 497 | **348.6s** | **60.9%** | **0.701** |
| **tests/dev** | 393 | **112.0s** | **19.6%** | 0.285 |
| tests/ui | 268 | 43.2s | 7.6% | 0.161 |
| tests/flow | 235 | 30.7s | 5.4% | 0.131 |
| tests/squad | 228 | 11.7s | 2.1% | 0.051 |
| tests/rules | 60 | 1.3s | 0.2% | **0.022** |

**Two folders are 80% of CI.** `presentation` is 18.6% of the cases and 61% of the clock, at **32x** the per-case cost of `rules`.

The cause is per-case fixture cost, not case count. `tests/presentation/test_board_mirror.gd` is 69s over 73 cases (~0.95s each), and its `before_test` instantiates a **whole 3D scene per case**:

```gdscript
func before_test() -> void:
	PlayerSettings.reset_for_test()
	var packed := load(SCENE_PATH) as PackedScene
	_scene = packed.instantiate() as Node3D
	...
	get_tree().root.add_child(_scene)
[...]

Issue #596, opened 2026-08-27
main is red: a content rename left four dangling .tres references, and Level_1 will not load


Phaazoid:
🤖 Claude says:

**`main` is red, and `Level_1` does not load.** A content rename deleted two `.tres` files without sweeping the four files that reference them.

## What happened

`356a2e7 content authoring` was a rename pass — three deletions, two additions:

| deleted | added |
|---|---|
| `Resources/WeaponMods/Super Scope.tres` | `Resources/WeaponMods/Line Sniper.tres` |
| `Resources/Weapons/WeaponVariants/brrr.tres` | `Resources/Weapons/WeaponVariants/Super Scoper.tres` |
| `Resources/Weapons/WeaponVariants/BOWW.tres` | — |

`BOWW.tres` had no referrers, so that deletion was clean. The other two still have live `ext_resource` lines pointing at them:

- `Resources/Weapons/WeaponVariants/Super Gun.tres` → `WeaponMods/Super Scope.tres`
- `Resources/Units/Noemie.tres` → `WeaponVariants/brrr.tres`
- `Scenarios/fixture/verticality.tres` → `WeaponVariants/brrr.tres`
- `Scenarios/missions/2.tres` → `WeaponVariants/brrr.tres`

A repo-wide sweep of every `ext_resource` path under `Resources/{Weapons/WeaponVariants,WeaponMods,Units}` finds **exactly these two dangling targets and no others**, so the blast radius is closed.

## What it breaks

`Noemie.tres` fails to load, and `Scenarios/missions/Level_1.tres` references Noemie, so:

```
ERROR: Could not load scenario at res://Scenarios/missions/Level_1.tres
   at: load_scenario (res://Classes/flow/ScenarioManager.gd:224)
```

Downstream, five test cases across four suites fail on clean `main`:

| suite | case |
|---|---|
| `tests/dev/test_board_lint.gd` | `test_loading_a_board_clears_a_report_about_the_old_one` |
| `tests/dev/test_mod_content.gd` | `test_a_mod_whose_file_writes_a_scaling_change_loads_it` |
| `tests/dev/test_mod_fitting.gd` | `test_the_picker_offers_nothing_the_family_refuses` (+ 1 error) |
| `tests/presentation/test_input_bridge.gd` | `test_a_board_swap_rebuilds_the_mirror_through_the_load_funnel` (+ 1 more) |

[...]

PR #623, opened 2026-08-28
#622: a law that makes board-scoped state declare itself

c3potheds:
First Mikado leaf of #622. No behaviour change — one new law test.

A shared test fixture (one `Battle3D` per suite instead of one per case, worth **~58% of CI** per #619) is only safe if no static quietly holds state that should die with the board. This makes that checkable, and it produces the classification everything downstream derives from.

## Why a law rather than a comment

`clear_board()` is the reset door and it is correct — but it is **hand-maintained**, and every line in it cites the leak it was added for: the Prolog accident, a sandbox inheriting the last mission's look (#253), camera start (#234), AI factions (#150), a tear-out left in the sky (#521), the playback lock (#484). Each of those comments records someone not remembering. Nothing makes the next person remember.

A `static var` in `Classes/board` + `Classes/presentation` must now be one of three things:

- **TUNING** — `static var` rather than `const` exactly so a knob can write it
- **CACHE** — derived and idempotent
- **BOARD-SCOPED** — must die with the board, and `clear_board` must be *seen* to clear it

Anything in none of the three fails, naming itself and what to do.

## The encouraging measurement

**28 statics in that domain; the knob table already covers 19 for free.** The leak class is *one concept* — `BoardSpace`'s staging trio, which `clear_board` already clears and which is static only because the mirror and the camera rig read it with no board reference to hand. The invariant we want is nearly true already; this stops it drifting back.

The tuning arm is **derived** from `GameKnobs.CLASS_KNOBS`, the same table `tests/dev/test_game_knobs.gd` already walks to snapshot those statics — Law #4, one enumeration, two readers, no drift.

## Two real gaps it surfaced

`OverlayManager.GUARD_RING_SCALE` and `UnitSprite3D.texels_per_unit` are `static var` for a knob's sake and **no knob ever arrived**, so nothing can move them — the inverse of #264's born-dead slider. Harmless to sharing, so they sit in `PROCESS_SCOPED` with a note rather than reddening a law about something else. Worth their own ticket against the 2026-08-12 knob rule.

## Verification

- `tests/law` — **189/189, 0 failures, 0 orphans**.
- **Falsified twice.** Adding an undeclared static reds case 1 and names it (`_fake_board_leak (OverlayManager.gd)`). And the non-vacuity case earned its keep on its *first* run: a floor I had miscalibrated from a wider grep revealed the other two cases passing over **zero** statics — precisely the failure it exists to catch, caught before the law shipped.

## Note for whoever reviews

While running this I hit a red `test_resource_uid_references` locally that is **not** a bug on main: switching to a branch cut from an older `main` and running `--import` there unregisters uids for files that branch lacks, and the stale registry survives switching back. A fresh `--import` on the current branch clears it. Worth knowing because it is the mirror image of #596's warning — there a local run cannot *detect* a fault, here it *invents* one.

### DIFF
--- a/tests/law/test_board_scoped_state.gd
+++ b/tests/law/test_board_scoped_state.gd
@@ -1,0 +1,138 @@
+# The invariant a shared test fixture rests on (#622): NO STATIC HOLDS BOARD-SCOPED STATE UNLESS
+# clear_board RESETS IT.
+#
+# Why this is a law and not a comment. `clear_board()` is the reset door, and it is correct -- but
+# it is HAND-MAINTAINED, and every line in it cites the leak it was added for: the Prolog accident
+# (stale last_loaded_path), a sandbox inheriting the last mission's look (#253), camera start
+# (#234), AI factions (#150), a tear-out left in the sky (#521), the playback lock (#484). Each of
+# those comments records someone not remembering. A new `static var` holding battle-scoped state is
+# covered only if the next person remembers too.
+#
+# So the classification is forced rather than trusted: a static in the board/presentation domain is
+# TUNING (process-scoped by design -- a `static var` rather than a `const` exactly so a knob can
+# write it), a CACHE (derived and idempotent), or BOARD-SCOPED (must die with the board). Anything
+# in none of the three fails here, naming itself, until its author says which it is.
+#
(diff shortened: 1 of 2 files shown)


Phaazoid:
Landed via #626.

Issue #624, opened 2026-08-28
WeaponInstance.spaces is lazily grown, so a weapon's saved form depends on what touched it


c3potheds:
🤖 Claude says:

`WeaponInstance.spaces` is an `@export` that is **lazily grown rather than sized from its template**, so whether a weapon writes it depends on whether anything happened to touch it — and a load touches it.

## Measured

Building the #622 fingerprint, `capture_scenario -> apply_scenario -> capture_scenario` on `Scenarios/missions/Prolog.tres` is **not a fixed point** on the first cycle. The in-memory diff, four weapons across two units:

```
board.unit_entries[7].inventory[0].spaces: 0 entries -> 3
board.unit_entries[8].inventory[0].spaces: 0 entries -> 3
board.unit_entries[8].inventory[1].spaces: 0 entries -> 3
board.unit_entries[8].inventory[2].spaces: 0 entries -> 3
```

Both forms mean the same thing — no mods fitted. `[]` is a weapon nobody has asked about; `[[], [], []]` is the same weapon after something called the space accessor, which grows the array on demand:

```gdscript
@export var spaces: Array[Array] = []
...
	while spaces.size() <= index:
		spaces.append(fresh)
	return spaces[index]
```

`WeaponInstance.make()` does not size `spaces` from `template.mod_spaces`; `copy_equippable()` copies whatever is there at the time. So the representation of "no mods" depends on history rather than on the weapon.

## Why this is worth a ticket rather than a shrug

**Measured:** the in-memory round trip above. That is the hard evidence, and on its own it only costs the test fixture a warm-up cycle (done in `tests/presentation/test_board_fingerprint.gd`, commented there).

**Inferred, and flagged as inferred rather than asserted** — someone should confirm before acting on it: a `.tres` omits a property at its default value, and the default here is `[]`. If that holds through `ResourceSaver`, then a scenario saved *after* a load writes `spaces = [[], [], []]` lines that the same board saved *before* a load does not, which would put this in the same family as [#111](https://github.com/Phaazoid/Godoiosis/issues/111) and [#540](https://github.com/Phaazoid/Godoiosis/issues/540): the code is correct, and the only symptom is a `git status` line the dev did not author. I have NOT measured the file-level half, and this repo's own rule about naming an unmeasured cause (#540) is why I am saying so instead of leading with it.

## Suggested shape

[...]

PR #625, opened 2026-08-28
#622: a shared board fixture, and the witness that makes it safe

c3potheds:
Leaves **A, B and D** of #622's Mikado stack. Builds on #623 (leaf C2, the law). No existing suite is converted — that is the next leaf.

One `Battle3D` per **suite** instead of one per case, with the leak check wired in rather than optional. Measured on the new smoke suite, identical assertions both ways:

| | shared | fresh |
|---|---|---|
| per case | **79–81 ms** | 825–1138 ms |

## Why a diff rather than "run the next test and see"

A leak observed by a *later* case is a flake — it depends on what that case happens to look at, reproduces sometimes, and blames the wrong test when it does. `BoardFingerprint` diffs the whole board against a baseline after every case, so a leak fails **at the case that caused it, every run, naming the field**. Speed traded for determinism would be the wrong direction; this is the version of the trade worth taking.

Coverage follows #622's taxonomy: the **board** via `capture_scenario`; **tuning** derived from the knob tables rather than listed, so this and the panels cannot drift; and the residue — staging, Dialogic, camera, `game_state`.

Reflection, not serialization: `ResourceSaver` mints `[sub_resource id="..."]` per save, so two saves of identical data differ and every case would report a phantom leak.

## The recipe needed two steps reasoning would have missed

The fingerprint found both on its first run. `apply_scenario` alone did **not** restore:

- **Dialogic stayed armed** from `load_mission` — `apply_scenario` is the *board* door and does not re-arm the lesson, so case 1 would run with a timeline live and every later case without.
- **`WeaponInstance.spaces` moved `0 entries -> 3`** — it is lazily grown rather than sized from its template, so a weapon's form depends on what touched it. Filed as **#624**; warmed around here rather than papered over.

## A design bug of mine, fixed

`check()` **resets before it diffs**, and my first draft had it backwards. Diffing what a case *left behind* fails every case that legitimately loads a mission or moves a unit — that is the case working, not leaking. What must be true is that the reset door can **undo** it. Keeping the call in `after_test` is what blames the case that caused a leak rather than its innocent successor.

## Fresh mode is the oracle and the escape hatch (leaf D)

`IOSIS_FRESH_FIXTURE=1` rebuilds per case exactly as before. Divergence between the modes **is** a leak, caught deterministically — and if sharing ever proves unsound the fallback is an environment variable, not a revert.

The smoke suite passes in **both** modes, and its last case is the demonstration: the previous case's damage is gone in fresh mode because the scene was rebuilt, and in shared mode because the reset door works.

## Verification

- `tests/presentation` + `tests/law` — **691/691, 62/62 suites, 0 orphans** (before these landed).
- The two new suites — **9/9**, and **4/4** again under `IOSIS_FRESH_FIXTURE=1`.

## Estimate correction

#619 and #622 put the saving at ~58% of CI on the assumption the reset is free. It is not, and I over-corrected in the other direction mid-analysis. The measured figure above is a suite whose case *bodies* are trivial, so it isolates fixture cost; a real suite keeps its body time and will see less overall. I would rather land the measurement than another estimate — converting a real suite is what produces the honest number.

### DIFF
--- a/tests/law/test_board_scoped_state.gd
+++ b/tests/law/test_board_scoped_state.gd
@@ -1,0 +1,138 @@
+# The invariant a shared test fixture rests on (#622): NO STATIC HOLDS BOARD-SCOPED STATE UNLESS
+# clear_board RESETS IT.
+#
+# Why this is a law and not a comment. `clear_board()` is the reset door, and it is correct -- but
+# it is HAND-MAINTAINED, and every line in it cites the leak it was added for: the Prolog accident
+# (stale last_loaded_path), a sandbox inheriting the last mission's look (#253), camera start
+# (#234), AI factions (#150), a tear-out left in the sky (#521), the playback lock (#484). Each of
+# those comments records someone not remembering. A new `static var` holding battle-scoped state is
+# covered only if the next person remembers too.
+#
+# So the classification is forced rather than trusted: a static in the board/presentation domain is
+# TUNING (process-scoped by design -- a `static var` rather than a `const` exactly so a knob can
+# write it), a CACHE (derived and idempotent), or BOARD-SCOPED (must die with the board). Anything
+# in none of the three fails here, naming itself, until its author says which it is.
+#
+# The tuning arm is DERIVED from GameKnobs.CLASS_KNOBS rather than listed: that table names 98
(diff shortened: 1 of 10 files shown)


Phaazoid:
Landed via #626.

Issue #635, opened 2026-08-28
The animation timing tool: authoring when a zoom frame plays, and where it stands


Phaazoid:
🤖 Claude says:

**Re-scoped 2026-09-14.** This was *"The other three zoom sheets: no cards, no timing, no registration"* — a question about four Fire Emblem rips in `Art/Units/ZoomAnimations/`. **Those sheets are deleted** ([#139](https://github.com/Phaazoid/Godoiosis/issues/139)'s last rip row, after [#937](https://github.com/Phaazoid/Godoiosis/issues/937) took the map sprites), and `CardSheet` plus `tools/zoomanim/` went with them.

The sheets were never the point. What survives is the thing that actually stopped that work, in the dev's own words:

> *"I never merged it because I was unhappy with the timing, and wanted to pause exploring that direction until I made a dev tool to help me create timings in animations."*

So this issue is now that tool. Old forks 2 and 3 already were it.

## The ask

A dev surface for authoring **when each frame of an animation plays** and **where its character stands**, producing a `SpriteFrames` the existing machinery can drive.

Both halves are needed and they are separate facts:

- **Timing** — how long each frame holds. This is what the dev was unhappy with, and it is the reason [#603](https://github.com/Phaazoid/Godoiosis/issues/603) is paused rather than abandoned.
- **Registration** — where the character stands inside each frame. `UnitSprite3D` reads this off the set as `ground_point` metadata and warns when a set carries none, so the door already exists and is already documented at the source.

## What already exists to build on

**`SpriteAnimator` and `UnitSprite3D.play_animation` survived the deletion on purpose.** They drive *any* `SpriteFrames`, so they are unfed rather than dead — this tool is what feeds them. `SpriteAnimator` is static and pure over a frame list, and `UnitSprite3D` already handles the borrow (an animation takes `texture` against the one door and hands it back), the `ground_point` pivot, and interruption when a unit goes down. **None of that needs rebuilding.**

`DevController`'s K hotkey still reaches for `Resources/ZoomAnimations/<art family>.tres` and warns that nothing is there. That convention is the output contract: land a set at that path and the hotkey plays it.

## Forks, still unpicked

1. **Where the authoring lives** — a dev-tools page with a frame strip and per-frame duration fields, or a `.tres` authored in Godot's own inspector with the panel only for *previewing*? The panel is the bigger build and pays off if more than one unit ever gets animated.
[...]

PR #643, opened 2026-08-28
#418: health bars get a third mode (damaged only), and DEFS gets a choice kind

### LINKED ISSUE #418: Health bars: a third player-settings option for damaged units only

Phaazoid:
Closes #418.

The player picks between three now: **under the cursor** (#229, still the shipped default), **damaged units** (new), **every unit** (#350). A player who never opens settings sees exactly what they see today.

## The model half is one term, as promised

`UnitMirror`'s gate is still `hovered or foretold or marked or preferred` — four disjuncts, one named expression. Only the last one moved, from a bool to a three-valued read. #357's icon row rides the new mode with nothing to keep in step, which is what the "one gate, deliberately" ruling was protecting.

`foretold` is untouched and has its own case under the *new* mode, because the always-on cases structurally cannot see it: with EVERY set, every bar is up for the preference anyway.

## The real work was the settings table — structural choices to rule on

`DEFS` described booleans and `SettingsScreen` is a pure projection of it, so this is **the first non-checkbox row the page has ever had**. Four calls worth naming in prose:

1. **A row's KIND is derived from the table, not stated in a second field.** A row with `options` is a choice; `is_choice` asks that. No `"kind"` key to keep in step.
2. **A generic core with typed façades.** `value_of` / `set_value` / `default_value` are what a caller walking `DEFS` uses — `board_fingerprint` is the one such caller. `is_on`/`set_on` and `choice_of`/`set_choice` sit over them. `default_of` folded into `default_value`. **The five production `is_on` readers are untouched**, `GameTool`'s battle-zoom checkbox included.
3. **The projection property survives, and that was the whole argument for this shape over two bools.** The page learns a row is a toggle or a choice, never *which setting* it is. The test that pins "every declared setting gets a row" now asks about both kinds — narrowing it to the toggles would have quietly retired the property for exactly the rows that need it.
4. **`HealthBars`'s enum values ARE the option indices** — a declared duplicate (Law #4), enum authoritative, with a test holding the two in step. Nothing else could notice them drifting: a list one label short leaves EVERY unpickable, silently, with the strip looking right.

## The widget: a segmented strip, and a dropdown was structurally out

An `OptionButton` opens an embedded `PopupMenu`, and `CLAUDE.md`'s SubViewport gotcha (2) records that those **do not dismiss on outside-click** inside `GameView` — the #26 reason the action menu has been Control-based ever since and "must not go back to a `Popup`". So it is three toggle `Button`s in one `ButtonGroup`: `ReportPanel._build_kind_row`'s shape, already shipped and player-facing, whose own comment states the other half of the reason. Two modes chosen at once is unrepresentable rather than merely avoided.

It listens on `toggled`, not `pressed`, so what the store mirrors is the segment's **state** — the same idiom as the checkbox rows above it, and the reason a test can drive it by moving the control rather than by emitting the signal the handler happens to listen for.

## One thing that lands on you

**`ALWAYS_SHOW_HEALTH` became `HEALTH_BARS`, so its cfg key goes dead and your saved "always on" resets to hovered-only once.** That is this store's already-declared behaviour (a deleted member leaves a dead key) and the project bans legacy shims, so there is no migration. Flagging it because your play-checks run under your own `settings.cfg` — if the board looks bare on first launch, that is this, and the fix is one click on the new strip.

## Rulings taken before building

- **DAMAGED is HP below max, full stop.** The alternative was "anything to report" (HP loss *or* a held element state), because #357's row rides the bar's gate — so a full-HP SOAKED unit wears no icons in this mode until you hover it. Reachable (heal a burned unit to full and it keeps the state); ruled acceptable, since thinning the board is the point.
- **A body needs no clause of its own.** `_go_downed` clings at 1 HP, so it is damaged by the ordinary rule and #322's glyph and rescue clock stay up.
- An out-of-range cfg value falls back to the default — the cfg is a text file you can open, so a bad index is reachable *input*, not a bug.

## Verification

`tests/core` 71/71, `tests/ui` 271/271, `tests/presentation` 536/536 — 0 failures, 0 orphans. Area folders only; CI owns the tree.

**Falsified, three mutants, one suite file each, committed first:**

| mutant | result |
|---|---|
| delete the `DAMAGED` clause from `preferred` | `test_damaged_only_thins_the_board_down_to_the_hurt` reds — **and the three EVERY cases before it stay green**, so the two do not overlap |
| `<` → `<=` in the damaged predicate | same case reds. This is the one I expected might no-op (it would, if the fixture had only one unit); it does not |
| the strip writes a constant index instead of `i` | `test_picking_a_segment_writes_the_chosen_mode` reds — the second half of that case presses a *different* segment for exactly this |

**What a headless run could not see, and is therefore yours:** how the three labels read at 520px across, and whether "damaged only" actually thins the board the way the crowding complaint wanted. `render_priority` crowding is untouched and still open — this mode is a workaround for it as much as a preference.

**Embedded-content sweep: not applicable, checked rather than skipped.** `PlayerSettings` is a static store, not a content type; nothing under `Scenarios/` or `tests/fixtures/` embeds it, and no `.tres` field was added.

## Canon

`visual-clarity.md` gains a #418 section under #350's; the inbox capture that produced the ticket is **deleted** (built means deleted). Two stale names fixed where the rule stayed right and only its justification moved — the gate's `always_on` → `preferred`, and the impact-above-the-gate argument now names `HEALTH_BARS`/`HOVERED`. Marker to #635. `CLAUDE.md`'s `PlayerSettings` entry gains the two kinds and the no-dropdown reason.

`docs/SCRATCHPAD.md:54` left alone deliberately — it is a dated Dispersed-log entry recording what was true at filing, not canon. Say if you'd rather it were annotated.


Phaazoid:
## Follow-up: the settings page could not be left (found in play)

The choice row made the card taller than the 1280×720 viewport, so **Close sat off the bottom edge and opening settings locked the game permanently.** Fixed on this branch rather than a new one, since this PR is what introduced it. Two halves, and the second turned out wider than the symptom.

### The body is bounded now

A `ScrollContainer` between the title and Close — `GlossaryScreen`'s shape, horizontal scroll disabled — sized from **what the viewport actually leaves**, not from a constant. A sixth setting cannot do this again.

The chrome is measured, and **my first attempt was wrong in a way worth recording**: asking the card root for its combined minimum size returns `(0, 0)`, because a plain `Control` aggregates nothing — only a `Container` rolls its children up. It asks the outermost container instead. Chrome measures 290; the body gets 430 of 720.

### Esc is a real door, and it wasn't only a settings gap

`game.gd:227` stands down on `ui_cancel` while any modal is open (#131), so **a card that doesn't answer the key swallows it** — which `PauseMenu` and `GlossaryScreen` were also doing.

The mechanism moved to `ModalCard`: one `_input` plus an `_on_cancel() -> bool` virtual. **Mechanism shared, policy per card**, and the bool is what made that possible — `ReportPanel` swallows the key mid-send *without acting*, which a plain `dismiss_on_cancel` flag couldn't express. The three hand-written handlers (`ReportPanel`, `SaveLoadScreen`, `ConfirmCard`) collapsed into it with their policies verbatim, including `SaveLoadScreen`'s `_busy` yield to a stacked `ConfirmCard`. `MissionEndBanner` and `MissionSelectScreen` keep the do-nothing default — an ending is chosen, not dismissed.

Esc now resumes from the pause menu and closes the glossary too.

### Falsification — two mutants passed, and both were real findings

`ui_cancel` had **zero test coverage anywhere in `tests/`** before this, so nothing here inherited a safety net.

| mutant | result |
|---|---|
| remove the scroll's ceiling | **PASSED at first.** It collapses the body to zero height, and a page showing *nothing* trivially fits any viewport — "the exit is on screen" is only half the property. Both fit cases now also assert the body is real; the mutant reds (`body is 0.0 tall`). |
| give the body an over-tall ceiling | reds at `bottom edge is at 2145.0 on a 720.0-tall viewport` — **and `test_pause_settings_close_lands_back_on_the_pause_menu` stayed green**, which is the blind spot demonstrated: it presses Close by calling it, so it's happy with a button nobody could reach. #131's shape one level up. |
| `_on_cancel`'s default returns `true` | **PASSED — a genuine no-op.** For a single card that only swallows the key; it dismisses nothing. The default's value is observable only in a stack. Recording it rather than inventing a contrived case. |
| the base dismisses every card on Esc | the realistic version of the above — reds at `Esc dismissed the mission-end banner, which has to be answered`. |
| the settings page loses its `_on_cancel` | reds at `Esc did not close the settings page`. |

**The stacking case is measured, not assumed** — I said in the plan I didn't know Godot's `_input` order for stacked siblings well enough to assert it. It does give the `ConfirmCard` the key first, and there's now a case pinning that Esc answers it *and does not reach the screen underneath*.

### Verification

`tests/presentation` + `tests/law` + `tests/ui` together: **1007/1007, 0 failures, 0 orphans.** Run after merging `origin/main`, which landed #636's `SharedBoard` change under the presentation fixtures — that's why all three areas rather than just `ui`.

**Still yours:** whether the scrolled page actually reads well — how much you see at once, and whether the strip and Close land where you expect. Headless can measure that it fits; it can't tell you it looks right.

### Canon

`CLAUDE.md`'s `ModalCard` entry gains both rules (Esc belongs to the card; a card whose content can grow must bound its own body, including the outermost-container measuring trap). `visual-clarity.md`'s #418 section records what the row exposed and the two testing lessons.

Issue #472, opened 2026-08-22
A unit shoved off a cliff starts falling before it finishes travelling, and reads as a midair teleport


Phaazoid:
🤖 Claude says:

*(Dev report, 2026-08-22, filed at his request.)*

## The symptom

Shoving a unit off a cliff, **the unit does not finish travelling horizontally before it starts falling** — it reads as having *teleported a bit in midair*.

## Where to look, and what I did NOT conclude

I read the two surfaces rather than diagnosing, because the mechanism a report names is exactly the thing worth re-deriving ([#228](https://github.com/Phaazoid/Godoiosis/issues/228)'s law). Candidates, in the order I would test them:

**1. The slide advances `cell` BEFORE its tween runs.** `MovementComponent._slide_to_next_cell` sets `cell = next_cell` and *then* tweens `unit.position` toward that cell. So the authoritative cell **leads the sprite by up to a whole segment**, and anything reading `cell` mid-slide is already at the destination. `UnitMirror` reads `unit.position` for the horizontal, so this may be innocent here — but it is the first thing to rule out, and it is a fact worth knowing whatever the cause turns out to be.

**2. The seam between `sliding`, `kb_settling` and `plummeting`.** `UnitMirror` drives a shoved sprite's height through three branches in one `elif` chain:
- `plummeting` → `stand_y -= plummet_depth * CELL_SIZE`
- `sliding` → held at `slide_origin`'s height while `airborne`, else the surface plane under the sprite
- otherwise `kb_settling` → eases toward `stand_y` at `shove_fall_speed`

A **void** shove runs slide → (slide ends, `sliding` false) → `plummet()` is awaited. In the gap, `kb_settling` is what owns the height, easing toward whatever `surface_point(over)` is for the void cell — and then `plummeting` takes over with a different expression of the same height. **Two authorities for "how far down is this sprite" meeting at a frame boundary is the shape a positional jump comes from**, and the comment above that chain already records that the plummet branch had to be checked FIRST for a related reason.

**3. `airborne` is per-segment and decremented at segment START.** `_flight_entries_left` is set to `landing_index` and each segment does `airborne = _flight_entries_left > 0` then decrements. Worth confirming by measurement that the final segment is still airborne on a **cliff** (a drop with no ramp), because the comment claims fly-then-fall is the intended flat-landing behaviour and the report says the fall starts early.

## First job

[...]

Issue #545, opened 2026-08-26
Battle zoom: a skip for playback - what it stops, and how you ask for it


Phaazoid:
🤖 Claude says:

Spun out of #520 (dev, 2026-08-26: *"I think we should provide a skip button, but leave that as a separate issue, it has some forks to decide"*).

#520's own *Done when* already assumes one exists — *"Manual camera input is suppressed for the duration and restored after — **including on skip**"* — and nothing like it has ever been built. This is that ticket.

## Ask

A way to cut playback short. Every pass is pure playback of an already-resolved plan (R3, Law #2), so nothing about skipping can change an outcome — the question is entirely about *what stops* and *how you ask for it*.

## What exists to build on

- **`Pacing.beat(host, seconds)` is the one door every pause routes through**, and it already has a "nobody is watching" escape (headless returns instantly). A skip is that escape, turned on at runtime instead of by `DisplayServer`.
- **`OrderExecutor.execute_orders` is the one path** every plan resolves through — player Execute, action-menu entry, and `AIController` all land there.
- **`AIController.take_faction_turn`** owns the outer loop over squads, with its own `pan_to` + `AI_PLAN_READ` beat per squad.
- Playback ownership is `CameraController.playback_locked`, saved and restored rather than cleared (#520), so a skip must not blind-release it.

## The forks, none of them decided

**1. What does one press skip?** The current **beat**, the current **pass** (this squad), or the whole **turn** (every remaining enemy squad)? These are three different affordances and the enemy turn is where the want is loudest.

**2. Fast-forward or jump?** Zeroing the beats plays everything at once and is nearly free. Jumping to the settled state is not free and needs care: `AttackAction.execute` awaits `movement.slide_along_path` for a knockback and `movement.plummet` for a void removal, and those awaits sit *between* state changes — the removal lands after the fall. Auditing what is safe to cut is most of the work in this option.

**3. Press, or hold?** A hold-to-fast-forward is self-cancelling and needs no "unskip"; a press needs a decision about how the state ends.

**4. Per-instance, or a setting?** "Skip this one" versus a `PlayerSettings` tenant meaning "always". Note the toggle that already exists is `Battle zoom`, and *off* is not the same as *skip* — off still paces (#519's BOARD profile), which was the whole point of that slice.

[...]

PR #658, opened 2026-08-29
#602: the camera rides a falling body down instead of watching it leave the frame

### LINKED ISSUE #602: Battle zoom: the cliff follow - the camera rides a falling unit down
## What

A unit shoved off a ledge or into a void falls. The camera should leave the board and follow it down, so the drop is a shot rather than something that happens below the frame.

Spun out of #520 diff 2b at the dev's call (2026-08-27). The rest of that slice — pitch driver, impact shake, micro-sway, knockback follow — is composition over channels that already exist. This one is not, and that asymmetry is the whole reason it is its own ticket.

## Why its sibling is free and this is not

**Knockback follow costs nothing.** `CameraController.pan_to(unit)` closes with `follow(unit)`, and `_process` re-reads `follow_unit.global_position` every frame, so a unit sliding along its knockback path is already tracked by machinery that has been there since #118.

**The fall is invisible to exactly that mechanism.** Both fall animations are **mirror-side Y offsets**, never board movement:

- `MovementComponent.plummet_depth` — the void fall (#431), tweened to `VOID_PLUMMET_CELLS`
- `MovementComponent.landing_fall_depth` — the drop onto a lower surface

`UnitMirror` folds both into `stand_y` when it places the sprite (`Classes/presentation/UnitMirror.gd:390-394`). The unit's own `position` never descends. So `follow_unit` tracks the body to the lip and then holds level while the sprite drops out of frame — the camera is dutifully following a unit that, as far as the board is concerned, has already arrived.

## What it needs

A **3D vertical authority for the camera**: something that can take the rig below the board plane and bring it back. Nothing has that today. The rig's position is `_aim + _lift`, where `_aim` comes from `BoardSpace.of_pixels(cam.global_position, 0.0)` — the flat board point, Y hard-zeroed — and `_lift` is #521's tear-out.

**Do not scope this as reuse of #521's lift.** That lift is per-**cell** staging: a second GridMap whose *node transform* displaces the ground the fight happens on. A falling **unit** is not a cell and does not live in that lattice. Sharing the channel would mean the tear-out and the fall fighting over one addend, which is a duplicate-seam failure that only shows up when a shove into a void happens during a staged fight — i.e. exactly the case this is for.

## Forks, unpicked

1. **Does the camera descend, or tilt down?** Descending keeps the body centred and is the literal reading. Tilting keeps the board's lip in frame and reads as watching someone fall away from you — and it is much cheaper, since pitch is already an eased channel after 2b. The cheaper one may also be the better shot.
2. **Does it come back up, or cut?** The view return already fires on the release edge, so "come back" may be free.
3. **Does the fall earn its own hold/linger rows**, or ride `HOLD_KNOCKBACK`?
4. **A void fall ends in `die()`.** Does the death beat — its linger, and 2b's death shake — land while the camera is below the board plane?

## Notes

- `plummet()` carries the same headless escape `Pacing.beat` does, so no test can watch the descent. Expect the pin to be the decision and the wire, with the motion itself declared a play-check — the pattern this arc has used since #519.
- Wants #520 diff 2b's impact layer landed first: it introduces the pitch channel and the flourish gate this would compose with.

Cross-refs: #520 (parent), #410 (umbrella), #431 (the plummet animation), #521 (the lift this must NOT reuse).

Phaazoid:
Closes #602.

A unit shoved off a ledge or into a void now gets a **shot** instead of leaving the bottom
of the frame. Its sibling item — knockback follow — cost nothing, because a shove writes
the unit's own `position`. A fall is invisible to exactly that mechanism: both fall
animations are mirror-side Y offsets, so the board thinks the body has already arrived.

## Three forks, your calls (2026-08-29)

**Ride it down**, not tilt and not a cut. **A real hold at the bottom** before the body is
removed. **Every fall**, not just the void plummet — the mid-shove cliff drops get the same
channel.

## The seam

`UnitMirror.stand_height` / `fall_depth` / `cell_under`, extracted verbatim out of `_sync`.
The camera reads the arithmetic the sprite is placed by — **a second spelling of this exact
fall is what #472 was filed for**, and a mutant deleting the plummet branch reds the mirror
suite *and* the camera suite, which is the proof the seam is genuinely shared.

There turn out to be **three** vertical animations, not the ticket's two: an airborne shove
holds a body *above* the ground it sails over, so the same subtraction goes negative there.
Declared cut — the camera floors it, and there's a case pinning the sign so the floor isn't
dead code. One `maxf` away if you ever want a body blown over a chasm followed upward.

## The channel

A fourth position addend on the rig, and **asymmetric**: down lands at once (any easing is
lag, which is the defect), up is eased on its own knob. The slow climb is also what keeps
the camera deep in the pit while the death jolt and `LINGER_DOWN` play over it.

Deliberately **not** a second use of #521's lift, per the ticket — that's per-*cell*
staging, and the two would fight on a shove into a void during a staged fight.

The hold lives **inside** `plummet()` rather than at the call site, because `plummet_depth`
is what the camera is riding: clearing the flag and then waiting would have the rig climbing
out during the very pause the pause is for.

## Two premises in the ticket had gone stale

The aim is **not** Y-zeroed — `_aim_over` has re-derived it from the board surface since
2026-08-23, which is exactly why a fall escapes it (it tracks the *surface*). And vertical
authority no longer has to be invented; `camera_lift()` is a driven channel since #521
slice B. The refusal to *reuse* that lift still stands. Recorded on the ticket.

## The bug the fixture found

**A board swap mid-fall left the rig under the new board for ever.** `frame()` calls
`drop_stashed_view()`, so `restore_view` early-returns and never runs, and the drop is
polled *below* the playback gate — nothing closed the channel. Found by `SharedBoard`'s
leak fingerprint, not by reasoning: every fall case leaked the same three cells into the
next one. `drop_stashed_view` is now the door, and it **cuts** where `restore_view`
**climbs**. New sharp edge in `CLAUDE.md`.

## Knobs — Playback page, new "The cliff follow" section

| row | ships at |
|---|---|
| How much of a fall the camera rides | 1.0 |
| How far down it will follow | 8.0 cells |
| How fast it climbs back | 1.6 e-folds/sec |
| Hold: the bottom of a void fall | 0.7s |

**Flat across profiles**, on the linger's reasoning rather than the sway's: a fall runs in
real time whichever profile is up, and the body leaves the frame just as thoroughly with the
battle zoom off. The strength row is the dial-out. **All four are guesses** — the descent
rate isn't (it's the body's), but the climb-back and the bottom hold are pure feel, and given
last session I'd expect the hold to want to be longer.

## Verified

- `presentation` + `law` **774 cases, 0 failures, 0 orphans**; `dev` + `core` **471, 0, 0**.
- **Six mutants, six precise kills.** The plummet branch (reds *both* suites), a symmetric
  ease, the follow ceiling, the board-swap door, the release-edge door, and the poll itself.
- **A mutant reddened the wrong case once** and it was informative rather than a defect: the
  leak fingerprint fires on the earliest fall case and truncates the file, so the board-swap
  case had to be checked with the earlier ones disabled. It reds on its own assertion.

## Declared blind spot

`PLUMMET_HOLD` is unobservable headless — `plummet()` returns before the flag is raised, and
`Pacing.beat` would spend nothing anyway. Deleting it turns nothing red. Said so at the call
site; it's a play-check like the rest of the pacing in this arc.

## What to look at

Shove somebody off a ledge, and shove somebody into a hole. Two things I can't check:
whether the shot below the board **reads** (it's sky, volumetric fog and the board's
underside receding above — that's what the ceiling knob is for if it doesn't), and whether
the beat at the bottom is anywhere near long enough.


Phaazoid:
## Round 2 — what a ten-level pillar found (2026-08-29)

Your first play-check on genuinely tall terrain. Four things, three of them mine.

### The tear-out looked at nothing — two causes, both #521's

Your third report is the proof: the frame is entirely empty, and the readout says the rig is centred
on `(30, -12)` while the fight is at `(16, -7)`.

**Nothing aimed the camera at the fight before the ground moved.** `_stage_the_fight` is the first
thing a pass plays, and `_frame_the_walk` returns early on an EMPTY span — which is exactly what
"Isaac holds position" produces. So seven to twelve seconds of brace, flight and settle played over
wherever you had left the view. It pans to the fight's own bounding-box centre first now.

**And the aim took the height of the wrong ground.** The rig's height is `aim + lift`, and the LIFT
channel has carried the rule in its own comment since #521 — *"the whole stage, never the cell under
the camera: the diorama is one thing at one height"* — while the AIM read `surface_height_at` under
the *camera*. Flat ground hides it; a fight on a pillar puts the diorama at the pillar's surface plus
the lift and the camera at the plain's. The rule existed, was correct, was written down, and governed
one of the two channels that answer the question. It is also your second-order jerk: a pan crossing
the pillar's edge stepped the camera by the pillar's whole height.

### The fall stops hugging the wall — your catch, and you had it right

I had diagnosed a vertical pop; the bigger half was the one you saw. The slide stopped at the
**midpoint of the shared edge** and fell from there — in the boundary plane, half inside the wall for
the whole 2.5s descent — then ran the leftover half-step in a sixteenth of a second, which is the
snap to centre. It steps clear into the landing cell first now.

Underneath it, a sharp edge worth having: **the midpoint of two adjacent cell centres sits exactly on
their shared boundary, and `floori` resolves a tie to the `+` side.** So the cell under the sprite
read as the one being LEFT for a shove north or west and the one being ENTERED going south or east —
wrong half the time, invisible the other half, and it survived a suite that only ever shoved one way.

### The void fall, as you described it

The follow ceiling ships at **3 cells** rather than the whole 8-cell plummet, so the body genuinely
leaves frame; `PLUMMET_HOLD` is the beat; and the death bricks burst a knobbed distance under **where
the camera stopped** rather than wherever the corpse ended up, so they rise into the bottom of the
shot. `Pacing.followed_fall` is the one answer to where the shot stopped — read by the shot and by
the readout, because two spellings would put the cubes somewhere the camera is not.

### ...and the camera comes home before the tiles do

The climb is eased and longer than `TEAR_OUT_AFTERMATH`, so the board used to start reassembling
mid-climb. The exit waits on the rig's own published depth rather than on a beat, which would be a
second answer to how long the climb takes.

### Verified

- `dev` + `core` + the camera suite **513 cases, 0 failures, 0 orphans**; `presentation` + `law`
  **1217 across the wider sweep**, one failure which was the knob law below and is fixed.
- **Six mutants, six precise kills**: the missing pan, the aim ignoring the stage, the old
  drop-then-step order, the exit not waiting, the rig never publishing, the burst never raised.
- The return-home case drives the **real exit** rather than the wait helper — my first version called
  the helper directly and would have passed with its one call site deleted (#103's shape).

### Two process notes, both mine

**I wrote two hunks into your working tree by mistake** — a pair of scripts run with relative paths
while the shell was still in `Godoiosis`. `Classes/dev/GameKnobs.gd` and
`tests/presentation/test_camera_follow.gd`, reverted there and applied here instead;
`docs/SCRATCHPAD.md` was never touched. The missing knob arms are what the CI-visible
`test_every_class_knob_resolves` failure was.

**New knob:** *Bricks burst this far below*, on the Playback page beside the rest of the cliff follow.
And **the follow ceiling default moved 8.0 → 3.0**, which is the change that makes the body leave
frame at all — if it now stops too early for you, that is the row to raise.

Phaazoid:
## Round 3 — the diorama is framed on its FIGHTERS (2026-08-29)

Your round-2 verdict: the camera is on the right cell now, but *"the units need to be at the
center"* — it stares at the pillar's wall with the fight above the frame.

### The framing was mine

`BeatSheet._gather_cells` puts the **whole knockback path** on stage. That is right for the *ground*
— a shoved body has to land on something — but round 2 made the camera's height the **average of
that ground**. On your board the pillar top is world y **6.0** and the cells Noemie was thrown onto
are **1.0–2.0**, so the average sits between them. The further a body is thrown, the worse it gets.

It reads the ground under the **units on stage** now, plus `Pacing.STAGE_AIM_LIFT` — a knob, because
"centred" on a sprite is a feel value and aiming at their feet is what leaves them high in frame.

**`DIORAMA_BYSTANDERS` was my first suspect and it is OFF in your `experiments.cfg`.** Checking it is
what stopped a fix for the wrong thing. Worth noting for the future: your report's `View:` line named
the fault in one number — *centred on (30, -12)* against a fight at *(16, -7)*.

### Latched, per your pick

*Stage height held, dip for a fall.* The height is decided when the ground leaves the board; only the
cliff follow moves the camera after that. Bodies get thrown mid-pass, and a live re-solve would find
them standing on whatever they landed on and walk the whole shot down after them.

### Two mutants survived, and that is the interesting part

I wrote that the latch stops a fall being counted twice. **Deleting the latch left every case green**
— the height reads the *ground*, and a falling body's ground does not move.

So I rewrote it: the ground read is what buys it. **Swapping the ground read for a stand-height read
also left everything green** — because the latch settles the height *before* anyone is falling, so at
the only moment either is evaluated the two agree exactly.

New sharp edge in `CLAUDE.md`: **a latch makes every choice behind it unobservable.** Ask what state
exists at the instant the latched value is computed; anything that only differs later is invisible to
every test, however carefully the case is written.

The latch's *real* property — a body thrown onto lower ground must not walk the shot down — has a
case now, and it kills that mutant precisely. The ground read stays, because it is the right thing to
mean, and the comment says outright that no test covers it rather than implying one does.

### Verified

- `presentation` + `dev` + `core` + `law`: **1252 cases, 0 failures, 0 orphans**.
- Four mutants: the units-vs-ground read, the latch (via the new throw case), the shot lift, and the
  two survivors above, both now accounted for rather than papered over.
- The #521 tear-out case **gained a term rather than being weakened** — the camera rises by the stage
  offset *and* the shot's lift, both re-derived from their own seams.

### New knob

**Shot sits above the units' feet** — Playback page, beside the rest of the cliff follow. Ships at
0.5 cells. If they still sit high or low for you, that is the row.

Phaazoid:
🤖 Claude says:

**Round 4 — the playback director.** Four rulings from the pillar play-check, all recorded in `visual-clarity.md`'s new Round 4 section (dev, 2026-08-29): *"we control the camera, fully. Their zoom gets overridden, period"* (restoring #520's own Done-when after the 2026-08-26 wheel carve-out — the no-floor rule keeps its scope outside playback, untouched); *"the camera should follow the unit... I want a close up of that unit's tumble, all the way through"*; *"I'll always want the death cubes in shot"*; and the camera-states idea, which turned out to need **no new enum** — the facts already published decide the shot.

## What each complaint was, mechanically

1. **"Stares at the wall before whiting out"** — the aim's stage-height read was gated on `BoardSpace._staged`, which fills only at `stage()`, so the whole pan+brace approach hugged the terrain under the moving centre: on your board that is frame centre climbing a five-unit cliff face. The stage is now **published before the pan** (`CameraController.shot_cells`, the fourth field beside angle/fit/weight), so the approach aims at the fighters from its first frame — and the WIDE shot widens to fit the whole staged volume, so the arrivals land in frame. `TEAR_OUT_EMPTY_SKY` default went 0.7 → 1.1 so the first tiles arrive after the flash tail (ramp+hold+ramp = 1.1s) instead of behind it. The cut-vs-travel treatment is still the `DIORAMA_CAMERA_CUTS_AHEAD` experiment if you want to watch the tiles *leave* too.

2. **"Not following falls over non-death pits"** — deliberate round-2 design, now repealed: `followed_fall` floored negative depth at zero, and a landing fall is *above* its destination surface the whole way down, so the camera held at the lip while the body dropped out of frame. The trained shot now rides its subject through **every** fall — landing falls to the ground, void plummets to the follow cap, and a subject standing above the stage's mean aims the camera UP (the drop channel runs signed now, eased through its negative range). The test whose name pinned the old rule was rewritten with it.

3. **The vanished death cubes** — two independent faults, and your tuning could never fix either:
   - The burst was raised relative to the **void's lip** while the frame hangs off the **stage's height** — one shared formula, two different bases, 5 units apart on your board (0 on flat ground, which is why round 2 looked right there). Computed offline against your report's `board.tres`: old burst at y −3.5 vs frame bottom −1.87 — 1.6 units below the frame at every slider setting. The burst is now anchored to the frame the camera actually holds (`UnitMirror.frame_floor`, wired like `report_impact`), and the under-frame offset is a **const, not a knob** — a slider that could push the show out of shot would quietly repeal the invariant.
   - The recovery **raced the debris**: the climb started the frame the body was freed, and won at every `CLIFF_RECOVER` setting. It is event-gated now — the shot holds its depth until the last cube lands (`death_show_live()`), then climbs; the tear-down waits for the whole show. Also: `CLIFF_RECOVER` is a *rate* — your 1.6 → 2.0 made the climb-out FASTER; the tip now says HIGHER IS FASTER out loud. And yes, `Void fall depth` on the Motion page fed the old burst arithmetic; it no longer touches the shot at all.

4. **"Reset to full battle view before the platforms drop"** — the tear-down pans back to the stage centre (clearing the follow, so the shot is WIDE again) before the existing camera-home wait, then the tiles go.

## The two shots

`follow_unit` valid ⇒ **TRAINED** (new `Shot: trained on a unit` knob, default 7.0, dolly rides on top); `shot_cells` published ⇒ **WIDE** (playback distance widened to fit the stage); neither ⇒ plain mirror. `Beat.subject()` was already the victim and `pan_to` already ends in `follow()`, so the tumble close-up mostly existed — round 4 added its vertical half and took the wheel.

**Knob changes on the Playback page:** `How much of a fall the camera rides` (the fraction) and `Bricks burst this far below` are **deleted** — the first is the ruling (always ride), the second is the invariant. `How far down a death fall is followed` is your one follow toggle; `Hold: the bottom of a void fall` is the beat before the bricks; your committed HOLD tuning (Crisis 0.8, Knockback 0.75) landed as defaults; `PLUMMET_HOLD` ships 0.9 with its new meaning.

## Verification

`tests/presentation` 553/553 green (three earlier reds were mine: an untyped-base typed-array assignment, a NaN from a 0/0 lifetime, and over-deleting the dolly floor's contribution cap — that last one caught by an existing case, and the arithmetic stands on a reason the old comment never named). **Six mutants, each killed by exactly its aimed case**: the burst arithmetic, the death-show hold, the ride clamp, the stage publish, the trained-zoom edge, and the stage latch — the one that survived two rounds ago now has a case with real teeth (`test_a_thrown_body_nobody_is_following_does_not_walk_the_stage_down`).

## Housekeeping

- Your scratchpad commit rides this push (`b8faf531`, bundled as asked).
- **Your working tree holds uncommitted `Classes/core/Pacing.gd` tuning** — its surviving values are round 4's defaults now, so `git checkout -- Classes/core/Pacing.gd` before `git pull`, or the pull will refuse.
- The F2-mid-animation selection bug is filed with a diagnosis and repro sketch as #661 (the camera doors held; it is a scenario-reset door, unit-side).
- The skip — your *"only control is skip"* — is #545, exactly as you left it, forks and all. Until it lands, playback is fully hands-off.

You are on `feat/602-cliff-follow` — discard the Pacing diff, pull, and launch.

— Claude (Fable 5) · 2026-08-29

Phaazoid:
🤖 Claude says:

**Round 5 — the flash rides the cut, and the bar forms off screen.**

## The white-out

Your read was exact: the exit's flash was riding the ENTRY's clock — it played at the exit flight's *start*, over nothing, and the 40-unit drop home ran bare after the tiles fell. Now the two are one schedule under one rule, both directions: **the cut sits at the flash's first full-white frame, and the flash anchors to the transition's cut-ward end** — the entry's start, the exit's landing.

- **Exit**: tiles fall dark (unchanged — you said the falls read right), the flash ramps the moment the last one is home, the camera drops at full white, the fade reveals the grounded view. The cut lands ~one ramp (0.35s) later than before, which is inside "the timing is fine now"; ramping *before* the last tiles landed would have veiled them instead.
- **Entry**: the cut up now waits for full white too (it used to teleport at t=0 with the flash only beginning). `Empty sky` at 1.1s already equals the flash's full length, so arrivals still land after the fade.
- The drop is **snapped, not eased** — the rig's lift used to glide, so a "hidden" teleport was still settling as the fade revealed it.
- The schedule lives in `StagingFlight` — the same artifact the tile travel already reads — so the flash, the camera, and the executor's await share one clock and cannot drift. All your existing flash knobs (`White-out ramp`, `hold`) keep working and move the cut with them, which is the "no matter what."

Verified offline across the full exit and entry timelines at shipped knobs: there is no instant where the camera moves with the flash below full white, in either direction.

## The death bar

Also exactly what you saw: round 4 anchored the burst 1.5 cells under the frame's *centre*, and at this game's narrow fov (30°) that is well inside the frame — so the staggered march assembled a visible full health grid before a single cube launched. The anchor is now the frame's true bottom **edge**: `distance·tan(fov/2)/cos(pitch)` below the aim (~2.5 units at the trained distance — the pitch matters, a tilted camera sees deeper below its aim), read live so the dolly's lean-in moves the edge with the frame you're actually watching. The grid assembles just under that line, off screen, and the death burst's own launch power carries the cubes up into view — bricks erupting from a pit you can't see into.

## Verification

`tests/presentation` 603/603 green. Three mutants, each killed by exactly its aimed case: the exit flash re-anchored at zero (your bug, reintroduced), the entry cut at t=0, and a re-landing guard in `advance_flight` — that last one is a fault the flash tail *exposed*: an exit tile past its land time was erased-and-readded every frame, bumping `staging_version` and rebuilding every prop per frame; a cell lands once now.

You are on `feat/602-cliff-follow` — pull and launch. (Your tree is clean this time; no local discard needed.)

— Claude (Fable 5) · 2026-08-29

Phaazoid:
🤖 Claude says:

CI caught one thing the presentation area couldn't: `tests/law/test_board_scoped_state.gd` — the #622 law that makes every new static in the board domain declare whether it is tuning, a cache, or board-scoped state that must die with the board. Round 5 added two (`_flight_total`, `_camera_lift_snap`) and both die exactly where their siblings do, inside the `clear_staging` chain — they just hadn't been declared. `70714e1a` declares them; the law worked precisely as designed.

— Claude (Fable 5) · 2026-08-29

Phaazoid:
🤖 Claude says:

**Round 5 is REVERTED** (`1b5081b5` — three revert commits; the branch is byte-identical to round 4's `230fdd92`, verified by diff). Your reports showed round 5 broke the world far beyond its two asks, and I could not close the root cause tonight — so the playable state comes back first and the fix gets rebuilt against a proper harness. Pull whenever; you're back on round 4's behavior (exit flash at the wrong end, death bar forming low in frame — the known state you gave feedback on).

## What your five reports established (this is real evidence, kept)

- The board DATA is intact in every snapshot — pure presentation/camera loss.
- `00-04-31` (earliest): already broken — one floating sprite in grey void — and its plan is aimed at the now-empty tile, so the first battle had COMPLETED, not hung. The breakage survives a completed pass.
- `00-04-58`: after your reload (Noemie restored), the world is STILL broken before the next battle — so whatever leaks also survives `apply_scenario` + rebuild, which eliminates most of the staging/rig channels (those all have reload doors that verifiably fire).
- `00-05-24` (zoom −7): the world seen from underneath/through — proves the render itself is alive; the camera's resting pose is what is wrong.
- `00-05-41` → `00-05-46` (5s apart): the death burst's cubes appear between them and the bar-row assembles IN frame — bug 2 confirmed live, and also: my round-5 frame-edge projection was mathematically wrong anyway (it dropped the recession term — a point below the aim also moves away along the view axis; the true edge is `D·tan(fov/2)/(cos p − sin p·tan h)`, ~30% deeper).

## What I measured

- A headless probe driving your exact battle (the report's own `board.tres`, Isaac (10,-8) shoving Noemie (9,-8) into the pit) through the real `execute_orders` ends CLEAN — staging cleared, columns home, lock released. The fault lives only in the frame-domain interleaving (real timers, real tweens, real flight frames), which headless structurally escapes — the declared blind spot of this whole arc, now with teeth.
- One genuine race found while probing, present with or without round 5: the executor's transition await (a SceneTreeTimer) and `advance_flight`'s last landing cross `total` on the same frame; if the timer resumes first, the final tile's landing bump never fires and its column is dropped with the flight maps un-reseated. Real but small (one tile, one frame window) — not the whole picture your frames show.

## The rebuild plan

Round 5's two asks stand (the flash tied to the teleport; the bar forming off screen). Before re-landing them I will build a frame-domain harness — a suite that drives `_drive_transition` / `_sync_staging` / the executor's await ordering with controlled deltas against the real scene — so the entry/exit interleave is pinned by cases instead of play-checks, close the timer race above while in there, and use the corrected edge projection. That lands as a fresh round with its own writeup.

Sorry for the broken evening build — the revert is CI-watched and you are on `feat/602-cliff-follow` as always.

— Claude (Fable 5) · 2026-08-30

Phaazoid:
🤖 Claude says:

**Round 6 is pushed** — the two real defects the round-5 regression was sitting on, fixed in isolation with instruments, no behavior redesign. This is the first round of the plan you approved; the flash-rides-the-cut re-timing itself is deliberately NOT in it and goes back in a later round, alone.

## 1. The disappearing platforms: a race, now closed

Your five reports all showed the same thing — after the rise-up, the staged patch renders as void while the board data underneath is intact. The cause is a **frame-order race between two clocks measuring the same transition**: the executor awaits a `SceneTreeTimer` cut to the flight schedule's total, while `advance_flight` walks the same clock per rendered frame. Both cross the landing time on the same frame, and which resumes first is scheduler whim.

When the timer won, `end_flight_now()` cleared the flight **with no version bump**. The version-gated render had last drawn that tile's column as *in flight* (excluded from the seated board), the per-frame flight draw stopped with the flight, and with no bump nothing ever re-read — the column sat in neither draw. A hole where a platform stood, for the whole battle on an entry. (On the exit, `clear_staging()`'s own bump happened to repair it, which is why you only ever saw it after the rise-up.)

The fix is one rule: **ending a live flight is a discrete event the render must hear** — `_end_flight` bumps `staging_version` whenever it clears a live flight, whoever killed it. This finishes the #521 slice-B law (a tile *landing* bumps; now the flight *ending* does too).

Verified by an interleave harness that replays the exact race by hand: advance the flight to just short of total, assert the last tile still flies, end it through the timer's door, settle, assert every staged column is seated in the GridMap. Mutant (bump silenced) reds exactly that case.

## 2. The on-screen health bar: the frame edge was the frame centre

Round 4 anchored the burst 1.5 cells under the **rig's height — the frame's centre** — and at this game's narrow fov (30°) that is well inside the frame, which is why the bar assembled in shot. The true frame-bottom drop for a pitched camera is:

```
drop = D · tan(fov/2) / (cos p − sin p · tan(fov/2))
```

The recession term in the denominator is the whole correction — the frame's bottom edge lands on ground *farther from the camera* than the aim point. Round 5's formula (`D·tan(fov/2)/cos p`) forgot it and read ~30% shallow at your pitch, an "off screen" anchor still inside the frame. `CameraRig3D.frame_drop` is the pure math (3 aimed test cases, incl. a regression pin against the round-5 formula specifically — mutant-verified), and `frame_floor_depth()` feeds it the **live** camera (rendered distance, live fov, live pitch), because the frame you're watching is the only frame "off screen" can be measured against. `PLUMMET_BURST_UNDER` retuned 1.5 → 0.5: a margin *under the edge* now, not a stand-in for the edge.

## 3. Your reports diagnose themselves now

The `View:` line in every `report.md` now stamps the rig's channel state:

```
[lift 12.0, drop 5.0->5.0, dist->7.0, tscale 1.00, staged 84, flight no]
```

Live→target pairs wherever a channel eases, so a report says both where the camera IS and where it's trying to be. Round 5's forensics spent a session proving from pixels that the eases were alive and the staging was stuck — those are six numbers now, on the line the reporter already writes. If anything else goes wrong in this arc, your first report carries the diagnosis.

## Verification

- `tests/presentation` full area: **600/600 green**; `tests/law/test_board_scoped_state.gd` 3/3 (the new statics fork was the round-5 CI red — nothing new declared this round).
- **Mutant A** (silence the `_end_flight` bump): exactly `test_a_flight_ended_by_the_timer_first_still_seats_every_column` reds. Reverted, tree verified clean.
- **Mutant B** (swap `frame_drop` to the round-5 recession-less formula): exactly `test_a_pitched_frame_edge_is_deeper_than_the_formula_that_forgot_recession` reds. Reverted, tree verified clean.
- Not checked headless, by construction: the race itself only exists with rendered frames (headless collapses the transition to zero frames), which is why the harness drives the orderings by hand rather than waiting on real timers — the declared blind spot of this arc, now instrumented instead of ignored.

**Canon:** `visual-clarity.md` gains the Round 6 section (the flight-end law, the corrected edge math, the self-diagnosing View line); marker re-checked through #661.

**You are on `feat/602-cliff-follow` — pull and launch.** The knock-into-the-pit on the verticality board is the check: platforms should stay seated after the rise-up, and the health bar should assemble below frame and burst up into view. The white-out timing itself is unchanged from round 4 (the re-timing returns separately, on top of a board that provably seats its columns).

— Claude (Fable 5) · 2026-08-30

Phaazoid:
🤖 Claude says:

**Round 7 is pushed — step 2 of the rebuild: the flash-rides-the-cut design is re-landed**, whole, on top of round 6's race fix. With your play-check confirming the platforms stay seated, the evidence chain closed: round 5's breakage was the pre-existing flight-end race (round 6's Mutant A replayed round-5's exact `_end_flight` semantics and redded the harness's timer-first case), so the re-timing itself was never the culprit and comes back as one diff instead of five pieces with play-checks between.

## What's in it

**The white-out and the teleport are one schedule now** — your rule verbatim: *"the white out should always be tied to the last thing that happens, the teleport, no matter what."* The cut sits at the flash's first full-white frame, and the flash anchors to the transition's cut-ward end:

- **Exit** (both arms): tiles fall bare — you said you liked watching them — the flash ramps once the last one is home, and the camera drops under full white. No more flash-over-nothing at the start and bare 40-unit drop at the end.
- **Cut entry**: hold with the intact board while the flash ramps, teleport up at full white, fade reveals empty sky before the first tile arrives.
- **Travel entry**: unchanged — no cut to hide, the flash covers the eased rise after the camera hold.

The schedule (`flash_anchor` / `whiteout_level` / `cut_over` / `entry_total` / `exit_total`) lives in `StagingFlight` beside the travel maths, so the flash, the camera, and the executor's await read **one clock** — three derivations of one timeline is how the old exit came to flash over nothing. The drop is **snapped, not eased** (`cut_lift`): an eased teleport is still settling as the fade reveals it.

**And a cell lands ONCE.** The re-land bug round 5 found turns out to have been live in every exit the game has ever played, not just the flash tail: an exit's landed tile keeps holding its socket offset, so every frame between its landing and `total` erased-and-re-added it — a version bump and a whole-board prop rebuild **per frame** through the exit's whole landing window. The guard fixes that today, tail or no tail. (If exits have ever felt heavier than entries, this was why.)

**Not re-landed:** round 5's `frame_drop` — round 6's version with the recession term is correct and stands untouched.

## Verification

- `tests/presentation` full area: **607/607 green** (round 6's 600 + the 6 schedule cases + the lands-ONCE case); `tests/law/test_board_scoped_state.gd` 3/3 (the two returning statics re-declared).
- Round 6's interleave harness runs green **on top of the re-timing** — the raced entry end and the raced exit end both still seat every column with the extended awaits in place.
- **Three aimed mutants, each killed by exactly its case**: exit flash anchored at 0 (the old bug) → `test_the_exit_flash_is_dark_while_a_tile_still_flies` reds; entry cut at t=0 → `test_the_entry_cut_waits_for_full_white_too` reds; lands-ONCE guard removed → its case reds (plus five downstream exit cases, which is what removing that guard deserves). Each reverted, tree verified clean.
- Headless caveat as always: the schedule is pinned as arithmetic and the orderings by the harness; the *feel* of the timing is yours to judge in play.

One pre-existing housekeeping find, not from this diff: the presentation run prints 4 silent "String formatting error" push_errors from the CardSheet suite (present in round 6's CI too) — flagged as a separate task chip, since a swallowed diagnostic may be hiding a real message about the sage sheet.

**Canon:** `visual-clarity.md` gains the Round 7 section; marker stays #661.

**You are on `feat/602-cliff-follow` — pull and launch.** The checks: (1) end a battle — white ramps *after* the last block lands, and the fade reveals the camera already home; (2) start one — the teleport up happens at full white; (3) the pit knock — platforms seated, bar assembling below frame. All the flash/hold timings are the same knobs as before (`TEAR_OUT_WHITEOUT` / `TEAR_OUT_HOLD` on the Game tab) if the beat wants tuning.

— Claude (Fable 5) · 2026-08-30

Phaazoid:
🤖 Claude says:

**Round 8 is pushed — the death show owns the whole shot now, not just the depth.** Your ruling scoped it exactly: *"the camera should never look at where it forms, only the upward directed results of the explosion should be visible."*

Your report diagnosed itself, which is what the View line was built for. `drop 5.0->5.0` — round 4's depth hold, working. Beside it: `zoom 8.3, dist->11.0` — **the camera mid-dolly-out over the forming bar.** A farther camera's frame bottom is deeper (~0.45 units per unit of dolly at your fov/pitch), so the bar never moved; the frame's edge descended onto it. "Slightly off picture, but then the picture immediately looks at it" is that dolly, frame by frame.

Round 4 gated one camera channel on the show. Three movers never joined it — round 8 is those three joining:

1. **The trained release edge defers.** The instant the followed body was freed, the shot handed its distance back and dollied out from trained 7 toward 11. While the show is live, the edge holds the trained frame; it fires the moment the last cube lands — deferred, never skipped.
2. **The teardown waits before it pans.** At pass end, `_bring_the_board_home` panned back to the stage centre *first* and waited for the depth *second* — so the aim travelled and the shot re-framed while the cubes were still assembling. Now it releases the follow, waits out the held depth and the climb, and only then pans. The new shot: hold on the pit while the cubes erupt up into it, rise out, pull back wide. Your earlier "camera comes back to the battle view first" ruling keeps its scope — the pan and the climb both still land before the aftermath beat and the tiles going home; only their order between themselves swapped.
3. **The anchor measures the settled frame, not the live one.** A body can die while the drop is still easing toward its held depth, leaving the frame descending onto sockets that froze at spawn. `settled_frame_floor` composes from the rig's *targets* — the deepest this frame can ever get (the dolly only moves closer, the eases only approach) — so a socket anchored below that line can never be scrolled onto. The live `frame_floor_depth` lost its only caller and is deleted.

## Verification

- `tests/presentation` full area: **610/610 green** (round 7's 607 + three new cases).
- **Three aimed mutants, each killed by exactly its case**: the show gate deleted → the trained-frame-hold case reds; pan-before-wait restored → the ordering case reds (it drives the real `_bring_the_board_home` with a held depth and watches the 2D camera not move); settled → live channels → the targets-not-live case reds. Each reverted, tree verified clean.
- The invariant, stated once: **at no moment while a cube is still waiting can the frame's bottom edge reach the socket row** — the show freezes every channel that could deepen the frame, and the anchor sits under the deepest frame the frozen targets allow.

**Known residual, declared not built:** a *mid-pass* void death followed by more beats still pans *away* with cubes flying — the show plays off-screen. That doesn't violate the ruling (the camera never looks at the formation), so it's left as-is; say the word if it reads badly in play and it becomes its own ticket.

**Canon:** `visual-clarity.md` gains the Round 8 section; the round-6 paragraph is amended for the settled anchor.

**You are on `feat/602-cliff-follow` — pull and launch.** The pit knock: the camera should ride down, hold on the pit while the bar erupts up into view from below the frame, rise out, and only then pull back to the wide shot. At no point should the assembling grid be visible.

— Claude (Fable 5) · 2026-08-30

Issue #685, opened 2026-09-02
The action queue wants a visual redesign: an early hand-authored shape whose containers do not hold it


Phaazoid:
🤖 Claude says:

Filed out of [#659](https://github.com/Phaazoid/Godoiosis/issues/659), where the dev's own addendum ruled that the queue's shape wants a redesign rather than a scaled copy of what it has. **This ticket owes a grill before any plan** — it is a look, and a plan built off inferred taste is the wrong artifact.

## The dev's words (2026-08-29)

> Can we put an addendum on that ticket - the action queue could use a visual redesign in general. It's a very old shape - one I hand created pretty early on. those labels clipping? That happens normally, not just in this version. it needs some work.

## What #659 did and did not do

#659 gave the whole UI a design space, so the queue panel now **grows with the window** like everything else. It did not touch the panel's shape, and deliberately: the content scale is a viewport-level change that touches zero queue code, so it is not the "scale factor bolted onto the old shape" the addendum warns against.

## The clipping has a diagnosis — it is a broken container chain, not a scaling symptom

Worth writing down before the redesign, because it explains *why* the shape reads as thin and why the `MOVE` / `ATTACK` labels clip at any resolution:

`Scenes/SquadActionQueueControl.tscn`'s `MarginContainer` sits at **absolute offsets** (`layout_mode = 0`, `offset_top 10 → offset_right 64, offset_bottom 42`) inside a `BackgroundPanel` that is 145×465. So its `VBox` is **64px wide and 32px tall** while its children are far bigger — `ActionQueueRow` has a 128px minimum, `ExecuteButton` a 145×36 one, and `OuterScroll` a 410px minimum height. Everything overflows its own box, and the section headers clip because they are laid out in a 64px column.

A redesign that goes through proper containers (`PanelContainer` → `MarginContainer` → `VBox`, all full-rect) fixes the clipping as a side effect of being built correctly. The point of naming it here is that **nobody should "fix the clipping" as a separate small ticket** — it is the same visit.

## Two shape constraints already in the tracker

Whatever shape this lands on has to hold these, or it gets redone again:

- **[#462](https://github.com/Phaazoid/Godoiosis/issues/462)** — show which step of a running pass the queue is on. That is a progress indicator living *in* this panel.
- **[#119](https://github.com/Phaazoid/Godoiosis/issues/119)** — battle log, i.e. scrolling back through past action queues. Same widget, different tense.

## Canon that constrains the CONTENT (not the chrome)

[...]

Issue #659, opened 2026-08-29
UI doesn't scale with window size: inspect panel and action queue stay fixed-pixel, tiny on 4K


Phaazoid:
🤖 Claude says:

Filed by Claude at the dev's request, from two of his Discord bug reports — not written by the dev directly.

## Symptom

Run the game, fullscreen it, and check Godot's **fit aspect ratio** — the 3D board stretches to fill the screen correctly, but the UI does not scale with it. Panels keep their design-resolution pixel sizes, so on a larger window they occupy a shrinking sliver of the screen. On the dev's display it is tolerable; on the codev's **4K display the panels were downright tiny/thin**.

## Evidence

Two bug reports from 2026-08-29, both `v0.116.1 -- feat/602-cliff-follow @ c5df491`, window ~2346x1320 (source of record: `#feedback-reports`; local copies under `user://reports/<stamp>/`):

- **`2026-08-29_16-12-58`** — the **inspect panel**: a narrow fixed-width column pinned to the left edge. HP/WIL bars, the stat grid, and the loadout slots all stay at design-resolution size while the board fills the rest of the width.
- **`2026-08-29_16-13-12`** — the **action queue**: the top-right panel holds its design-resolution width while the window grows around it, leaving a thin strip whose order icons barely fit the row.
  - **Amended 2026-08-29:** this originally cited the clipped `MOVE` / `ATTACK` labels as evidence. That was wrong — per the dev, those clip at his normal resolution too, so they are a pre-existing defect of the panel's layout rather than a scaling symptom. See the addendum comment below.

(Screenshots not linked here — Discord CDN attachment URLs carry a signed expiry and rot. Sent to the dev directly; the originals in `#feedback-reports` are the source of record.)

## Mechanism

This is the defect class #133 predicted, now observed for real: the game renders through a SubViewport with **stretch mode disabled**, so UI is real pixels — anchored panels keep fixed pixel offsets and simply don't grow when the window does. Same family as the documented `BOTTOM_LEFT_POS = 410` hover-panel bite (the "no hardcoded pixel positions" sharp edge), one level up: the positions are anchored, but the *sizes* are still design-resolution pixels.

## The ask

The inspect panel and action queue (and by implication the rest of the battle UI) should **grow with bigger screens** — a 4K player should get the same relative panel footprint the dev sees at design resolution.

[...]

PR #688, opened 2026-09-02
#659 follow-up: the .uid sidecar the commit forgot, and the guard for it

Phaazoid:
Follow-up to #684 (merged). Two files and a workflow step.

## The defect

#684 added `tests/ui/test_ui_scales_with_the_window.gd` **without its `.uid` sidecar.** The import that would have minted it ran *before* that file was written — correctly, so `GameSurface`'s `class_name` would register — and never ran again before the commit.

It was the only `.gd` in the repo outside `addons/` without a tracked sidecar, and the symptom is the one you ruled a ticket defect at both ends: **your tree went dirty by itself** the moment the editor opened it. Committed at the exact uid your editor already minted (`uid://diev6f1c0uf3w`), so the untracked copy sitting in your tree now goes **clean on pull** rather than conflicting.

## Why nothing caught it

- The post-import guard already in the workflow diffs `project.godot` — a **tracked** file. A missing sidecar arrives **untracked**, so `git diff` never sees it.
- A gdUnit assertion could not have caught it either: CI imports before tests run, so the file exists by the time anything looks. It would pass vacuously in exactly the case it existed for.

So the guard is a CI step on `git ls-files --others --exclude-standard`, placed after the import and before the test run. It **fails** rather than warns, unlike its neighbour — that one's cause is an addon upgrade unrelated to whichever PR happens to hit it, while this one's cause is always the PR in hand.

**Falsified both ways** in a fresh worktree off `main`:

| | result |
|---|---|
| import on `origin/main`, sidecar absent | regenerates **exactly** that one file, nothing else — so this would have reddened #684 |
| same, once the sidecar is staged | silent — no false positive on a clean tree |

## Also

`docs/design/grill-queue.md` gains #685 (the action queue's redesign). That list aggregates from many source docs and its documented failure mode is going stale from *other* tickets' work — a ticket that explicitly owes a grill has to appear on it. Numbered **17**, not 16: quick hit 16 is a resolved item and reusing the number would read as a reopen. Marker bumped.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -99,6 +99,25 @@ jobs:
             git checkout -- project.godot
           fi
 
+      # AN IMPORT MUST NOT PRODUCE FILES THE COMMIT WAS MISSING (#659 follow-up). Godot mints a
+      # .uid sidecar beside every script the first time it imports one, so a commit that adds a
+      # .gd without its sidecar leaves every clone -- the dev's included -- going dirty by itself
+      # the moment the editor opens. That is the "a dirty tree is a ticket defect at both ends"
+      # rule, and nothing was watching for it: the guard above diffs project.godot alone, which
+      # is a TRACKED file, while a missing sidecar arrives UNTRACKED and shows up in neither
+      # `git diff` nor the test run (the import has already created it by the time gdUnit4 looks,
+      # so an on-disk assertion would pass vacuously in exactly the case it exists to catch).
+      #
+      # Fails rather than warns, unlike its neighbour: this one's cause is always the PR in hand.
+      - name: Guard against sidecars the commit forgot
+        run: |
+          untracked="$(git ls-files --others --exclude-standard)"
+          if [ -n "$untracked" ]; then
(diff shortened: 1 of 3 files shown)


Phaazoid:
---

**Update:** the guard's first cut asked `git ls-files --others` with **no path filter** and reddened every shard immediately — this workflow downloads Godot to `godot-bin/` *inside* the working tree, untracked and not ignored, so "anything untracked" means "the engine binary" on every run. Now scoped to `*.uid` and `*.import`.

My falsification was the wrong shape rather than absent: I ran both controls, but in a local worktree that has no `godot-bin/`, because I run Godot from outside the repo. That tested the guard's logic against a tree CI never has. Re-falsified against CI's actual condition — with `godot-bin/godot` present the narrow form is silent while the old form reports it, and with the sidecar un-staged it still fires. Narrowing is the better rule regardless: it names the thing being guarded, covers `.import` for the same failure on assets, and cannot be broken again by whatever the workflow drops in the tree next.

All five shards green.

Issue #117, opened 2026-07-30
AI catch-up: capability gaps and open design questions (evergreen)


Phaazoid:
🤖 Claude says:

**Evergreen tracker — do not close.** The AI is permanently behind the feature set, and always will be: every system we add creates AI work that lands *after* the system ships. This is where that debt lives, so it stops being rediscovered one bug at a time.

Scope: the archetype AI in `Classes/ai/` — what it can't see, can't do, or does badly. Canon is [`docs/design/ai-tactics.md`](https://github.com/Phaazoid/Godoiosis/blob/main/docs/design/ai-tactics.md). Genuine *bugs* (a soft-lock, a crash, a rule violation) still get their own issue; this tracks capability gaps and design questions. Tick items off in place and add new ones as features land.

## The integration contract (why this list should stay short)

From #78's ratification: **content-shaped changes reach the AI with zero AI-side wiring.** A new weapon family, carving, mod, readiness flag or strain rule should be visible to the AI for free, because the AI walks the player's own surface — `get_selectable_attacks` → `active_attack` → `AttackAction.declare()` → `queue_action`. **If the AI can't see a feature, the feature bypassed the player surface, and the feature is what to fix.**

The Carbine (#84) is the model: zero AI work, because it rode an existing declared verb (`RELOAD`, already in every archetype's `MAIN_ACTION_PRIORITY`) plus the `is_attack_fireable` filter the chooser already applies. The AI reloads a dry carbine on day one.

So an entry here means one of two things: a verb that needs a *scored builder* (a real decision, not wiring), or a genuine design question about how the AI should think.

## Open: mechanics the AI cannot use

- [x] ~~**BURROW** sits in `AIArchetype.MAIN_ACTION_NEVER` for every archetype with no scored builder.~~ **CLOSED by #726 (2026-09-03):** `AITactics._try_burrow` mirrors the rev pair, Hold/Sentry declare BURROW *and* REV (Rushdown keeps BURROW `NEVER` as a declaration — a rusher does not entrench), and `DrillWeaponRoutine` refuses to re-dig a covered cell, read at the projected destination where the deposit lands. A Sentry at its post takes the fallback walk now, so it entrenches, revs and reloads there — it had queued nothing at all before. *(This item said "Rev and Burrow" until 2026-09-02 and was stale: `AITactics._try_rev` shipped 2026-08-06.)*
[...]

Issue #57, opened 2026-07-14
Build 8: AI Crisis stances + lethality preview


Phaazoid:
🤖 Claude says:

**v2 build spine — session 8 of 13** · size **S** (palate-cleanser; slot wherever a short session fits, anytime after 6). Prompt: [docs/session-prompts/8-ai-crisis-and-preview.md](https://github.com/Phaazoid/Godoiosis/blob/main/docs/session-prompts/8-ai-crisis-and-preview.md)

**Builds:** per-archetype AI Crisis stances (RUSHDOWN always / HOLD-SENTRY never), the crisis-aware lethality preview (Law #2), AI deprioritizes downed units (the #29 leftover).

**Canon:** [will-and-death.md](https://github.com/Phaazoid/Godoiosis/blob/main/docs/design/will-and-death.md) (AI Crisis stances, resolved 2026-07-04) — co-dev ratified 2026-07-14.

**Spine position:** after #6, otherwise free-floating.

— Claude (Fable 5) · 2026-07-14

Phaazoid:
🤖 Claude says:

Code landed and committed (`7a90725`). Every archetype declares a Crisis stance on the registry (`AIArchetype.CRISIS_STANCES`: RUSHDOWN always accepts, HOLD/SENTRY never), and `Unit.is_crisis_eligible()` is now faction-agnostic — worth flagging: **enemy Crisis was actually unreachable before this build.** The shipped "auto-accept" gate lived on `is_crisis_eligible`, which was PLAYER-faction-only, so the auto-accept path in `_offer_crisis` could only ever fire for an AI-driven *player* faction. This build is the first time enemy Crisis genuinely turns on. Nothing on your current roster changes behavior though — eligibility needs WIL 20 and no authored enemy carries it yet (BadGuy1 sits at 4), so it stays dormant until you author a high-WIL enemy. That's the intended balance lever, not a gap.

The queue preview is now honest about it: a would-be-down on a full-Will, Crisis-accepting enemy shows a distinct CRISIS icon (new asset, `Art/Icons/CrisisIcon.png`) with its own summary text instead of the old DOWNED icon lying about it, and a second lethal hit against a unit already in Crisis correctly previews KILLED (no safety net, per the design doc). `PlanResolver._predict_lethality` also got refactored to take the threaded hypo as one argument instead of eight positional params — that was the exact shape that caused a compile error last session, so this closes the risk permanently.

[...]

PR #706, opened 2026-09-02
AI: score a candidate against the squad's plan, and pick attacks jointly (#117)

Phaazoid:
Slice 1 of the AI catch-up ([#117](https://github.com/Phaazoid/Godoiosis/issues/117)): **a candidate is scored by what it adds to the squad's plan, and the squad's attacks are chosen together.**

## What was wrong

`AITactics` scored every candidate on a throwaway plan holding **one volley, against the live board**, and walked members in fixed order. A member could not see what its squadmates had committed to, which made three things impossible rather than merely unwise:

- two members overkill one target while a second enemy goes untouched — visible on the shipped Castle Assault board, where `Soldier23` and `Soldier24` both aimed `(3,0)`;
- a finishing blow looked like any other hit (`OVERKILL_CEILING = 10`, so every kill is down-then-finish);
- **the AI could not open a combo at all** — `Splash.tres` is `deals_no_damage`, so soaking a target scored `(0,0)` and the bar is *beat* `(0,0)`.

## The seam

`SquadManager.resolve_plan` splits over an action array (`_resolve_actions`), so `resolve_hypothetical` runs the **same pass** over "the queue as it would be with this in it" — `_hypothetical_actions`, already the queue gate's builder, now taking a list so the lookahead can ask about a pair. The split deliberately writes no cache: `resolved_plan_for` must never hand a hypothetical to the rescue clause.

## Rulings taken during the build

- **A reaction's damage is not scored, only its removals** (2026-09-02). Priced at par a counter cancels exactly — two units with the same weapon trade 3 for 3 — so every even exchange scores `(0,0)` and a squad facing mirror-statted enemies declines every attack and reloads. Removals are the currency, damage the tie-break; a counter that *fells* one of ours counts, a scratch is the price of engaging. Worth knowing before re-opening: C1/C4 make the counter a **per-plan fixed cost the opener bears alone**, so at par the AI would never open on a multi-member squad at all.
- **Pairs in v1** (2026-09-02). A candidate scoring nothing alone that puts a state on an enemy is priced by its best squadmate follow-up.

## Two things the mutants found, not the tests

- **Every knockback attack was unqueueable by the joint pass.** `queue_action`'s whiff gate is the one reader of published knockback and is correct only because that knockback belongs to the *already-queued* aims. A scoring pass publishes a candidate's shove too, so the gate hunted for the target on a cell a rejected hypothetical had thrown it to and refused the winner as a whiff. The pass now restores the real plan **before** it queues. No fixture with a plain weapon could see it.
- **The projections case was vacuous twice.** First draft had no shove to publish and no refusal to end on; the rewrite needed both.

## Cost

`tools/profile_ai_turn.gd` (new), Castle Assault, every AI squad decides once:

| | before | after |
|---|---|---|
| whole board | 739.65 ms | **578.11 ms** |

Faster despite far more resolves, because both reaction catalogs dir-scan per call and are now hoisted out of the scoring loop.

## How to check this

- **What to do:** load Castle Assault or Prolog with enemy AI on, put two of your units in reach of an enemy squad, and end turn. Headless twin: `play_drive.ps1` → `endturn`, read `ai_events`.
- **What you should see:** an enemy squad that spreads onto a second target rather than piling onto one it already downs, and that focuses when a second hit actually secures a down. If an enemy alchemist carries a WATER carving and a squadmate a SHOCK one, the soak queues *first* and the shock pops *Electrocuted!*.
- **What a failure looks like:** two enemies aiming a body that is already down; an enemy squad that reloads instead of attacking (that would be the counter ruling coming out wrong); or a ghost shove preview sitting on a unit nobody hit (a restore failing).
- **What the tests already cover:** `tests/ai` 97, `tests/squad` 240, `tests/law` 201, `tests/play` 65 — all green. Ten new cases in `tests/ai/test_squad_scoring.gd`, falsified against six mutants (overkill clamp, per-victim removals, reaction ledger, lookahead, pre-queue restore, cache write). **Declared blind spots:** the suite cannot see *feel* — whether "declines a bad trade" reads as smart or passive is yours to play. And the **trailing** restore is declared unpinned: no case reds when it is deleted, because everything downstream happens to re-resolve; it is kept for what it means, and said so at the line.
- **Still stacked on Castle Assault:** `Soldier23`/`Soldier24` both aim `(3,0)` in the profile above. That is now a *measured* decision — the overkill clamp says the second hit still lands real damage on a survivor — rather than the old blind tie. Worth an eye in play.


Phaazoid:
**Added 2026-09-02 — the felling-beats-finishing ruling.**

> "AI will always prioritize downing new units over finishing off already downed units, or getting overkill… this is important to reward the player for using squads correctly." — and, to be clear, *when choosing between things in their range*: they finish a downed person if that is their only option, but choose a new person over a body whenever both are reachable.

Checked before building: #57's two passes already gave that as a **hard precedence, not a weight** — a body is neither aimed at nor scored while anyone is upright in reach — and the overkill half falls out of the clamp. Two clauses were needed to make it honest, one found by review:

- **Pass 2 needs an EMPTY pass 1, not a losing one.** It opened whenever pass 1 produced nothing *queueable*, so a member whose only standing target was a bad trade (a lethal counter) went and executed a downed unit instead — the moment the body most wants sparing. `considered` separates *nobody to fight* from *nobody worth fighting*; a **refused** candidate deliberately does not count, since the queue gate turning it down means it was never a real option.
- **Pass 1 means "still standing once the plan so far has run"**, not "standing right now" — Fable's catch on review. Planning does not execute, so a unit a squadmate felled *this round* is still `is_active()`; read live it counts as a standing target and wedges pass 2 shut, leaving the last member idling beside an old body it should finish. `_attack_candidates` asks `base_plan.hypo` instead, which is why `_best_candidate_for`/`_lookahead` now take the base **plan** rather than its score — one source for both facts. It only ever drops a candidate the clamp was going to score `(0,0)`, so nothing else moves, and it costs one fewer resolve per round.

Also fixed in passing: the base score is now measured on the **same `include_downed` terms as the candidate**, or a pass-2 marginal double-counts damage an earlier round already put on a downed enemy.

`tests/ai` 99, `tests/squad` 240, `tests/law` 201, `tests/play` 65. Two new cases, one per clause, each falsified against its own mutant — and the four earlier mutants (clamp, reactions, lookahead, pre-queue restore) all still kill.

## How to check this (updated)

- **What to do:** as before, plus — leave one of your units downed next to an enemy squad and end turn.
- **What you should see:** enemies engage anyone still on their feet before touching the body, and only finish it when there is genuinely nobody standing in their reach. An enemy facing a bad trade idles rather than executing your downed unit.
- **What a failure looks like:** a body finished while an enemy stood next to one of your live units; or an enemy reloading beside a finishable body with nothing else in reach (that would be the second clause reading the live board).
- **What the tests already cover:** both clauses, mutant-checked. Still blind to *feel* — whether the AI now reads as too passive when it declines a trade is yours to play.

PR #717, opened 2026-09-03
A body in the blast is a bonus, not a reason to aim elsewhere (#716)

### LINKED ISSUE #716: AI declines an AoE that would also finish a corpse, preferring the shot that spares it
🤖 Claude says:

**Found in play (Castle Assault, 2026-09-03), diagnosed to the layer, fixed.**

> *"There was an instance where a unit (the general in the front row) with an AoE attack had two options where he could hit enemies. One of his options would hit and kill a downed enemy, and hit an upright enemy. His other option would just hit the upright enemy. He chose the option to only hit the upright enemy. This feels like incorrect behavior... It felt like the AI was playing for the wrong team."*
>
> *"I think we should have the rules change so that while AI will prioritize units who are still standing, they will not de-prioritize otherwise better attacks just because downed units are in them."*
>
> — dev, 2026-09-03

## The cause, and it was a duplicated rule

`AITactics._score_plan` skipped damage on an already-downed victim whenever `count_downed` was false, i.e. during #57's pass 1:

```gdscript
if victim.is_downed() and not count_downed:
    continue   # already on the ground before this plan -- #57, pass 1
```

Both of his options aim at the **upright** enemy, so the corpse only ever appears as splash. With that skip the two candidates priced identically, and the tie fell to candidate order.

**The skip was a SECOND spelling of #57's precedence.** `_attack_candidates` already refuses to *aim* at a body while anyone is standing in reach — that is the whole rule, and it is untouched. All the skip added was erasing the incidental value of an attack that catches a corpse on the way to a standing target, which is exactly the behaviour he is objecting to. Law #4: one question, one answer.

## The fix

The skip is deleted, which makes `count_downed` a dead parameter, which in turn lets `_best_candidate_for` compute **one base for both passes** — they now differ only in which enemies get AIMED at, never in how the resulting plan is priced.

**The magnitude is self-limiting rather than tuned.** A downed unit clings at 1 HP (`Unit._go_downed`), so the overkill clamp values finishing one at exactly **+1 damage** — enough to break his tie, never enough to outrank felling someone on their feet. `_plan_removes` still answers false for a body, so an incidental corpse cannot earn a removal either. The precedence survives intact and is still pinned by its own cases.

## Verified

A regression case builds his exact shape — one attacker, two attacks of identical power, an upright enemy adjacent and a body directly behind it, so one option catches both and the other does not. It asserts the aim stays on the standing enemy (or the case would be measuring #57 rather than this) and that the reaching attack wins. Restoring the skip reds it with *"the AI took the shot that spares the corpse, over the identical one that also finishes it"* — and reds an existing case besides, so the deletion is load-bearing in two places.

`tests/ai` + `law` + `play` 375, green.

**Stacked on #711** (PR #713): both edit `AITactics`' scoring, and the dead parameter's call sites are lines that ticket also moves. Merge #713 first.

Phaazoid:
Closes #716. **Stacked on #713 — merge that first.** Base is `ai/always-attack`, so this PR's diff shows only its own two commits.

> *"He chose the option to only hit the upright enemy... It felt like the AI was playing for the wrong team."* — dev, 2026-09-03, from a Castle Assault playthrough

## The cause was a duplicated rule, not a weight

`_score_plan` skipped damage on an already-downed victim during #57's pass 1. Both of his options aim at the **upright** enemy, so the corpse only ever appeared as splash — with that skip the two candidates priced identically and the tie fell to candidate order.

**That skip was a second spelling of the precedence it cited.** `_attack_candidates` already refuses to *aim* at a body while anyone is standing in reach, and that is the whole of #57. All the skip added was erasing the incidental value of an attack that catches a corpse on the way to a standing target — exactly what you objected to.

Deleting it makes `count_downed` a dead parameter, so that goes too, which lets `_best_candidate_for` compute **one base for both passes**: they now differ only in which enemies get aimed at, never in how the resulting plan is priced.

## Why this needs no knob

A downed unit clings at 1 HP (`Unit._go_downed`), so the overkill clamp values finishing one at exactly **+1 damage** — enough to break your tie, never enough to outrank felling someone on their feet. `_plan_removes` still answers false for a body, so an incidental corpse cannot earn a removal either. The precedence survives untouched.

## How to check this

**Set up the shape you saw:** a unit with an AoE, an upright enemy in reach, and a downed one positioned so that one aim catches both and another catches only the upright.

- **You should see:** it takes the shot that catches both.
- **A failure looks like:** it still picks the shot that spares the corpse — or, the new failure to watch for, it starts *aiming at* bodies while someone is standing, which would mean #57's precedence broke rather than this being fixed. The regression case asserts the aim stays on the standing enemy for exactly that reason.

**What the tests cover, and what they don't.** One case builds your shape directly: one attacker, two attacks of identical power, upright enemy adjacent, body directly behind. Restoring the skip reds it with *"the AI took the shot that spares the corpse, over the identical one that also finishes it"* — and reds an existing case besides, so the deletion is load-bearing in two places.

The blind spot is that both are single-attacker fixtures. Whether the **squad** now prefers blast lines that sweep the field is unmeasured, and the +1 is small enough that I would not expect it to reshape group behaviour — but that is reasoning, not evidence, and it is the squad-fixture gap again.

`tests/ai` + `law` + `play` 375, green.


Phaazoid:
🤖 Claude says: **this PR has no CI, and that is structural rather than a failure.**

`.github/workflows/tests.yml` triggers on `push: branches: [main]` and `pull_request: branches: [main]`. This PR targets `ai/always-attack`, so neither fires — the checks list is empty because nothing was ever asked to run, not because anything went red.

It resolves itself: when #713 merges, GitHub retargets this PR's base to `main` automatically and the next push runs the full suite. **Do not merge this one before that happens** — it would land unverified by CI.

What was run locally in the meantime: `tests/ai` + `tests/law` + `tests/play`, 375 cases, green, plus a mutant that reproduces the reported behaviour (restoring the deleted skip makes the AI take the shot that spares the corpse again).

Issue #675, opened 2026-09-01
Colour-blind mode: a palette over the whole board-markup vocabulary


Phaazoid:
🤖 Claude says:

*(Split out of [#422](https://github.com/Phaazoid/Godoiosis/issues/422) at the dev's request while its aim palette was being built, 2026-09-01. He set the priority: P3.)*

## The ask

Real colour-blindness support: a palette over **the whole board-markup vocabulary**, not the three channels an aim wears.

#422 shipped the mechanism and pointed it at the aim alone (dev scope call): the player picks a named palette, `PlayerSettings.AIM_PALETTE` stores which, and `OverlayManager.AIM_PALETTES` authors the alternatives. What it deliberately did NOT do is widen that vocabulary — so a player on the colour-blind palette gets a re-hued attack/heal reach and aim footprint, and everything else on the board stays exactly as authored.

## What is still red-against-green out there

The reach fork is the loudest case and #422 fixed it, but it is not the only one. The `Board markup colours` group in the Game tab holds ~20 rows, and the ones that carry MEANING by hue include:

- the move / invalid-move fills;
- the three planned-move arrow tints — cyan a queued move, **red** a refused one, **green** a member falling behind ([`OverlayManager._arrow_modulate`](Classes/board/OverlayManager.gd)), which is a three-way red/green/cyan distinction doing real work;
- the capture and extraction zone fills;
- the squad and squad-range fills;
- the guard shield / link pair, and the watch mark.

## The one concrete finding to inherit

**The watch aim is a measured near-collision, not a hypothetical.** #422 left `WATCH_REACH_MODULATE` outside the palette (correctly — it was scoped out), and pinned that so a half-widening is loud. But the watch's red-orange `Color(1.0, 0.35, 0.25)` and the colour-blind palette's vermillion `Color(0.835, 0.369, 0.0)` are neighbours, where against the DEFAULT red they had been deliberately tuned apart. Worth a look in play before deciding anything else here.

`tests/ui/test_aim_palette.gd`'s `test_a_watch_aim_is_untouched_by_the_palette` is the case that should be **deleted on purpose** when this lands, not worked around.

## What the mechanism already gives you

Nothing structural is missing, which is why this is a widening rather than a design:

- `PlayerSettings.AIM_PALETTE` is one ordinary choice row; adding options is a table line.
[...]

PR #729, opened 2026-09-04
#685: the action queue's chrome, and elemental effects that can be read

### LINKED ISSUE #685: The action queue wants a visual redesign: an early hand-authored shape whose containers do not hold it

Phaazoid:
Closes #685.

The action queue was the oldest hand-authored UI in the game, and its two defects were one defect at two scales.

**Structurally**, `BackgroundPanel/MarginContainer` sat at absolute offsets — a ~64×32 box inside a 145×465 panel — so every child overflowed the container meant to hold it. That is what clipped the `MOVE` / `ATTACK` headers at every resolution, which is why #659 could rule it pre-existing rather than a scaling symptom.

**Per row**, `_overlay_behind()` stacked four facts into three 32px squares through `show_behind_parent`: the crown under the actor, elemental state icons under the verb, the `hp→hp` readout *over* that same verb, and the fired reaction's art under the target. The elemental half was drawn and effectively invisible — your words: *"the elemental sprites being behind the items in the row they effect doesn't really do the trick."*

## The three rulings this was built to

Taken in chat against a rendered mockup, before any code. Recorded on #685.

1. **Palette → SLATE.** `Scenes/UnitInfoPanel.tscn`'s authored tokens verbatim rather than a third palette — the inspect panel is docked on the opposite edge of the same screen, so matching it collapses three disagreeing palettes into one, and dark chrome holds contrast over a lit 3D board.
2. **One line per order, panel 145 → 216 design-px.** The two-line row was rejected on the count it costs: a section would show three orders where it shows five.
3. **Elemental split BY WEIGHT.** A state *applied* gets a compact chip; a reaction that *fired* gets its own named line carrying its own authored `popup` — a word the queue has never shown at all.

**One deviation from the mockup, flagged before building:** the chips moved from the main line down to the consequence line. At 216px the line has ~59px left for the readout after three 32px sprites, the X and the gaps; one chip eats 22 of that and two overflow — i.e. it clips, which is the defect this ticket exists to remove. Only rows with elemental consequences grow a second line; a plain sword swing stays one line.

## Why "fun border effects" was the wrong answer

Your own suggestion, and it collides: a row **already** colours itself to say *this order is refused*, and one motif per meaning is `visual-clarity.md`'s Principle 2. So the allocation is **border = validity, rail = element** — a separate 3px channel that can sit beside a red border without arguing with it. Validity also gained a real predicate on the way (`BaseAction.is_refused()`, with `get_ui_modulate()` derived from it), so the border and the icon tint cannot disagree.

## Two facts became RECORDED rather than derived

- **`ResolvedOutcome.elements`** — the *surviving* set, stamped where the resolver already computes it. `PlanResolver._source_elements` is private and answers the **pre**-insulation question, so a rail read off the authored attack would claim a fire hit the target shrugged off. An insulated hit honestly wears no rail.
- **`ResolvedOutcome.fired_reactions` replaces `reaction_icons`** (Law #4). The row needs three facts per reaction — its word, its art, and its element — and the reaction holds all three, so recording the icon alone made the other two unrecoverable. Three sites in the whole repo. **Its producer had also been dropping any reaction with no `icon`**, which silently muted half the authored catalog (three of six reactions have a popup and no art); that guard went with it.

`outcome.popups` also carries `"Fell 2!"`, `"Drowning!"`, `"Into the void!"` and `"Insulated!"` — consequences the resolver has always recorded and the queue has never spelled out. They render now as neutral pills.

## Element colour is a first in the project

Nothing answered *what colour is Fire* — the board's own forks are keyed on heals-vs-damage and watch-vs-attack, never on `Elemental.Element`. `Classes/ui/ElementPalette.gd` is `StateIcons`' shape for colour instead of art, with **seven `GameKnobs` rows on the Elemental tab**. Deliberately not part of #422's aim palette; widening that vocabulary is #675's, at P3.

The queue's own chrome (`QueueStyle`) gets **no knobs, deliberately** — it is a copy of a neighbouring panel that has none either, so a knob would be a second place to tune one look. Say the word after a play-check and each is one table line.

## How to check this

**What to do**

1. Launch, start a mission, select a squad and queue a few orders — a move, an attack, and an attack whose target is already Wet if you can arrange it (a Water hit then a Shock hit on the same enemy is the easy path).
2. Open the dev tools → **Game → Elemental → Element colours** and drag the seven sliders while the queue is on screen.
3. Drag a row within its section to reorder it, and hover rows.

**What you should see**

- A slate panel matching the left inspect column, an `ORDERS` title strip, and each section (`MOVE`, `ATTACK`, `END OF TURN`) as its own bordered card with a tinted header — headers no longer clipped.
- Each row as its own bordered card, with a coloured **rail** down its left edge naming the element the hit carries, and every fact in its own slot: nothing drawn over anything.
- A Water hit shows a **Wet** chip; the Shock follow-up shows **Electrocuted!** as its own coloured line. Hovering either explains it.
- A refused order's row goes **red-bordered** — and its rail stays whatever colour its element is.
- Dragging the sliders repaints the rows live.

**What a failure looks like**

- Text running out of its box anywhere (the clipping is the whole ticket).
- A row whose consequence line wraps oddly or pushes the X off the edge — that is the width arithmetic being wrong, and case 7 below is what should have caught it.
- A slider that moves and changes nothing on screen (a born-dead knob).
- Drag-reorder silently doing nothing — that is the failure mode this redesign was most at risk of, since the drag walks the row's parent chain by hop count.

**What the tests already cover, and what they cannot**

`tests/ui/test_queue_row_readout.gd` is new (8 cases) and each pins a rule rather than a look: nothing shares a slot; a chip can speak (non-empty tooltip); a fired combo says its authored word in its element's colour; a **setup** is said once by its chip and not again as a line; the rail reads the *resolved* element and a non-elemental row stays neutral; every element and state resolves to a colour; and the widest row still fits its section. **No case asserts what a colour is** — the seven are knobs you drag, so pinning a hex would red the suite the first time you tuned one.

Falsified against four mutants, each red on the right case: deleting `outcome.elements` (rail case); re-parenting a Label into the action icon (overlap case); letting a setup reaction stop claiming its popup (setup case — and it names the symptom, `entries were ["Chilled", "Chilled!"]`); and pointing a knob's write arm at the wrong static (`test_game_knobs`'s existing no-placebo law).

Ran locally: **`ui squad presentation elemental dev` — 1674/1674 across 148 suites, 446s.**

**What the suite structurally cannot see, and is yours to play:** whether it *looks* right. I don't launch the game. Specifically — the 216px dock's footprint over the board, whether a section still shows enough orders at the new row height, whether the seven element colours read against the slate ground **and against each other**, and whether the consequence line is the right weight or too loud.

## Declared gaps

A **heal** short-circuits above the elemental stage, and a **tile hit** is built by `TileHitAction.make` rather than `_resolve_one` — so neither records `elements` and both wear the neutral rail. Stated so it does not read as an oversight; each is one line if content ever wants it.

**Embedded-content sweep:** nothing embeds `ResolvedOutcome` — it is a `RefCounted` built per resolve, never serialized — so no `.tres` fixture predates the two new fields. `ElementalReaction` gained a *method*, not an authored field, so the six `Resources/Reactions/*.tres` are untouched.

## Merge note

Rebased onto `main` after #727 landed. The code auto-merged (its change is in `_render`/`clear()`'s visibility gate, mine in the header, `_ready`, `_start_section` and `_new_row`); I verified both features survived rather than trusting the clean merge. The one conflict was `visual-clarity.md`'s canon marker, resolved to #727.


Phaazoid:
🤖 Claude says:

**Round 2 — all six corrections from the 2026-09-03 play-check.** Four of them named something structurally wrong rather than mis-tuned, so they are recorded as rules in `visual-clarity.md` → *Round 2*, not as a changelog.

| Ask | What was actually wrong |
|---|---|
| Rows too dark, don't stand out | The row was `0.137` against a `0.09` section card — a step of nothing. **The section is the gutter and the row is the card lying on it**, so the contrast belongs on the row (`0.23`), not in a darker section. |
| Rows too tall; shrink icons to 16; cut the crown | The action art is **authored at 16 and was being upscaled 2×**, so the height bought no detail at all. Also found: the row carried a `custom_minimum_size` height floor of **34**, which would have held rows tall whatever the slots did. |
| Stretch to fit, don't scroll early | Each section owned its own `ScrollContainer` capped at `SECTION_MAX_HEIGHT = 160`, so a section scrolled while the dock still had ~300px free. Sections take their natural height now; only `OuterScroll` catches overflow. |
| Status shows under the applier | The consequence line started at the row's left edge. Not a layout preference — **a chip under the attacker is a lie about who is wet.** Indented to start under `TargetTexture`. |
| Numbers need a card | Same `QueueStyle.tint_box` the chips wear, team-coloured, hidden on rows with no number. |
| Execute doesn't stand out | It had **no stylebox at all** and was drawing the engine's default grey on a grey panel. Crimson now; the disabled look is still `EXECUTE_DULL`'s modulate over it rather than a second style to keep in sync. |

Title reads **Action Queue**.

Deleting the per-section cap also deleted `_section_scrolls` and `_render`'s measure-after-a-frame pass — a per-section cap was a second answer to *how tall may this get*, and the panel's own height was always the first one.

## The trap the readout card set on the way in

Worth knowing before any future slot is added to this row: the card carried the line's horizontal expand, so a row that **hides** it — a move, which has no number — let the whole line collapse and packed the cancel X against the target sprite, breaking the row's own *"the X keeps its slot on every row so content stays aligned"* rule. **A slot that can be hidden must not be the slot holding the expand.** A dedicated `Slack` control owns it now.

## How to check this

Same flow as the PR body. What should be different: rows about a third shorter, clearly lighter than the section cards behind them, no scrollbar until the dock is genuinely full, the Wet/Chilled chip sitting under the **target** sprite, damage numbers in their own tinted pill, and an Execute button you cannot miss.

**Verification:** `ui squad presentation elemental dev` — **1677/1677 across 148 suites**, 453s. Three new cases, each falsified against a mutant: the consequence line's alignment (mutant put it at 1076 against the target's 1112 — exactly the 36px indent), the X's right edge on a numberless row (1146 against 1183), and no section owning a scrollbar while the queue fits.

**Still yours to judge, and one I'd flag:** I applied *"all the icons → 16×16"* to the **unit sprites** as well, because leaving those at 32 would have kept rows 32px tall and not solved the height problem at all. But a 32px map sprite rendered into a 16px slot is a fairly small character. If they read as mush, say so — putting the two unit slots back to 24 is one number, at the cost of ~8px of row height each.

Phaazoid:
🤖 Claude says:

**Round 3 — both corrections from the second play-check.** One of them reverses a rule round 2 had just written down, which is worth saying out loud rather than quietly re-editing.

## "All the icons" meant the UPSCALED ones

Round 2 read *"all the icons should be shrunk down to the 16x16 size"* as covering the unit sprites, on the reasoning that leaving them at 32 would keep rows 32px tall and not solve the height problem. That was the wrong half to give up, and you caught it: *"I like that the big ones shrunk, but the small ones, the unit sprites, shrunk too, and now they're not very readable."*

The action art is **authored at 16** and was being drawn at 32 — shrinking it costs nothing, which is the whole point. A map sprite is **authored at 32** and shrinking it destroys the one thing a row is read by. The rule the two share, now in canon: **shrink what was being upscaled, never what was authored at the size it draws.**

Actor and target are back to 32; the action icon stays 16.

## The chips came up into the line

You spotted slack round 2 was ignoring. Measured, it is **~40px**: the line is 181 wide and its fixed contents (32 + 16 + 32 sprites, a 35px readout card, a 16px X, five 2px gaps) take 141.

`Consequence` is now an `HFlowContainer` sitting **in the line** between the target sprite and the readout, and it does three jobs at once:

- **it is the slack** — it holds the line's horizontal `EXPAND` whether or not it has chips, which is what keeps the cancel X on the right edge of a numberless row. That is the thing round 2 needed a dedicated `Slack` spacer for; the spacer is gone.
- **it wraps rather than clips** — a crowded hit costs the row a second line and never a word. Measured: two chips take the row 24 → 40px, with its minimum at 209 inside a 215px section.
- **it keeps the receiver rule** — a chip *after* the target sprite is still not a chip beside the attacker, which was the part that was never about layout.

## "Shock" is its own authored field, not a renamed popup

Shortening the authored `popup` would have done it in zero code and is what you proposed — but `popup` is read by the glossary's composed interaction line and by the bug report, where there is room for the dramatic word and it is the better one.

`ElementalReaction.short_name` is therefore a **declared** second representation (Law #4): two genuinely different questions — what a reaction *shouts* vs what fits a 40px badge — one accessor (`badge_name()`, falling back to `popup`), and blank on reactions whose word already fits. Authored: **Shock**, **Temp Shock**, **Dried**, **Deep Chill**. `Chilled!` and `Wet` were already short enough to leave alone.

## A claim I made this round that was wrong

The first version of the chip comment said text-only was *forced* by the measurement. It is not — a mutant put the 16px icon back and nothing clipped. **Text-only is a choice about height** (an icon costs ~19px, which is what decides whether a second chip wraps), and the icon keeps its place in the tooltip. The comment says that now instead. If you'd rather have the icons back for a faster read, the layout tolerates it — say the word.

## How to check this

Same flow as before. What should be different: unit sprites legible again at their old size, action icons still small, and a status showing as a coloured pill **beside** the target sprite rather than on a line of its own — so an elemental hit no longer costs a row of height. The Shock combo should read `Shock`, not `Electrocuted!`, with the full wording on hover.

**Verification:** `ui squad presentation elemental dev` — see the run below. Two new cases, each falsified: the chip's placement between target and readout (mutant put it back under the line: chip at 1076 against a target ending at 1160), and a crowded row wrapping inside its section rather than overflowing.

**Embedded-content sweep:** nothing embeds `ElementalReaction` — the six reactions are standalone `.tres` under `Resources/Reactions/` read by `ReactionCatalog`, and no scenario snapshot or test fixture carries a copy. Grep-verified both ways (`ext_resource` referrers and `sub_resource` blocks), so `short_name`'s default reaches every reaction that has one.

---

## Round 4 (same push) — the world-event pills were grey on grey

*"The text for the void and falling is grey against grey, which is not readable."* Correct, and it was a **reuse** bug rather than a colour choice: those pills — `Fell 2!`, `Drowning!`, `Into the void!`, `Insulated!` — were tinted `ElementPalette.NEUTRAL`, which is the **rail's off state**, a structural grey deliberately chosen to *disappear*. Promoting it to text inverted its whole purpose. Measured at the fix: **luma 0.29 against a row at 0.23**, a delta of 0.06.

`QueueStyle.EVENT_TINT` is its own value now, deliberately **off the element wheel** — every element colour is saturated, so a cool near-white cannot be read as one, and the obvious warm amber would have collided with Earth's ochre.

**It takes a knob**, which is the one exception to this panel's *no knobs on the chrome* line, and worth stating rather than quietly doing: the chrome mirrors the inspect panel, which has none either, while this is a colour the queue **invents** and that has to read against the element chips beside it. Leaving it un-tunable next to seven knobs you are already dragging is the inert-slider problem in reverse. Find it on **Elemental → Action queue → World event text**.

The case drives a real insulated hit rather than asserting the constant, and asserts the **defect** — not the neutral, and real contrast against the row — so tuning the colour cannot red the suite unless you tune it back into grey. Falsified: both assertions red under the revert, the second printing the two lumas.

The rule went into canon in the general form: **a colour picked to recede cannot be promoted to a colour that must be read** — check what a value was chosen *for* before borrowing it.

**Verification for both rounds:** `ui squad presentation elemental dev` — **1680/1680 across 148 suites**, 510s.

Phaazoid:
🤖 Claude says:

**Round 5 — the parchment palette, as a player setting.** *"I really do like the parchment option too. Can you give me a player setting toggle for it?"*

It is a `PlayerSettings` choice row and no UI work at all: **Settings → Action queue colours → Slate / Parchment**. The page is a projection of `DEFS`, so a new row is one enum member and one table entry, which is the whole reason that table was paid for.

## The fork you ruled, and what it bought

The seven element colours are tuned to *glow* against a dark ground, and unchanged on cream they are pastel mush. So parchment either authored a second seven or adapted the one set — and you said adapt.

That turned out to be the better half on the merits too, in a way worth stating: **an element's hue is semantic** (fire is orange on either ground) so only its ink *weight* moves, and because the ink is derived from the same seven statics your sliders write, **dragging Fire moves both palettes**. The aim palette re-authors its colours, which is why a colour knob goes inert there whenever a non-default aim palette is picked. This one strands nothing.

The **chrome** is the opposite case: near-white header text on slate has to become near-black on cream, which is a role *inversion* with nothing to derive from. So `QueueStyle.Role` names the 22 colours this panel answers for, and `PALETTES` holds parchment's 22 — eleven from the mockup, eleven authored the way slate's were. Two of those reasonings invert rather than transpose: a **hover goes darker** on cream (a row already at 0.98 has nowhere up to go), and the **header strip darkens** its section where slate's lightens it. Two roles deliberately do *not* invert — the title sits on the dark frame and the Execute label on crimson, in both palettes.

## Two things had to grow, and both were invisible with one palette

- **the stylebox cache keys on the palette.** Every box is built out of the accessor, so a name-only key hands the parchment panel the slate box it built first — the chips repaint and the frame, the sections and the rows do not.
- **`restyle()` re-applies the chrome**, which was set once in `_ready`. Without it a swap re-renders every row and leaves the frame, the title and Execute wearing whatever the game booted in.

`SettingsScreen`'s exit act now fires both surfaces. The general form, and it is the one to remember: **a surface painted on an EDGE rather than polled needs the settings page to fire that edge.**

## Knobs

**Elemental → Action queue** gains **Parchment ink depth** and **Parchment ink saturation** — how heavily an element inks onto cream, and how far it is pushed toward pure hue first. They are **inert while Slate is live**, same as your aim-colour rows are under a non-default aim palette; tune them with Parchment picked, where the dock repaints under the slider.

## How to check this

1. Launch, start a mission, queue a few orders — a move, an attack, and an elemental hit if you can arrange one.
2. Pause → Settings → **Action queue colours** → **Parchment**, then Close.
3. Queue another order, hover a row, and refuse one (queue an attack then move the target out of reach).
4. Dev tools → **Game → Elemental → Action queue**, drag the two ink sliders with Parchment live.

**What you should see:** the whole dock changes on Close — dark brown frame, cream section cards, near-white rows, dark ink — without needing another order queued first. Element chips read as dark saturated ink rather than pastel. A refused row still goes red-bordered. The rail on a non-elemental row should be almost invisible against the row, exactly as it is on slate.

**What a failure looks like:** rows repainting while the frame stays slate (the chrome re-apply), or nothing changing at all until you queue an order (the exit act), or chips that read as washed-out pastel (the ink depth wants lowering).

## What the tests cover, and what they cannot

Nine new cases in `tests/ui/test_queue_palette.gd`, plus the settings-page wire. **No case asserts what a colour is** — the seven elements, the event tint and the two ink values are all knobs. What they pin instead: that DEFAULT is a fall-through and never a table row (behaviourally, by moving the knob and watching slate follow); that every role has a parchment value; that hue survives the swap for every element; that a colour knob still reaches parchment; that the cache does not serve the other palette's box, in both directions; and that the rail's off state stays quiet in *both* palettes.

The contrast law is asserted on **parchment only**, and that is a deliberate narrowing of what I planned. Parchment's ink is derived, so readability is a property of the adaptation that no colour knob can trip — an HSV colour at value *v* has luma ≤ *v*, so the floor is met by construction. Putting the same floor over the authored slate seven would quietly make seven of your knobs un-tunable past a threshold, which this ticket never agreed to.

**What the suite structurally cannot see, and is yours to play:** whether the seven adapted colours are still seven *distinguishable* things on cream. The law proves each reads against the paper; it deliberately cannot prove Ice and Water are still told apart, because that is a look question and pinning it would pin a tuned value. If it reads muddy, the two ink sliders are the first thing to try, and the fallback is the fork you declined.

**Verification:** `ui squad presentation elemental dev` -- **1690/1690 across 149 suites** (152s + 440s). Six mutants, each red on exactly the case it targets: the fall-through reading a copy of EVENT_TINT (behavioural, not just the table check); the cache key dropping the palette; the adaptation no-oped (Ice at luma 0.78 and Air at 0.71 against a 0.95 row -- the two the arithmetic predicted, and the other five clear the floor unadapted); the ink rebuilt from a fixed hue; parchment reading an authored COPY of Fire rather than the static, which is the second-authored-set design and is what proves the knob really reaches; and `restyle` re-rendering rows while leaving the chrome alone, which reds the assertion whose message says exactly that.

**Embedded-content sweep:** nothing to sweep — no content type gained a field and no `.tres` was touched. The one authored artefact in the diff is a `PlayerSettings.DEFS` row, whose persistence is keyed by the enum member's name and which no resource embeds.

**No `.tscn` touched**, so there is nothing for an editor save to re-dirty.

— Claude (Opus 5) · 2026-09-04

Phaazoid:
🤖 Claude says:

**Correction to the verification line above — CI found one failure my local run could not, and it is a miss in how I picked areas rather than a bug in the feature.**

`tests/core/test_player_settings.gd` holds the declared-duplicate law for choice rows (a row's enum values ARE the indices into its options list) and it **refuses any choice row that has not pinned its enum**, which is exactly what it exists for. `QUEUE_PALETTE` had not. One line, in `5369a7e8`; all five CI shards green.

**Why my areas missed it:** I picked `ui squad presentation elemental dev` from the surfaces the feature *reaches*, and on that axis the list was complete. But `Classes/core/PlayerSettings.gd` gained a line, and **a law suite lives beside the file it governs, not beside the feature that trips it**. The rule I should have used, and now have written down: union *where the changed file lives* with *what the change reaches* — the first list is mechanical and is the one that gets forgotten, precisely because reasoning about blast radius feels like the more thorough exercise.

So the honest verification line for this round is: **1690/1690 locally across the five areas I ran, one failure in CI's `rest` shard, fixed, and the full tree green afterwards.** The six mutants described above all stand.

— Claude (Opus 5) · 2026-09-04

Issue #739, opened 2026-09-04
Pre-mission: placement and squad-building on the board


Phaazoid:
🤖 Claude says:

The board half of #731 — the surface the dev asked for in Fire Emblem's shape:

> "you're in the menu, with a button press that can preview the full board at any point, and you can swap back and forth between the board preview and menu at any point during this phase. Moving unit spawns is done on the board preview, most everything else is done in the menus."

## Placement

- Click a cell inside the `DEPLOYMENT` zone with a unit held → place it.
- Click a placed unit → pick it up, or undeploy it.
- Click an empty legal cell → a dropdown of every roster unit not yet deployed, per the dev's description: *"click a blank spot and click add, and get a dropdown of all deployable units to bring one in."*
- The zone draws as an overlay so legal cells are obvious before the first click.

Deploying also works from the unit card (#731's menu half). **One state underneath, two surfaces editing it** — so "brought but unplaced" never exists as a state at all.

## Squads

The dev's ruling, 2026-09-04:

> "the player should be given tools to configure squads before starting the mission, but the mission does not impose rules about that on the player... Player agency is king."

And squads are built here rather than in a second interface, because this is where the units already are.

**Reuse the existing flow.** `game.create_squad` → `enter_target_pick_mode` → `SquadManager.join_squad` is immediate, with no action queue involved — so it works during a phase that has no turns. `joinable_squads` / `join_squad_mode` is the other half.

**`_formation_basics_ok` works unchanged**, and this is why the umbrella does not split it: deployed units have real cells, so the cohesion half is answerable; `has_acted` and `has_any_actions` are false on a board where nothing has moved. Building squads through the same door the game gates is also what makes over-capacity unreachable rather than enforced — dev: *"we just never run into this problem, it's not an endpoint the player can reach."*

## The one thing that must not happen

A click on a deployed unit must **not** open the radial action menu. The phase's `GameState` is what gates that (#731's phase ticket wires it into `menu_is_up()` / `_board_locked_for_player()`); this ticket is where the gap would show up if it were missed.

## Toggle

A key returns to the menu and back, from anywhere in the phase, without losing what is placed.

## Verify

[...]

Issue #740, opened 2026-09-04
Pre-mission: the menu shell and the unit card grid


Phaazoid:
🤖 Claude says:

**This is a LOOK ticket. It owes a rendered-mockup grill before any plan** — real sprites, 2x, the candidates side by side. Prose cannot settle a layout, and the build cannot be checked by whoever writes it.

## The shape, from the dev's sketch (2026-09-04)

Four regions. This ticket builds the shell and the top-left region; the stash is #S7, the bottom strip is #S9.

**Top left, scrollable — a grid of unit cards**, one per roster unit. More than six units, so it scrolls. Each card is self-contained; there is no drill-down.

**The card**, from his zoomed sketch:

- **Top left** — the unit sprite (portraits later, when they exist), name beneath it, stats beneath that in two columns.
- **Beside the sprite** — limb condition, and the job title. The job title is the job picker's home (#S8).
- **Top right** — the inventory slots. Interactable: equip, remove, trade. Remove sends the item to the stash. `Unit.MAX_INVENTORY_SIZE` is 6.
- **Bottom right** — abilities, as inspectable 16x16 badges. These are the unit's *live* abilities from every source — job, innate, worn gear — which is one `Unit.get_live_abilities()` call, the same answer the inspect panel gives. Not a list of everything that exists.
- **A deploy toggle**, which drops the unit onto the first free deployment cell.

## What to reuse, and what not to

The board preview's enemy inspect is the **existing inspect panel, unchanged** — that is what FE's map inspect is, and the dev agreed explicitly. This card is a second *view*, which is fine. What must not be duplicated is the **derivation**: effective stats, the computed weapon view with mods folded in, live abilities. Those have single answers already (`get_effective_stat`, `get_live_abilities`, `inventory_panel`'s computed view) and this card reads them.

`info_panel.gd` and `inventory_panel.gd` are the closest existing tenants and worth reading before drawing anything.

## Watch

- **Density.** Sprite, name, eight stats, limbs, job, six inventory slots and an ability row is a lot for one card at two columns. The mockup exists to find out whether it fits before the build does — laying candidates out to scale is what caught the row-width problem in #685.
- **UI scale** — #659 made the HUD scale with window size and #686 is the player-facing multiplier. A new full-screen surface should ride the same seam, not fix its own pixel sizes.
- Ability badges need 16x16 art per ability. `Item.icon` exists; `AbilityData` has no icon field yet.

Parent: #731

— Claude (Opus 5) · 2026-09-04

PR #775, opened 2026-09-05
#774: a board-side affordance for the pre-mission phase

### LINKED ISSUE #774: Pre-mission: the phase has no on-screen presence while the board is up
Reported by the dev after playing #740: *"while inspecting the board, there's no way back to the loadout menu."*

## What is actually missing

The swap-back mechanism exists. `Tab` is bound (`project.godot`, physical keycode 4194306), registered in `Controls.gd` as a BOARD-context key, and dispatched from `game._input` beside `commit_deployment`. The "Board" button's own tooltip names it.

What does not exist is any **on-screen sign that the phase is still running**. #739 hides exactly two HUD surfaces during the phase -- `SquadActionQueueControl` and `EndTurnButton` -- so once the screen swaps away the player is looking at a board with a lit deployment zone, no banner, no button, and nothing that says they are mid-anything. The way back is a key nobody told them about, on a screen that just vanished.

The same is true of `Enter`: the phase's one exit has no on-screen affordance either while the board is up.

## Also uncovered

No test drives either key. `tests/ui/test_pre_mission_screen.gd` calls `toggle_deployment_menu()` directly, so the function is pinned and the wire from the keypress is not -- the blind spot the "test the wire" law names. Whether Tab was working-but-invisible or silently dead was not answerable from the code.

## The slot this drops into

`EndTurnButton` reserves a 145x36 slot at bottom-right with an 8px margin, and `MissionStatusPanel` holds `BUTTON_CLEARANCE := 44` above the corner whether or not that button is visible. During the phase that slot is empty and still reserved -- it is the "advance the phase" corner, vacated by the ticket that created this gap. Anything the same height dropped there reflows nothing.

## Scope

A `PreMissionBar` in `game.ui_layer`, built in `_open_deployment` and freed in `_close_deployment_menu` so all three phase exits already take it. Two buttons:

- **Loadout** -> `toggle_deployment_menu()`, tooltip naming Tab
- **Begin Mission** -> `commit_deployment()`, tooltip naming Enter

Both stay live with nobody placed rather than greying: #739 already ruled that refusal has a voice, and the screen's own Begin carries the same rule.

Part of #731.

Phaazoid:
Closes #774.

The pre-mission screen swaps away for the board preview and nothing was left on screen saying the phase was still running. #739 takes the queue dock and End Turn down for the phase, so the board came up with a lit deployment zone, no chrome, and two keys nobody had been told about.

**The mechanism was not missing -- the affordance was.** Tab is bound, registered in `Controls.gd`, and named in the tooltip of the very button that swaps you to the board. It works; `test_the_tab_key_reaches_the_swap` passes on unmodified code. It was just invisible.

## What this adds

`PreMissionBar` fills **End Turn's slot**, which is empty and still reserved during the phase -- `MissionStatusPanel` lifts itself `BUTTON_CLEARANCE` clear of the corner whether or not that button is visible, so this reflows nothing. The height is read off that clearance (`BUTTON_CLEARANCE - CORNER_MARGIN`) rather than measured a third time: the slot stays one fact with one owner, which is also why there is no test pinning the geometry.

Two buttons, **neither greying** -- which is the opposite of what the screen does with the same act, on one shared reason. `commit_deployment`'s refusal speaks through `TurnBanner`, a plain child of Game: under the screen's `CanvasLayer` it cannot be seen, so the screen greys its Begin and puts the reason in a tooltip. Out on the board the banner is in plain sight, so the button stays live and lets the refusal speak.

**Begin asks first** (dev tweak, 2026-09-05). `MissionController.confirm_and_commit` owns the card rather than either button, because the commit has three doors -- the bar, the screen, and Enter -- and a confirm on one is one the other two walk around. The empty-force case is deliberately *not* asked about: a card in front of an act already destined to be refused is two dead ends where one would do.

Lifecycle joins the screen's in `_close_deployment_menu`, the one place documented as *"every path that ends the phase has to take it"*, so reset, commit and abandon take the bar for free rather than growing a second teardown. `toggle_deployment_menu` moves both halves as one act.

`FOCUS_NONE` on both buttons is load-bearing rather than tidiness: Tab is also Godot's built-in `ui_focus_next`, so a focusable button here would collect a focus ring off the very key that swaps the screen back.

## The blind spot underneath

Every existing case drove `toggle_deployment_menu()` directly, so **a bound-and-dead key would have passed a green suite**. The new key case builds its event on `physical_keycode`, the way `project.godot` binds it, so a re-bind that misses this action fails here instead of in someone's hands.

## Tests

Three cases in `tests/ui/test_pre_mission_bar.gd` -- the swap and the Loadout button in one real sequence, Begin's wire, and the key. The three phase exits are not re-tested: they route through `_close_deployment_menu`, which `test_pre_mission_screen.gd` already has a case each for, and its Abandon case carries the one added assertion that the bar goes with them.

Four mutants, all killed by the case that names them:

| mutant | killed by |
| --- | --- |
| the bar's half of the swap removed | the swap case |
| the bar not freed with the screen | Begin's case + the Abandon assertion |
| `pressed.connect` removed | the swap case (Loadout) |
| the Tab dispatch removed from `game._input` | the key case |

Scoped run: `ui` 358/358 green, `law` + the controls pane 214/214 after the `Controls.gd` text change. CI owns the full tree.

## Canon

`visual-clarity.md`'s *"End Turn is permanent, with two exceptions"* block, `missions.md`'s phase section and CLAUDE.md's pre-mission paragraph all carried #739's stand-down without the hole it left. The travelling rule: **a surface that stands down owes the frame whatever the player still needs from that corner.**

## How to check this

1. Start Level 1 and hit **Board** on the pre-mission screen.
2. Bottom-right, where End Turn normally sits: **Loadout** and **Begin Mission**. Objectives panel unmoved above them.
3. **Loadout** brings the menu back with its scroll position intact. So does Tab, from either side.
4. Place someone, swap to the board, hit **Begin Mission** there -- the battle starts and the bar goes with the screen.
5. Undeploy everyone, then press Begin on the bar: refused, with *"Deploy someone first"* on the turn banner -- which you can now actually see, because there is no screen over it.
6. Pause menu -> Abandon from the board preview: the bar does not survive onto Mission Select.


Phaazoid:
**Three tweaks pushed** (`3628cbcd`).

**1. Begin asks first.** The card lives on `MissionController.confirm_and_commit`, not on a button, because the commit has **three** doors -- the bar, the screen, and the Enter key -- and a confirm bolted to one of them is a confirm the other two walk around. Enter is the one most likely to be hit by accident, so it would have been the worst door to leave open.

The empty-force case is deliberately **not** asked about. A card in front of an act already destined to be refused is two dead ends where one would do, so it goes straight to `commit_deployment` and its voice.

No re-read of `_deploying` after the await, and that is deliberate rather than an oversight: the phase *can* end while the card is up (a dev key is exempt from the modal freeze), but `commit_deployment`'s own first line already refuses outside the phase. A guard here would be a second copy of that one, and an unfalsifiable one.

**2. The parchment fix, and the shape underneath it.** Loadout took its box from `row_box()` -- palette-aware -- and its font colour from the theme, which is authored for the dark shipped ground. Under Parchment that is white on cream.

`HEADER_TEXT` is the role to pair with a row, because it is one of the ones that **inverts** with its ground; `TITLE_TEXT` and `EXECUTE_TEXT` stay light in *both* palettes by design (they sit on the dark frame and on crimson), so neither can ever sit on paper. And all four font states need it: a Button falls back to the theme's `font_hover_color` independently, so a normal-state-only fix brings the bug straight back on mouse-over.

**3. Size.** 150x40 rather than shrink-wrapped. 40 rather than End Turn's 36 because the corner's whole height budget is `BUTTON_CLEARANCE` = 44, and spending 40 leaves the 4px that keeps the bar and the objectives panel reading as separate surfaces. Above 44 the row would run under the objectives.

## A correction to this PR's own description

I wrote that the bar's Begin stays live "the rule the screen's own Begin already carries". **The screen's Begin greys** -- `test_begin_is_refused_with_nobody_placed_and_carries_the_reason_itself` has pinned that since #740. The two surfaces genuinely differ, and the reason is better than the false symmetry I claimed: `commit_deployment`'s refusal speaks through `TurnBanner`, a plain child of Game, so under the screen's `CanvasLayer` it cannot be seen at all and the screen has to grey and explain in a tooltip. Out on the board the banner is visible, so the button can stay live and let it speak. Two surfaces, two answers, one question -- *can the player hear the no.* The body and the canon are corrected.

## Tests

Four cases now, up from three. The new one is the reported defect's own shape: both buttons carry their own ink in every font state. The confirm folded into the existing Begin case instead of taking one of its own -- it now walks the whole sequence: empty force is refused unasked, then the **Enter key** puts the card up and No changes nothing, then the **button** puts it up again and Yes starts the battle. That is all three doors and all three answers in one real sequence.

The pairing itself is a law rather than a fact about this button, so it went to `test_queue_palette.gd`: `HEADER_TEXT` reads against `ROW_BG` **and** `ROW_HOVER_BG` in both palettes. Both palettes is allowed here where the element cases are parchment-only -- those are `GameKnobs` rows the dev drags, these are authored consts.

Four more mutants, all killed:

| mutant | killed by |
| --- | --- |
| the confirm never appears | the Begin case |
| the refusal gets asked about first | the Begin case |
| only `font_color` overridden | the ink case |
| parchment's `HEADER_TEXT` goes light | the pairing law |

`ui` + `flow`: 675/675 green.

Issue #629, opened 2026-08-28
Frame animation: read a card-format spritesheet, play it on a Sprite3D


Phaazoid:
## What

The project has **no frame-animation system at all**. Grepped `AnimatedSprite2D|AnimatedSprite3D|SpriteFrames|AnimationPlayer|AnimationTree` across `Classes/ Scenes/ Resources/ tests/ game.gd`: zero hits outside `addons/`. A unit's visual is one still `Texture2D` swapped between three authored stills — map, move, downed.

This builds the missing layer, end to end, against one sheet: **read a card-format spritesheet into a `SpriteFrames`, and play it on a `Sprite3D`.** It stops deliberately short of the beat sheet.

## Why it is its own ticket

Three tickets name this infrastructure and none owns it:

- **#603** (the sprite half of the battle zoom) is *which gesture plays when*, and has five forks unpicked. It cannot start without something that can play a gesture.
- **#358** wants one effect stack for status reads on the sprite.
- **#414** shipped its block moment as slice one with a declared deferral in shipped code — `Classes/actions/AttackAction.gd:89-92`: the bodyguard plays *the same lunge an attack plays*, and "the loud jump-in-front the design wants waits on the battle-zoom / sprite-FX stack (#358) — no bespoke animation machinery ahead of it."

## What the sheets actually are (measured 2026-08-28)

The dev dropped four GBA Fire Emblem battle sheets into `Art/Units/ZoomAnimations/`. They are **not the same kind of artifact**, and the difference decides everything:

| Sheet | Format | Layout | Timing | Registration |
|---|---|---|---|---|
| **Sage** | PNG | **162 regions, 159 exactly 66x42 on a 70px pitch** | **printed per frame** | **preserved** |
| Soldier | GIF | merges — attack row is one **257px** blob, crit row **377 + 285** | none | tight bbox only |
| Brigand | GIF | mostly separates, some merges | none | tight bbox only |
| Sniper | GIF | mostly separates | none | tight bbox only |

The Sage is a **card-format** sheet: every frame sits in a fixed 66x42 card, with its **duration printed above it in GBA frames** (60, 6, 15, 11, 18, 8...), the animation's name beside the row, and a return pointer ("go to Attack frame 1", "go to Attack frame **3**"). Three palette blocks of ~53 frames each.

The other three are older rips carrying none of it. Two things no extractor recovers from them:

1. **Timing.** Not on the sheet.
[...]

Issue #603, opened 2026-08-28
Battle zoom: the sprites carry the blow - beat-driven attack, recoil and impact animation


Phaazoid:
## What

The battle zoom so far is entirely **camera and staging**: the shot list (#524), the timing profiles (#519), the director (#520), the tear-out (#521). The units themselves still play the same small placeholder gestures they played on the plain board. This is the other half of a blow — what the *sprites* do once the camera is finally close enough to see them.

## Why now

The dev is mocking the target directly (2026-08-27): the map sprites currently in use come from a GBA-era Fire Emblem, and that game's **zoomed combat sprites** are going to be used to mock how the battle zoom animations should look, as part of this arc.

**One structural note, so the mock is read correctly.** FE's combat zoom is a *separate side-view scene*: the camera holds still, a flat backdrop replaces the map, and the sprites carry the entire punch — wind-up, strike, recoil, crit flash, screen shake. This arc does the opposite: #521 keeps the real board and lifts it, #520 moves the real camera across it. So the mock is a **feel target — what the moment should read like — and not a construction blueprint.** Any proposal that assumes a fixed camera or a bespoke combat scene is answering a question this arc has already answered the other way.

## What exists

`UnitVisuals` already owns per-unit sprite effects — `play_attack_lunge`, the flash, the invalid-order shake — and `UnitMirror` mirrors that offset into 3D (#321, `Classes/presentation/UnitMirror.gd:415`). So there is a working seam and a working mirror.

What there is **not** is any relationship between those gestures and the beat sheet. A lunge is the same lunge whether the blow is a scratch, a killing hit, or a cap that should have killed and did not — and nothing recoils, flinches, or reacts. Every distinction #524 computes and #519 times is currently expressed only by how long the camera waits.

## Two customers are already waiting on this by name

This is not a third seam beside them — it is the thing they were deferred *to*:

1. **#414's block moment.** It shipped as slice one and the code says so outright, at `Classes/actions/AttackAction.gd:89-92`: the bodyguard plays *the same lunge an attack plays*, and "the loud jump-in-front the design wants waits on the battle-zoom / sprite-FX stack (#358) — no bespoke animation machinery ahead of it." That is a declared deferral sitting in shipped code.
2. **#358**, the one effect stack for status reads on the sprite (wet drip, frost sheen, Crisis).

[...]

PR #782, opened 2026-09-05
#603: the Brigand's axe swing plays in the battle zoom

Phaazoid:
Imports the Brigand's axe attack, builds an animation from it, and plays it when a brigand attacks. First slice of #603 and the Brigand half of #635.

Until now a blow was the same 0.18s placeholder lunge on the board and in the zoom. #629 built frame playback and stopped before beats; this is the first gesture the game plays on its own.

## What it looks like

A contact sheet of all twelve re-carded frames, with the ground line and the map sprite it replaces for scale, is in the chat alongside this PR. 12 frames, 74 GBA frames, ~1.23s. The axe lands on frame 8.

## Three parts

**`LooseSheet` — a sheet with no cards is RECONSTRUCTED, not read.** It is `CardSheet`'s sibling rather than a mode of it, because the two make different promises. A card is a fixed viewport and therefore carries the animation's footwork; this sheet was ripped to tight bounding boxes and laid out by hand, so the only thing left to align by is the ink. Frames are 8-connected regions of not-page ink — 8-connected because a blade meeting a hand diagonally is one frame, where `CardSheet` needs 4-connected for its overhang finding — then **re-carded**: ink bottom flush with the card's ground line, ink centred across it.

**The generator learns a second kind of entry.** Durations are authored in the manifest because the sheet prints none. From `_write_frames` down the two kinds are one path; the origin became a parameter, and nothing on the Sage path changed (regeneration verified byte-identical).

**The wire is a version-gated poll, not a signal.** `UnitVisuals` counts its lunges, `UnitMirror` compares that count to its own last-seen value (#308's idiom). A poll of the offset cannot answer this — the invalid-order shake moves the same channel. The mirror keeps death as the one signal it listens to. The ledger is **seeded from the unit's own count when a sprite is built**, so a unit mirrored after its first blow (mid-pass spawn, #739 redeploy) does not replay a swing already struck. "Which set belongs to this art" now has one home, `UnitSprite3D.zoom_set_for`, caching misses because most families have none and it is asked on every blow.

## What it gets right, and what it cannot

**Vertical is right by construction** — the lowest ink *is* the ground, so the leap keeps its height off its own drawn shadow with nothing told about jumping.

**Horizontal drifts, and this is the honest part.** An axe swung out to one side widens the bounding box that way and walks the body the other. Measured on the shipped sheet: the worst frame-to-frame body shift is **about 10 texels, a third of a cell**, at the two frames where the swing arc appears. I measured two alternative anchors before accepting it — the ink centre of the bottom rows, and of the lower half — and **both were worse**, with 6–9 texel steps of their own. There is no rule that recovers registration a rip discarded. Hand-authored per-frame offsets are the only real cure and they are #635's fork 2, which is yours to pick.

## Declared, not discovered later

- **The beat still fires damage before the axe lands.** `AttackAction` awaits the 0.18s tween and then applies damage, so the jolt and the cubes land ~0.45s early. Making a beat wait for a gesture is #603's forks 1 and 2 — the next slice.
- One consequence **repeals a declared cut**: #520 2c recorded that "freeze mid-swing" could not exist because the sprite was back at rest when the death signal fired. With a 1.23s gesture running past an 0.18s tween, there is now a swing to freeze mid-. Not wired; just possible.
- **The Sage swings too.** Celest is drawn with the Sage art and its set carries `attack`, so she plays her 5s cast on a blow. The plan recommended letting it, as the clearest evidence for the timing question — say the word and it becomes a per-set opt-in.
- The swing outlives its beat, and hands back from a 42×74 card to the 32px still with a pop. Both inherent to mixing placeholder sources.

## Tests

`test_loose_sheet` (7 cases, synthetic sheets — the shipped art is content and gets property assertions only), plus wire cases on `test_unit_mirror` and lookup cases on `test_unit_sprite`. Run locally: the `law` and `dev` areas in full (644 cases) plus the three touched `presentation` suites (25 cases), 0 failures and 0 orphans throughout. The rest of `presentation` is CI's, per the standing rule.

**Five mutants, and the fifth is the finding: it SURVIVED.** "Any vertical overlap is the same row" left the ordering case green, because my fixture put row 2 five pixels *below* the tall frame so the two never overlapped — the case was named for a rule it could not see. Fixed in its own commit (`2242a06c`) against the Brigand's real geometry, and the mutant then redded. The other four killed precisely: silence the lunge counter → the wire case; anchor at the bbox centre → the ground case; drop the size filter → the label case; seed the ledger from zero → the arrival case.

Verified end to end outside the suite: on three real boards a live brigand swings the real atlas, the pivot moves from (0,16) to (0,37), and 9 distinct frames play.

Generator verified idempotent over a clean tree.

Closes nothing — #603 keeps four unpicked forks and #635 keeps the other two sheets.


Phaazoid:
🤖 Claude says:

Closing unmerged, at the dev's call — and recording why, because the branch is being deleted and the reason is not "this was bad work".

**His words, 2026-09-14:** *"I never merged it because I was unhappy with the timing, and wanted to pause exploring that direction until I made a dev tool to help me create timings in animations."*

So the pause is about **authoring timings**, not about this PR's code. That question is now [#635](https://github.com/Phaazoid/Godoiosis/issues/635), re-scoped from *"the other three zoom sheets"* into the timing/registration authoring tool it was always really about — its forks 2 and 3 already were that tool.

## What goes with it

This branch's art depends on `Art/Units/ZoomAnimations/`, which is being **deleted** as [#139](https://github.com/Phaazoid/Godoiosis/issues/139)'s last Fire Emblem row — [#937](https://github.com/Phaazoid/Godoiosis/issues/937) took the map sprites, this takes the battle-scene sheets, and the demo zip stops carrying art that cannot be licensed at any price. `CardSheet` and `tools/zoomanim/` go with them: they could only ever read the one sheet that was card-format, and a duration OCR'd off printed digits is not a rule that survives its sheet.

**`LooseSheet.gd` and `tests/presentation/test_loose_sheet.gd` are the part worth naming** — 191 lines plus a suite, the second reader built precisely because the Brigand sheet is *not* card-format. That is real work and it is not being thrown away: **GitHub keeps a closed PR's head ref**, so it is recoverable from this page even after the branch is deleted. If #635's tool wants a starting point for reading a sheet with no cards, it is here.

## What survives on `main`

`SpriteAnimator` and `UnitSprite3D.play_animation` stay. They drive any `SpriteFrames`, so they are **unfed rather than dead** — the seam [#603](https://github.com/Phaazoid/Godoiosis/issues/603) comes back to once there is a way to author timings. `DevController`'s K hotkey stays too and now warns for every unit, which is the truth.

— Claude (Opus 5) · 2026-09-14

Issue #805, opened 2026-09-06
Attack patterns PR 3: spread origin points (a burst that spreads from where it lands)


Phaazoid:
🤖 Claude says:

PR 3 of the attack-pattern revamp (parent issue holds the rulings). Blocked by the data model PR. Needs its own design pass before building; this issue holds the ask so it does not evaporate.

## The ask (dev, 2026-09-06)

> *"one could lob a bomb type transmutation, it lands, and then an explosion spreads from that landing point of origin, but that second half of the attack could be blocked by walls, etc. This would also help if say, a melee unit smashes the ground in front of them, creating sideways shockwaves. Those shockwaves would originate from the point struck, not the unit."*

## Where the rules stand today

- The lob / flat settings (`arc_clearance`, the tolerances, `vertical_rule`) gate the shot's path from the ATTACKER to the AIMED cell (`Reach.vertical_aim_ok`, #258).
- A self-anchored spread is truncated lane by lane from the shooter (#756, dev: "truncate, and all 8").
- A footprint anchored at range is BOARD-BLIND: nothing traces the burst outward from the aimed cell, for any pattern. #218's session deferred the 3D blast-extent question and the revamp keeps that (ruling 8).

## What has to be decided

- The spread's origin is the aimed cell: its elevation is the origin height, and the shooter's own cell is not in the picture once the shot has landed.
- Is the burst a flat trace from the origin (a ground burst), or does it carry its own clearance? The lob was spent getting there.
- Whose tolerances judge a burst cell: the attack's up/down tolerances read from the origin cell, or a separate pair?
- The truncation's lane rule assumes ONE facing. A cross at range spreads four ways, so the predecessor rule has to be re-stated for a stamp that spreads from its centre (a flood from the origin through stamp cells is the obvious shape, but it must not break #756's rule that a side lane never rides the centre lane for self-anchored spreads).
- Whether a self-anchored stamp may ALSO name a spread origin other than the shooter (the ground-smash case: aim a facing, spread from the struck cell).

## Related

- #756 (spread truncation), #258 (vertical aim), #218 (the verticality session that deferred blast extent).

Phaazoid:
also, right now, there is not a way to author a ranged attack that gets stopped by the first unit in its path, but is not aoe.  For example, a gun you cannot shoot past allies, etc

Issue #803, opened 2026-09-06
Attack patterns PR 1: one AttackPattern class (range + stamp) and the content/test sweep


Phaazoid:
🤖 Claude says:

PR 1 of the attack-pattern revamp (parent issue holds the rulings).

## Scope

- `AttackPattern` becomes the one concrete pattern: `min_range` / `max_range` / `max_and_a_half` plus a `stamp` of `Vector2i` offsets relative to the anchor, authored by coordinate (an `x,y x,y` text row in the Attack Editor, or Godot's inspector) until the grid widget lands (sub-issue 2).
- `ManhattanRangePattern`, `ForwardLinePattern`, `ForwardWidePattern` deleted, with their uid sidecars.
- The anchor rule (max range 0 = self-anchored, aim a facing), the rotation rule (grid up = forward), the near-to-far emission order, and the centre offset, each pinned by a law case that builds its own pattern.
- **Amended at build:** min range 0 under a larger max is NOT a lint refusal. It is authored content already (`Heal2.tres`, `heal2.tres`, mission 2: a self-targetable heal), so an aim at the attacker's own cell is a rule instead: no cardinal, the stamp is placed unturned, grid-up forward. Pinned by its own case.
- `AttackLint` gains: an empty stamp under a live ring BLOCKS (the attack can be aimed and lands on nothing). Its "reaches nothing" hint and `board_view`'s pattern string move onto the one class; the tooltip table with them.
- The sweep: every authored `.tres` rewritten as range plus stamp, both ways in one diff, uids preserved (a runtime save drops them). Before the old classes went, every authored attack's geometry was snapshotted headless in a worktree at `origin/main` and diffed against the rewritten files: identical. That equivalence check is a session run reported in the PR, not a committed test.
- The 22 test files that construct pattern classes by name go through one fixture helper.
- Every Attack Editor function today keeps working: the three modes, load / new / Save As / Update / Delete, the refuse-unfireable lint at save, live sync through `DevWidgets._adopt`. The `attack_pattern` class picker now offers `AttackPattern` itself, since there is one class and a new attack still needs a way to be given one.

## Not in this PR

- The clickable grid (sub-issue 2).
- Truncating a footprint anchored at range (sub-issue 3). Anchored splash stays board-blind exactly as today.

Issue #802, opened 2026-09-06
Attack patterns: range plus an authored stamp replaces the three pattern classes


Phaazoid:
🤖 Claude says:

## The ask (dev, 2026-09-06)

Two attacks have been wanted for a while, a **cone** and a **cross** (placed at Manhattan range, then covering a plus shape). Rather than add two more pattern classes, rethink how an attack's geometry is authored:

> *"I want a system where I specify two things about an attack - its range (manhattan range does this quite well), and its pattern. Right now, cleave and line are just patterns that always originate at a manhattan distance of one. What if I wanted to fire a line off at a range? There's no way to specify that now, and it feels silly to create another attack type for it when the current line already exists."*

The range half stays as it is (min / max / the half-step bevel). The pattern half becomes a **stamp**: a set of cells the dev clicks on an odd-sized grid in the Attack Editor, centred on the cell the attack originates from. Cleave, Line and the single-cell Manhattan attacks become examples authored in that editor rather than classes.

The code already asked for this: `ForwardLinePattern.gd`'s header (#20, re-verified #104) says *"if they still haven't diverged when the next pattern lands, consolidate rather than add a third."* Cone and cross are that pattern. The Chemical Spitter's materia cone (#97) is the first content this unblocks.

## Why it is cheaper than it looks

Every runtime consumer (the click handler, hover, the AI, `SquadManager`, `PlanResolver`, overwatch, the plan validator, the play session) goes through `Reach` and never names a pattern class. The game side is one resource class plus one editor widget. The cost is the content and test sweep: 38 `.tres` files author one of the three classes (31 under `Resources/`, 7 embedded in scenarios and fixtures) and 22 test files construct them by name.

## Rulings (dev, 2026-09-06, from the design conversation)

1. **Anchor is derived from range, not a flag.** Max range 0 = the stamp sits on the attacker and the aim is a FACING (today's directional path). Max range 1 or more = the stamp sits on the aimed cell and the aim is a cell. *"I agree with 1."* This is what "a melee-only toggle" meant.
2. **Orientation: the top half of the grid is forward.** The stamp rotates to the facing (self-anchored) or to the attacker-to-target cardinal (anchored at range), using the existing diagonal tie-break. *"yes, and the top half of the grid intuitively represents forward."*
3. **Emission order is near to far along the facing**, so #756's truncation keeps its input contract and volley order stays deterministic. *"near to far makes sense here."*
[...]

Issue #741, opened 2026-09-04
Pre-mission: the stash, and every way gear moves


Phaazoid:
🤖 Claude says:

The top-right region of #731's layout, and every way gear moves.

## The stash

The roster's loose gear (#S1 puts it there), rendered as a scrollable list. Dev, 2026-09-04, on why it is finite and authored rather than the whole catalog:

> "B, and I'll author generous stashes when I want to test broadly."

The reasoning worth keeping: `progression.md`'s law is *"economy is a dev lever, not a grind — each mission grants exactly what the designer placed."* An infinite catalog is not a generous version of that law, it is the law switched off — every loadout collapses to "take the strictly best item", and the balance signal a playtester's choice carries disappears, because there was no choice. A generous stash is a dial; the catalog is no dial at all.

## Movement

- **Stash → unit** — select an item and add it to a unit, or click-drag onto a card.
- **Unit → stash** — the card's remove action.
- **Unit → unit, directly.** Not routed through the stash: *"Should be able to trade between units directly as well."*
- Equip and unequip within a unit.

Every one of these has to respect `Unit.MAX_INVENTORY_SIZE` (6) and say so when it refuses, rather than silently dropping the drag.

## Reuse

`inventory_panel.gd` already renders slot rows with the computed weapon view (elements including mods) and an equip/unequip/toss popup. It is the closest tenant, and its computed view is the thing not to reimplement.

`EquippableData.copy_equippable()` is how a unit receives its own copy — `WeaponInstance` overrides it to keep its shared template uncopied. Anything moving gear has to go through it rather than assigning references, or two units end up sharing one instance's battle state.

## Not here

Loose **mods** are not in the stash — that travels with #732, along with making `WeaponModData` an `Item` at all. Until then the stash is weapons, armor and runes.

## Verify

Move an item every direction the list above allows and assert the source loses exactly what the destination gained. Fill a unit to six and assert the seventh is refused with a reason. Trade an equipped weapon away and assert the unit's equipped index follows correctly rather than dangling — that index is a position in the inventory array, so anything that reorders the array is a place it can go stale.

Parent: #731 · Related: #732

— Claude (Opus 5) · 2026-09-04

PR #807, opened 2026-09-06
#804: the attack stamp is authored on a clickable grid

### LINKED ISSUE #804: Attack patterns PR 2: the clickable stamp grid in the Attack Editor
🤖 Claude says:

PR 2 of the attack-pattern revamp (parent issue holds the rulings). Blocked by the data model PR.

## The ask (dev, 2026-09-06)

> *"I would be given a grid to click tiles in, and the tiles I click light up, indicating the pattern I'm creating. I could choose how big the grid is, but it must always have a center, so odd numbered grid sizes only. The center tile would be where the attack originates, but wouldn't necessarily need to be filled in."*

## Scope

- A custom row in the reflective editor (`DevWidgets`) for the `stamp` field: a clickable odd-sized grid with a size stepper, the centre cell marked, the top half meaning forward. Size is the widget's view only (ruling 5); the data stays offsets relative to the centre.
- The row feeds the existing live-sync copy path so a click reaches the board the way a spinbox edit does.
- No shape presets (ruling 6): the rewritten Cleave / Line / point attacks are the examples, and copying one is the workflow.
- The coverage law in `tests/dev/test_property_tips.gd` must still see the field's tip on the built control.

Phaazoid:
Closes #804. PR 2 of the attack-pattern revamp (#802); the data model was #803, spread origin points is #805.

## What changed

The reflective editor draws an attack pattern's stamp as an **odd-sized grid of toggle cells** around a marked centre, with the top of the grid meaning forward. Clicking a cell toggles it. A stepper grows and shrinks the grid, and the coordinate row from #803 stays underneath.

- **The size derives from the stamp's own extent and is never stored** — the widget's view alone, per the ruling. It is floored so there is always a ring to click into and capped so no stamp can demand a grid wider than the panel. **Minus greys rather than dropping an authored cell**; a text edit grows the grid but never shrinks it mid-keystroke, so deleting a digit does not collapse and re-expand it.
- **Which fields get a grid is DECLARED by the resource**, through a static `grid_fields()` beside the export it describes, exactly as `property_tips()` works. It is not inferred from the type: `ScenarioUnitEntry.watch_cells` is also an `Array[Vector2i]` and holds absolute board cells, which a grid centred on 0,0 would lie about. Nothing draws that entry reflectively today, and this keeps it that way by construction rather than by nobody having tried.
- **The caption names the anchor and the pattern answers it** (`grid_caption`, reading `is_directional()`, so it cannot drift from the rule). It follows `max_range` live through a new `emit_changed` on the editor's write path. `DevWidgets._adopt` has emitted `changed` since #589 with the comment "the hook a panel can redraw off; nothing listens yet" — this is its first listener. A rebuild driven from the range spinbox's own signal was the alternative and is refused: it frees the emitting node, which is #741's trap.
- Cells are out of the focus chain, and nothing rebuilds the form: a click restyles one button, the stepper and the line rebuild only the cells, and neither is a cell.

## A bug this found, shipped in #803

`get_property_list()` reports a typed array's element as a **type id, not a class name** — an `Array[Vector2i]` comes back with `hint_string` `"6:"`, never `"Vector2i"`. #803's coordinate row compared against the class name, matched nothing, and **the stamp was unauthorable in the running Attack Editor for the whole of that ticket's life**, reachable only through Godot's own inspector. Nothing caught it because the tips law reads the property table rather than the built controls and the live-sync suite drives a scalar, so no case ever asked whether the row appeared. Fixed here, recorded as a sharp edge in CLAUDE.md, and pinned by a case asked of the real property dictionaries.

## How to check this

**What to do.** Launch, open the dev tools, Attack Editor. Load `Sweeper` (Cleave): a 5x5 grid with the three cells one row above the centre filled, `Max Range 0`, and the caption reading that the centre is the attacker. Click cells to add and remove them. Press the plus to grow the grid, then the minus to shrink it back. Now set Max Range to 2 and watch the caption change to say the centre is the aimed cell. Then author the cross: New, Max Range 2, click the centre and its four neighbours, save, put it on a family's extras and fire it at range two.

**What you should see.** The centre cell wears a brighter, thicker border whether it is filled or not. The grid never shows an even number of columns. The minus button is dead whenever shrinking would drop a cell you authored. The coordinate line under the grid tracks every click, and typing into it fills the grid.

**What a failure looks like.** A cell that lights up but does not survive a save, which would mean the click is not reaching the resource. A caption that still says "the attacker" after the range moves off zero. A grid that jumps size while you type in the coordinate line. Boxes instead of a caret in the forward marker, which would mean the font lacks the glyph.

**What the tests already cover.** 19 cases in `tests/dev/test_stamp_grid.gd`: the size derivation and its floor, the odd-span invariant, the stepper's walk and cap, the shrink refusal, click-through to the resource, the centre being authorable and marked in both states, stable reading order in the stored array, the focus chain, both directions between the grid and the coordinate line, the caption following the anchor through the real spinbox, and the hint-parsing bug above. Four mutants each redden their target: writing the click to a copy, unclamping the minus, deriving the span from the cell count instead of the extent, and restoring the `"Vector2i"` comparison. Locally `dev` and `weapons` ran green at 693 cases, which includes the window-fit law and the tips coverage law.

**Blind spots, stated.** A headless suite cannot see whether it looks right. Cell size, the legibility of the centre marker, and whether the caret reads as forward are yours to judge in the build. The forward marker is deliberately ASCII rather than an arrow glyph, since the dev window uses Godot's default theme font; if it reads poorly it is a one-character change. I also have not measured the grid against the panel at the 15-wide cap in a running window, only against the fit law.


Phaazoid:
🤖 Claude says:

Three commits added after you found this in play. The grid was unreachable for a brand new attack, which made the whole feature look absent.

## What was wrong

Two faults behind one symptom, and both are mine.

**New built an attack with no pattern**, so there was nothing for the grid row to draw. That alone was survivable, because the resource picker is supposed to let you give it one.

**The resource picker could not give it one, and lied about it.** `OptionButton.add_item` auto-selects the row it is handed, so a null field displayed `AttackPattern` as though it were set. And Godot only emits `item_selected` on an actual change, so re-picking the row already showing does nothing. With #803 collapsing three pattern classes into one, that single row was the only row, and there was no click anywhere in the control that could assign a pattern. Measured all three facts headlessly rather than reasoning about them.

The lie predates #803 — the picker has always shown a selection for a null field — but it was survivable when three classes meant you could pick a different one. Collapsing to one turned a confusing control into an impossible one.

## The fix

- **New arrives with geometry.** An attack with no pattern reaches bare adjacency and is a legal but useless starting point, so the grid is now there to draw on immediately.
- **The picker carries a `(none)` row**, selects whichever row matches reality, and can both show and set the empty state. That state is one `AttackLint` deliberately passes, so it has to stay reachable rather than becoming a one-way door. `ItemEditorTool.NO_MAIN_KEY` is the same idea one domain over, and its comment already gave the reason.

## A mutant that passed, and what it taught

Deleting the explicit `select(0)` I had written beside the `(none)` row reddened nothing. With `(none)` added first, `add_item`'s own auto-selection already lands on it for a null field, so the line was dead and the **ordering** is what does the work. The comment says that now, the dead line is gone, and a case pins `(none)` at row zero — moving it to the end reddens.

## Verification

The `dev` folder is green at 465 cases. Four mutants, each reddening its target: New not seeding, the `(none)` row removed, `(none)` moved last, and the grid cases from the original diff. One mutant passed and is described above.

**Still yours to judge in the build**, unchanged from the PR body: whether the grid looks right, and whether the caret reads as forward.

**Also worth knowing:** the pattern picker is now nearly vestigial, since every resource-typed field in the editor has exactly one candidate class. You raised that, and the answer is a separate ticket rather than more of this one — see the shape library discussion.

— Claude (Opus 5) · 2026-09-06

Issue #810, opened 2026-09-06
Reactive fire ignores readiness: a dry Carbine's watch fires, and a cold Spitter's fires charged


Phaazoid:
🤖 Claude says: split out of [#97](https://github.com/Phaazoid/Godoiosis/issues/97) as a **declared residual**, not discovered afterwards — the plan named it before building and the build left it standing deliberately.

## The hole

**Reactive fire never asks whether the weapon can pay for it.** `PlanResolver._watch_triggered_by` checks spent/intact/covers/faction/lifecycle/anchor and **never readiness**; a counter derives through `Unit.get_counter_attack()` → `effective_main`, which reads live state at derive time.

Two consequences, and the Carbine has had the first one since #84:

- **A dry Carbine's standing watch still fires.** `consume_readiness_for` clamps at `maxi(0, …)`, so the magazine is already empty and the shot happens anyway.
- **A Spitter's watch fires CHARGED for free.** A watch's attack is frozen at arm time (`Watch.attack`), so a watch armed while the tank was hot keeps the empowered form after a counter in an earlier pass drained it.

## What #97 DID fix, so this is narrower than it looks

The **chip cannot lie**. `_Hypo.charges` threads the count across a resolve pass and `ResolvedOutcome.charge_spent` is stamped only while the pass still has a charge to give, so a counter and a watch shot in one pass cannot both claim the same one — exactly one wears the Tank badge, and the tank cannot go negative.

What is left is that the second shot still fires the **stronger geometry** having paid nothing.

## Why it was not fixed with the tank

The frozen stamp is the mechanism Law #2 rides on: `declare()` freezes what will fire, and execution is pure playback. Re-deriving a watch's attack at trigger time is a change to that doctrine, and it belongs to a ticket that can weigh it for **both** families rather than one that arrived carrying a spitter.

## The fork

- **Refuse the shot** when the weapon can no longer pay — the Carbine reading, and it makes a dry watch simply not fire.
- **Downgrade the shot** — re-derive the form at trigger time, so a cold tank fires the baseline. Costs the frozen-stamp guarantee.
- **Leave it**, and say so in canon: a standing watch is loaded when it is armed, and what it fires is settled then.

## Cross-refs

[...]

Issue #808, opened 2026-09-06
Attack shapes are their own named resource, shared by reference


Phaazoid:
🤖 Claude says:

## The ask (dev, 2026-09-06, found while authoring on the new grid)

> *"I wanted to save shapes as their own type of resource that could be loaded in and re-used."*

and, from the umbrella's own opening ask:

> *"Right now, cleave and line are just patterns that always originate at a manhattan distance of one. What if I wanted to fire a line off at a range? There's no way to specify that now, and it feels silly to create another attack type for it when the current line already exists."*

The motive is not only reuse. It is a CENSUS:

> *"I'd still like to build a little collection of them as I make new attacks. Being able to see them all in a list helps me visualize how much variety I've added to attack patterns in total."*

## Why the current picker cannot answer it

Cleave and Line are the display names of ATTACKS, not of patterns. Cleave is `Sweeper.tres`; Line is `POINT.tres`, whose attack is named Lance. **A pattern has no name and no file of its own** — every one is an anonymous inline sub-resource, and there are zero standalone pattern files in the project. The `attack_pattern` control is a SCRIPT-CLASS picker, which is why it listed `ManhattanRangePattern` and friends before #803 and lists one entry now.

Eight attacks carry a shape at all today; everything else is a single cell at some range:

| Attack | Shape |
|---|---|
| Cleave | three across, one ahead |
| Splash | three across, two deep |
| Lance, Line Snipe | six ahead |
| Zap Cannon | five ahead |
| Overwatch | four ahead |
| Freeze | three ahead |
| Spring | two ahead |

Lance and Line Snipe are byte-identical geometry authored twice, which is the duplication a library removes.

## The ruling (dev, 2026-09-06)

**Shared by REFERENCE, like a weapon family template** — not copied in. Chosen against the stated alternative:

> Editing a saved shape changes every attack using it. The folder is then an exact census: every shape in use is there and nothing in it is stale. To diverge, Save As forks a new named shape.

The cost was stated and accepted: retuning a shared shape reaches every user, so Lance and Line Snipe either share a destiny or get forked apart at migration.

**The project already splits on this question and lands here.** `ScenarioData.look_preset` stores a NAME rather than a reference because a scenario is a SAVE FILE. `WeaponData.main_attack` is a direct reference because a weapon family is AUTHORED CONTENT. An attack is the second kind, so a reference is consistent.

[...]

Issue #744, opened 2026-09-04
can_equip returns a reason, and the stash shows incompatibility before the attempt


Phaazoid:
🤖 Claude says:

Dev, 2026-09-04, on why compatibility is shown before the attempt rather than after:

> "we don't want people creating loadouts they think will work, only to get a nasty surprise when they start the scenario."

**This one does not depend on the screen existing** and can ship ahead of its position in #731 — the in-battle equip door gets the same sentence for free.

## The gap

`EquippableData.can_equip(wielder)` is the polymorphic gate, and it returns a bare `bool`. Two kinds override it:

- **`ArmorData`** — stat minimums read off the wearer's **body**: base → limb → jobs → temporary effects, and *never* gear (#112, amending #55/#89). That is what keeps `_enforce_gear_gates` to a single pass with no cascade.
- **`RuneData`** — at least one channelable carving, never all (#157). A dead rune is carryable, not wieldable.

A screen built on a bool can say **"incompatible"** and nothing else. For armor that means the player cannot tell whether they are two points of CON short or holding something their build can never wear.

## The change

Widen it to a **reason string**, following the precedent already in the codebase rather than inventing one:

- `WeaponInstance.fit_block_reason(index, mod)` → `""` or *"Space 2 holds 1 of 2 — this needs 2 more."*
- `TransmutationData.channel_block_reason(wielder, temper)` → the same shape.

So: a `can_equip_reason(wielder)` carrying the sentence, with `can_equip` becoming the empty-string test over it. One rule, one place, both callers.

Armor's sentence should name the stat and the gap; a rune's should say what it would need to be channelable.

## The proactive half

The stash (#S7) marks incompatible entries and shows the reason on hover, rather than waiting for a failed drag. Same for a unit card's own inventory when something it holds has *become* unwearable — which is a real state: a gate reads the body, so a debuff can strip a unit's kit mid-battle, and `_enforce_gear_gates` already handles the taking-off.

## Verify

Every override returns a non-empty string exactly when it returns false, and empty exactly when true — that invariant is the whole seam and it is cheap to pin across all kinds at once. Then assert the in-battle equip refusal surfaces the same sentence the stash does, so the two surfaces cannot drift.

Parent: #731 · Related: #112, #157, #55, #89

— Claude (Opus 5) · 2026-09-04

Phaazoid:
**Scoping note, 2026-09-05 — the proactive half moved before this ticket was built.**

[...]

Issue #697, opened 2026-09-02
Carried pure materia: the vial item and the per-cast burn


Phaazoid:
🤖 Claude says: filed from the **materia grill** (2026-08-29; co-dev signed off 2026-09-02). Doctrine lives in the canon-sweep issue; this is ruling **6**, and it is what #97's spitter injection consumes.

## The ask

**Carried pure materia** as an inventory item: **element-tagged vials**, plus rare **alkahest-pure** that matches anything.

## Rules

- **Burned per empowered cast, as part of the cast.** No separate action, no separate verb — declaring the cast spends it.
- **Grants exactly what a source grants.** Not more. "Carried is stronger" was considered and rejected at the grill: it breaks the binary empowered read, reintroduces stacking questions, and quietly tells the player the terrain game is the budget option.
- **Portability is the premium.** The vial buys empowerment *where there is no source* — that is the whole of its value, and it is enough.
- **Supply is the authored, faucet-free economy** (`progression.md`: each mission grants exactly what the designer placed, on a drawn curve). Nothing here is a faucet, and there is no crafting.
- **Alkahest-pure matches any element**, mirroring the vein split.

## The rhyme it preserves

This is the already-canon *pay-up* shape, one step wider: **materia substitutes for position the way it substitutes for aura.** Mid-battle re-flourishing already worked this way (adjacent matching source, or burn carried to do it anywhere) — see #513, whose pricing this settles.

## The shared economy is deliberate

The same vial serves the alchemist's empowerment burn **and** the spitter's injection (#97). That squad-supply tension is the intended design.

It is only safe because **both claims are marginal.** The dev caught the skew at the grill: if one item is both a one-time superpower *and* a requirement to function at all, the existential claim always wins and the alchemist's fun pays the tax. The doctrine's one law — materia never gates function — is what makes neither side bricked without it, and therefore what makes the shared item a real choice rather than a forced one.

## Law #2

The burn is part of the cast, so **the queue must show it** — which vial, spent when. A plan that spends a vial the player did not see spent is the queue lying. This matters more than usual here because the same pool is being drawn on by two different systems, so the player must be able to audit it before pressing Execute.

## Open

- Whether a vial is a plain `Item` or needs its own type — it needs an element tag and it is consumed, neither of which `Item` currently expresses.
- Stacking and count display in the inventory panel.
[...]

PR #811, opened 2026-09-06
#97: the Chemical Spitter's tank, Tank Injection, and the Corrosion element

### LINKED ISSUE #97: Chemical Spitter signature mechanic: baseline spray + materia injection
🤖 Claude says: split out of [#84](https://github.com/Phaazoid/Godoiosis/issues/84) at the dev's direction (2026-07-25). #84 closed at **6 of 7** families; the Chemical Spitter was deliberately deferred rather than held open, because it can't be built ahead of a content pass that hasn't happened yet.

## Why this is its own issue

Every other family signature was buildable from substrate that already existed (or that the mechanic itself paid for — Rev made DEF real, Pummel made displacement real, Burrow made terrain-creation real, the Carbine made the reload action generic). The Spitter is the one that **depends on undesigned content**: it needs materia to exist first.

## Current sketch (dev capture, 2026-07-23 — *not* ratified)

The generic spitter is **filled with materia** to output an **elemental AoE cone** in whatever element you loaded. **Prototypes are element-locked** but more efficient with their own element than the generic is with anything.

Its family row in [`docs/design/weapons.md`](../blob/main/docs/design/weapons.md) still carries the older framing — *Bio-hazard*: high front-loaded close damage, the status-delivery family (fire/ice/shock/corrode/heal variants). Whether the materia cone **replaces** that or **sits alongside** it is an open dev call, the same fork Rev and the Carbine each resolved by parking the older framing rather than deleting it.

## The fence

⚠ **Blocked on the materia pass (parked grill 9, [`docs/design/grill-queue.md`](../blob/main/docs/design/grill-queue.md)).** The dev's own flag on this one is *"still needs workshopping"* — materia has no design and no code, so a "fill it with materia" mechanic would be inventing that system sideways through a weapon. Don't build ahead of that fence.

## What's already in place

- `ChemicalSpitterWeaponInstance` exists ([#82](https://github.com/Phaazoid/Godoiosis/issues/82)) and is now the **only** remaining pure pass-through subclass — battle-scoped state has somewhere to live the moment there's state to hold.
- `Resources/WeaponAttacks/MainAttacks/Chemical_Spitter.tres` is still a **stub** (no display name, no power, no pattern) — real attack content is part of this issue, not a separate one.
- The elemental substrate it would ride is built: `Elemental.Element`, per-attack `elemental_damage_type`, reactions, terrain deposits, and gear-side elemental immunity ([#89](https://github.com/Phaazoid/Godoiosis/issues/89)). A cone pattern would be a new `AttackPattern` subclass — the geometry layer is unaffected by everything else here.

## Suggested order when it unfreezes

1. Materia pass (grill 9) resolves what materia *is* — then come back here.
2. Dev call on cone-vs-Bio-hazard framing.
3. Cone `AttackPattern`, real main attack `.tres`, and whatever state the loaded element needs on `ChemicalSpitterWeaponInstance`.

Doctrine for the family table lives in [`docs/design/weapons.md`](../blob/main/docs/design/weapons.md) (*The seven weapon families*); mod ideas tagged `[SP]` are banked in [`docs/design/weapon-mod-ideas.md`](../blob/main/docs/design/weapon-mod-ideas.md).

— Claude (Opus 5) · 2026-07-25

Phaazoid:
Closes #97 — the last of the seven weapon-family signatures. #84 closed at 6 of 7 and deferred this one behind the materia pass; #693/#694/#697 landed that pass, and this builds on top of it.

## What it does

**The tank.** `ChemicalSpitterWeaponInstance` was the last pure pass-through. It now banks supercharged shots, starting **empty** — so a mission's first injection is a real decision about a scarce item.

Deliberately **not** the readiness seam. `requires_readiness`/`consumes_readiness` express *gating*, and the one law of materia is that it never gates function: a dry spitter fires its ordinary Spray forever, unmetered. Wiring the tank through those flags would be the Kinetic Mace's mistake in reverse.

**Injection.** The first rearm in the game that consumes inventory, which is why `can_reload`/`reload` now take a wielder and `reload_block_reason(wielder)` is the rule the boolean derives from (`can_equip_reason`'s shape, #744). Injecting takes the element-specific vial over alkahest when both are carried.

**The supercharge is AUTHORED, per the ruling in chat** — "definitely something authored other than just power up", changing range. `WeaponAttackData.empowered_form` is what an attack *becomes* while its weapon is hot: a whole attack, so it can differ in range, shape, power, anything. A number added to power could never have said that. It is the weapon-side twin of the empowered form a carving authors at a source (#695).

**Substituted at one point** — `WeaponInstance.effective_main` — so the menu, `Reach`, the resolver, the queue row and the AI all see the charged attack with no per-site knowledge of a tank.

**Corrosion becomes a real element**, inert: no reaction, no paired `State`, no terrain source, never a sigil. A **Vial of Vitriol** is its materia — the name canon already banked in `elemental-interactions.md`.

## Three rulings taken in chat, folded in

1. **The baseline is a regular attack**, not a weaker form — materia supercharges it, exactly as it does a cast. This repealed wording in two canon docs.
2. **Corrosion vial rather than widening alkahest** — though `VialData.matches()` still lets alkahest charge any spitter, since canon's word is that it matches anything.
3. **The injection verb lists greyed with its reason, for every family.** A hidden row cannot tell "the tank is full" from "you carry no matching vial". Consequence you will see: **a full Carbine and a wound Springspear now light the Weapon Action button** for a greyed Reload row, where before they showed nothing.

## What the Fable review changed

Thirteen findings; the confirmed ones are in the diff. The four that changed the design rather than the wording:

- **`_mods_for` keyed on `attack == effective_main`**, and substitution made that answer time-dependent for the first time — a stamp of the *other* form would silently drop every MAIN_ATTACK mod's power, element, kind and knockback. It asks `is_main_form` now (both forms are the main).
- **The charged form would have been unreachable from the Attack Editor.** A lone object `@export` auto-renders as a swapper that can only `.new()`, embedding an inline sub-resource no catalog can see, and its nested editor writes into an object Update never saves. It gets `replaces_main`'s catalog dropdown instead.
- **Reactive fire can double-draw.** Counters and watch shots resolve before anything executes, so a live read would let two of them claim one charge. The count is threaded on `_Hypo` and stamped as `ResolvedOutcome.charge_spent`.
- **`play_session._apply_attack` returned before its post-fire economy on a null target**, where `execute()` has always run it — so a headless cell shot rearmed for free. Pre-existing for the Carbine; fixed here.

## The falsification finding

**Two mutants, because one cannot cover both twins — and that is the finding, not a precaution.**

| Mutant | Result |
|---|---|
| delete `play_session`'s tank spend | `tests/play` **reds** on `test_a_fired_charged_shot_empties_one_tank_slot` |
| delete `AttackAction.execute`'s tank spend | `tests/play` **and** `tests/weapons` stay **green, 326/326** |

The second is why `tests/flow/test_spitter_tank_execute_wire.gd` exists: the play path never calls `execute()` at all, so a suite written only against the headless twin would have shipped the hole. #697's mutant found the twin *missing* a spend and the play case caught it; here the game is the side nothing was watching. Both mutants reverted.

## How to check this

**You are on `feature/97-spitter-tank` — launch it.**

1. **Give somebody a Chemical Spitter and a Vial of Vitriol** (both in the dev Item/Unit editors; Vitriol is also in the Company stash). Open the ring → **Weapon** slice.
2. **You should see "Tank Injection" listed and GREYED**, saying *"No matching vial to inject"*, if they carry no vial — that greying is the ruling, and the row existing at all is what `has_weapon_actions` now allows.
3. **With a vial: inject.** It costs the main action for the turn (same price as a Carbine reload), the vial leaves the inventory, and the weapon's status line reads `Tank 3/3`.
4. **Now aim.** The attack should be **Pressurised Spray at range 3**, not Spray at range 1 — that is the substitution. Queue it and look at the queue row: it wears a **Tank** chip saying *"Spends a supercharged shot"*.
5. **Cancel it.** The tank should still read 3/3 — nothing is spent until Execute.
6. **Execute, then aim again.** 2/3, and still the charged form. Fire three times and it drops to the ordinary Spray — which still fires, at full ordinary power. It never bricks.

**What the tests already cover:** the injection rules and refusals, the substitution and its mod interaction, the snapshot round-trip, the spend through *both* executors, and that one pass cannot claim one charge twice. 650 green across weapons/flow/play, plus ui/law/items/dev.

**What they cannot:** anything about how it *feels* — `TANK_SIZE = 3` and the charged form's `max_range = 3` are placeholder numbers, tunable in the Attack Editor. And the charged form's geometry deliberately does **not** anticipate #808's cone; re-basing it on that stamp is the one content follow-up this hands off.

## Declared residual, filed as #810

Reactive fire ignores readiness: a watch armed while the tank was hot still fires the charged form after a counter drained it, having paid nothing. **The Carbine has had the identical hole since #84** (a dry Carbine's watch still fires). The queue cannot *lie* about it — the threaded count means exactly one shot in a pass wears the spend — only the geometry is free. Fixing it means re-deriving a watch's frozen stamp at trigger time, which is a change to the doctrine Law #2 rides on and belongs to a ticket that can weigh it for both families.

## Not in scope

`Spray`'s geometry, which is #808's cone. This diff touches its element, power and `empowered_form` and leaves `min_range`/`max_range`/`stamp` alone.


Phaazoid:
🤖 Claude says: **rebased onto #808, which merged while this was building — and the collision was worse than a text conflict.**

#808 replaced the geometry model this branch was written against: range (`min_range` / `max_range` / `max_and_a_half`) moved **onto `AttackData`**, the shape became a shared `AttackShape` resource beside it, and `Classes/weapons/patterns/AttackPattern.gd` is gone.

The one that mattered: `Chemical_Spitter_Charged.tres` embedded an `AttackPattern` sub-resource, so after the rebase its `ext_resource` was **dangling** — a hard parse error that fails the whole file, and takes `Chemical_Spitter.tres` down with it for naming it. Re-authored on the new model: the range sits on the attack, the shape is null (the anchor cell alone), which is what its one-cell stamp always meant.

Also moved: the two new suites' fixture preload (`pattern_fixtures.gd` → `shape_fixtures.gd`, whose builders take the *attack* now), plus a conflicting rewrite in `AttackEditorTool`'s skip list and `CLAUDE.md`'s geometry paragraph — #808's wording kept in both, this ticket's rows folded in.

**Spray's own geometry is still untouched**, as promised. Its `min_range`/`max_range`/`attack_shape` are #808's to author; the charged form's range stays deliberately provisional and is the one content follow-up this hands off.

Re-verified on the new base: **weapons/flow/play 652 green, ui/law/dev/items 1171 green.**

— Claude (Opus 5) · 2026-09-06

Issue #724, opened 2026-09-03
Reorder the pause menu: the list is a build history, not a decision


Phaazoid:
🤖 Claude says:

*(From the 2026-09-03 scratchpad sweep. The dev's entry was three words — **"reorder the pause menu"** — with no order given, and at filing he chose to leave the ordering unpicked: **the decision IS the ticket**, the [#421](https://github.com/Phaazoid/Godoiosis/issues/421) shape.)*

## The order today

Read straight off `PauseMenu._build`, whose `_add_button` call sequence **is** the order (it is a plain vertical `BoxContainer`):

| # | Row | Gate |
|---|---|---|
| 1 | Resume | always |
| 2 | Restart Mission | `can_restart` — hidden on the Sandbox |
| 3 | Save Game | rides Restart's gate (#144) |
| 4 | Load Game | always; **greyed** with a reason when no save exists (#166) |
| 5 | Return to Title | always |
| 6 | Glossary | always (#135) |
| 7 | Settings | always (#350) |
| 8 | Report a Bug / Feedback | always (#131) |
| 9 | Quit Game | always, tinted red |

## The thing that makes it read wrong

**Row 5 leaves the board. Rows 6–8 do not.** Glossary, Settings and Report are read-only detours that hand the player straight back to this same card — `game._open_pause_menu` reopens itself after each one, deliberately. So the list currently puts the second-most-destructive row in the middle of three rows that cost nothing, and the two ways *out* of the game (Return to Title, Quit) are split apart by them.

Each row was appended as it was built — #132, then #144, then #135, then #350, then #131 — so the order is a build history rather than a decision. That is the whole finding.

## Candidate principles

**(a) Exits last, in blocks** — nothing that leaves the board sits above something that doesn't, and the two exits land together at the bottom, escalating:

```
Resume · Restart Mission
Save Game · Load Game
Settings · Glossary · Report a Bug / Feedback
Return to Title · Quit Game
```

**(b) Frequency-first** — most-reached rows at the top regardless of kind (Resume, Settings, Restart, …). Optimises the common case at the cost of the grouping.

**(c) Something else entirely** — the dev's call.

## What it costs to build

- **The reorder itself is a re-sequence of `_add_button` calls.** No new machinery.
- **Blocks are not free.** `ModalCard._build_button_row(parent, vertical, separation)` gives one `BoxContainer` with a single separation value — grouping needs either separators added to the base or a nested row per block. Principle (a) needs this; (b) does not.
[...]

PR #819, opened 2026-09-07
#723: Esc opens the pause menu everywhere, except the title screen

### LINKED ISSUE #723: Esc on the title screen opens a bug report card: the title needs its own answer
🤖 Claude says:

*(From the 2026-09-03 scratchpad sweep. The dev's entry was **"no pause menu on title screen?"**. The sweep checked the premise first: the pause menu structurally cannot reach the title screen — but Esc there does something nobody decided, which is the real ticket. Dev's call at filing: **the title screen needs its own answer for Esc, filed as a design question**.)*

## What Esc actually does at the title screen

Traced, not assumed:

1. `MissionController.open_mission_select` sets `game.game_state = GameState.MENU`.
2. `MissionSelectScreen` deliberately does **not** claim `ModalLock` — that is load-bearing, not an oversight (a Game left DISABLED behind it would never run the mission picked next; pinned by `test_return_to_title_lands_on_the_menu_with_the_game_thawed`). So the game subtree is not paused, and `game._input` keeps running underneath the title screen.
3. Esc reaches `game._input`. `ModalLock.any_open` is false, so it falls through to the branch:

```gdscript
if _board_locked_for_player():
	open_report_card(BugReporter.Kind.BUG)
else:
	_open_pause_menu()
```

4. `_board_locked_for_player()` is true via `menu_is_up()` (`game_state == MENU`).

**So pressing Esc at the title screen opens a bug report card, defaulted to `Kind.BUG`.** The pause menu is unreachable there — the premise behind the scratchpad note is correct, just not for the reason it sounds like.

## Why that is a defect and not merely a surprise

**The title screen's own Feedback button defaults to `Kind.FEEDBACK`, on purpose, with the reason written down six lines away in the same file:**

```gdscript
# Defaults to FEEDBACK, not BUG: nobody reaches this screen mid-defect (#131 item 6).
_select_screen.feedback_chosen.connect(func(): game.open_report_card(BugReporter.Kind.FEEDBACK))
```

The Esc path contradicts it, and reports `GameState.keys()[MENU]` as the state the player was in.

**The justification for the report branch has gone stale in exactly one of the three cases it covers.** Its comment reads:

> *A locked board is an AI turn, a finished mission, or the menu — exactly the moments a stranger most wants to complain, and the ones the pause menu cannot serve.*

That holds for an AI turn and for a finished mission. It does not hold for the **title screen**, which is also "the menu": a stranger who has just booted the game has not seen anything to complain about yet. `menu_is_up()` covers `MENU` and `MISSION_OVER` together, and the title screen rode in on a predicate written for the other one.

## The question to answer

What *should* Esc do on the title screen? Three readings, none picked:

- **(a) Nothing.** The cheapest and the most conventional — a title screen's Esc is a no-op almost everywhere. Costs the stranger nothing, since the screen already carries a visible **Feedback** button that reaches the right card with the right default.
- **(b) Confirm-quit.** Esc as "back out of the game" — the one thing there is left to back out of at the front door. Needs a `ConfirmCard`, which already exists.
- **(c) Close whatever is on top, else nothing.** The Glossary and Settings pages open *over* the title screen and answer `ui_cancel` themselves; a title-screen Esc rule has to not fight them. Worth stating whichever option wins.

Either way the change is narrow: the report branch keeps `AI_TURN` and `MISSION_OVER` and stops claiming the title screen, which most likely means splitting `menu_is_up()`'s two states at this one reader rather than everywhere.

## Sharp edge

`menu_is_up()` is read by four callers and `_board_locked_for_player()` composes it into nine more. **Only this one reader is wrong** — the board *is* correctly locked at the title screen for every other purpose. Narrow the branch, do not narrow the predicate.

Cross-refs: `game.gd` (`_input`, `_board_locked_for_player`, `menu_is_up`) · `Classes/flow/MissionController.gd` (`open_mission_select`) · `Classes/ui/MissionSelectScreen.gd` · [#131](https://github.com/Phaazoid/Godoiosis/issues/131) (the report affordance) · [#132](https://github.com/Phaazoid/Godoiosis/issues/132) (the player-facing shell)

Phaazoid:
Closes #723.

Esc forked on `_board_locked_for_player()` and opened a BUG-defaulted report card on the far side. That predicate covers **four** situations, not the three its comment named — the title screen, the pre-mission screen, a finished board being inspected, and an AI turn — and neither `MissionSelectScreen` nor `PreMissionScreen` claims `ModalLock`, so `game._input` kept running underneath them and took that branch. Hence both sightings.

**The rule now: Esc opens the pause menu whenever a board exists, and at the title screen it does nothing.** Cards keep first claim through `ModalCard._on_cancel`, so a held item in the pre-mission screen still drops on the first press and the menu opens on the second.

`menu_is_up()` and `_board_locked_for_player()` are **not** narrowed — they are right about their nine other readers (#740's board lock among them). `MissionController.mission_select_is_up()` asks which *screen* is up, which is a different question.

## The title screen's feedback door

You asked for a dedicated feedback option there. **One has existed since #131** (`MissionSelectScreen.gd:129`, ungated, `Kind.FEEDBACK`), and the card has a kind toggle, so a bug is filable from it too. It was **below the bottom edge** — #418's bug one screen over.

`ModalCard`'s unframed path returns a bare `CenterContainer`, which centres its child at the child's *minimum* size and overflows both edges with no scrollbar. Measured by the test before any fix: **778px wanted against a 720px design space** with no save on disk, ~832 with one — so *Send Feedback* and *Quit Game* sat off the bottom. Not window-size dependent, which is what made it worth fixing: `GameSurface` lays the UI out in a design space that is exactly 1280x720 for **any** 16:9 window, so a maximised 1440p monitor has the same 720px as your editor. The frame step is overridden to a full-rect `MarginContainer` and the mission list expands into whatever the fixed rows leave — the list is the right part to give, and a row added later (#724's reorder) now costs list height instead of falling off.

## The AI-turn safety pass

You accepted this risk when you picked "everywhere", and it is real rather than theoretical. `ModalLock` disables the **Game node**, deliberately; every wait in `Pacing` is a **SceneTreeTimer** owned by the tree with `process_always`. So the freeze stops `_process` loops and has never stopped a coroutine — the pass would have played on behind a card calling itself PAUSED.

- `Pacing.beat` parks while a modal is up, **above** the headless escape. That placement is a correctness requirement, not tidiness: everything past that line is invisible to a headless suite, so a park written below it could not be tested and a mutant deleting it would pass.
- The pause menu greys Restart / Save / Load / Return to Title mid-pass, with reasons (#166) — those free the Unit nodes the parked coroutine resumes into (#107). Resume / Glossary / Settings / Report / Quit stay live.

Neither half is redundant: parking alone still resumes into a freed board, greying alone leaves the battle playing.

Two things the build found that the plan did not predict:

- **`mid_pass` had to be PASSED, not asked.** `_open_pause_menu` writes `game_state = MENU` on the way in and `playback_owns_board()` reads `game_state`, so by build time an AI turn no longer looked like one and the gate greyed nothing. Found by the test.
- **The restore became racy.** A pass that finishes under the card writes its own state; putting the stashed one back over it is the "leaves the board locked for good" failure the GLOSSARY and REPORT arms already warn about, arriving by a new road. `_restore_state` guards all six sites on `game_state == MENU`.

## How to check this

**What to do**

1. Boot to the title screen and press **Esc**. Then open Glossary or Settings over it and press Esc there.
2. Look at the bottom of the title screen: **Send Feedback** and **Quit Game** should both be fully on screen, and the mission list should scroll rather than push them off.
3. Start a mission into the pre-mission phase and press **Esc** with nothing in hand. Then pick up a piece of gear and press Esc twice.
4. Finish a mission, choose **Stay (inspect)**, press Esc.
5. During an enemy turn, press **Esc** mid-pass. Watch the board behind the card, then press Resume.

**What you should see**

1. Nothing at the title screen — no report card, no pause menu. Glossary/Settings still close on their own Esc.
2. Both rows visible. Resizing the window should not change that.
3. The pause menu, not a report card. With gear in hand the first Esc only drops it; the second opens the menu.
4. The pause menu.
5. The pause menu, with Restart / Save / Load / Return to Title greyed and each carrying a reason on hover; **the battle should be frozen behind the card**, and pick up exactly where it left off on Resume.

**What a failure looks like**

A bug report card appearing anywhere Esc is pressed. A pause menu over the title screen. Units still moving behind the pause card during an enemy turn — that is the park not holding. A board that stays locked after Resume (no End Turn button, clicks do nothing) — that is the restore guard. Any of the greyed rows being clickable mid-pass; worst case that crashes on a freed unit.

**What the tests already cover, and what they do not**

`tests/ui/test_escape_routing.gd` (9 cases) drives a real `InputEventKey` through `Input.parse_input_event` for every routing case above, and asserts the greyed rows carry non-empty tooltips. `tests/ui/test_title_screen_fits_the_viewport.gd` (2) measures the column against the viewport's own height, never a pixel count. Four mutants were run and each killed only its intended case: deleting the park, deleting the title carve-out, neutering the greying, and reverting the restore to a blind write.

**The blind spots are yours to play.** The suite never drives a *real* AI pass behind the card — `Pacing` collapses every beat to zero headless, so case 6 drives the `beat` seam directly and case 5's MISSION_OVER/AI_TURN preconditions are set on `game_state` rather than reached by playing. So "the enemy actually stops moving, and resumes correctly" is the one claim only you can confirm. Likewise nothing headless can see the title screen's *rendered* bottom edge; the measurement is of the layout, not of pixels.

Ran locally: `ui` 423/423, `core`+`ui` 498/498, `flow`+`ai` 469/469, all 0 orphans, exit 0. CI owns the tree.

## Noticed in passing, not fixed here

The suite log shows `user://saves/slot_1.tres` and `slot_2.tres` failing to load: they reference `res://Classes/weapons/patterns/ManhattanRangePattern.gd`, which #808 deleted. **Your existing save slots are dead**, and `ContentRepair` deliberately refuses to repair a missing *script* (a code fault, not content). Worth its own ticket if you had saves you wanted.


Phaazoid:
🤖 Claude says:

**Correction to the sidecar commit's own message (`ecab828`).** It says *"No law covers this ... so CI could not have caught it."* **That is wrong, and the opposite of wrong in a useful way.**

`.github/workflows/tests.yml` already carries **"Guard against sidecars the commit forgot"** (#659 follow-up): after `--import` it runs `git ls-files --others --exclude-standard -- '*.uid' '*.import'` and **fails** the job on anything untracked. So CI would have caught this on the first push; the dev simply got there first.

I checked before writing that sentence, but I grepped `tests/law/` for `.uid` and stopped at the empty result — the guard is a workflow step, not a gdUnit4 case, so the search was aimed at the wrong half of the repo. The rule for next time is the one Law #4 already states: **grep the question, not the location.**

Worth recording because the obvious follow-up is a trap. A `tests/law/` case asserting *"every `.gd` has a `.uid` beside it"* is the natural-looking fix and it is the **wrong design**, for the reason that workflow's own comment gives verbatim:

> *the import has already created it by the time gdUnit4 looks, so an on-disk assertion would pass vacuously in exactly the case it exists to catch*

The guard has to read the **git index**, not the disk, which is why it lives in the workflow and not in the suite. Nothing to build; the existing guard is correctly placed and correctly scoped (to the sidecar suffixes, so the downloaded engine binary in `godot-bin/` cannot red it).

The repo is currently **577/577** clean on the underlying invariant — every tracked `.gd` outside `addons/` has a tracked `.uid` — which is what makes the guard worth keeping loud.

I could not amend the commit message itself: that is a history rewrite on a pushed branch, and force-pushing is your call rather than mine. Say the word if you want it squashed out on merge; otherwise this comment is the correction.

— Claude (Opus 5) · 2026-09-07

Issue #835, opened 2026-09-08
A generic instance of every weapon template, derived rather than authored


Phaazoid:
## The ask

A plain, unmodded version of **every** weapon template should always be available for picking, without anyone authoring a file for it. Dev, 2026-09-08: *"I generally just want a generic version of each weapon always available for picking"* and *"since this is a thing I'll always want to happen, is there a way we can make this automatic?"*

## Why

Two problems, one cause. `WeaponCatalog.get_editable()` returns `get_saved()` -- the `WeaponVariants/` folder -- and `ItemCatalog.SOURCES` lists that same folder as the only weapon source. So the **template layer is invisible to every picker**, and the only bridge is hand-authoring one instance per template.

1. **The chore never ends.** Six families are bridged as of `9afb0803`; **eight designed weapons are still unreachable** -- the PROSTHETIC family base and all seven prototypes (Bow, Sweeper, POINT, Shock Rod, Water Staff, Firespitter, The Jaw). Every one has real identity and a named main attack; weapons.md calls all fourteen "instantiable end-to-end". None can be picked in the roster editor, the stash, the Unit Editor or the Spawn tab. Nothing warns you -- a weapon simply is not in a list.

2. **A hand-authored generic is a derived artifact stored as content.** It carries no authored decision: its whole identity is "this template, nothing fitted". That duplication already bit twice --
   - it **drifts**: the five new generics each hard-code the name their template already states, and `WeaponInstance.shown_name()` would have read it through for free, so renaming a family stops reaching its own generic;
   - it **collides**: a generic is named after its family, which is the most collidable name there is. `vroom.tres` carries `display_name = "Chainsword"`, so `ResourceCatalog.by_name` collapses it with the new `Chainsword.tres` and one of them is unreachable. That is also why only four of five new variants tripped the #597 sweep (see #833).

   It is also not what the folder is for: weapons.md defines `WeaponVariants/` as "saved NAMED fitted instances -- the get-attached-to-your-weapon feature". A generic is neither named nor fitted.

## The ruling this rests on

#80 narrowed `get_editable()` because *"a bare family template isn't a real carryable weapon"*, and the pickers going empty was recorded there as **"the intended effect of the new rule, not a bug"**. That rule stands. See the ruling comment below for its scope as re-stated 2026-09-08.

## Shape

[...]

Issue #745, opened 2026-09-04
Tooltips derived from the data, preview-at-decision, and a description pass on the base weapons


Phaazoid:
🤖 Claude says:

What every item, ability and stat says when the player looks at it. **Largely independent of the screen** — the tooltips it produces render in the existing inventory and inspect panels too, so it can ship ahead of its position in #731.

## The rule

**Numbers are read from the data at display time, never authored as prose.** Fit a mod and the readout says what `power_delta`, `weight`, `added_element`, `scaling_change` actually hold; retune it and every tooltip in the game changes with it.

This is the Glossary's own standing rule — numbers interpolated from the constants that rule them, never retyped, *"so tuning a value re-words the glossary for free"* — and `ArmorData` already does exactly this for its inventory tooltip (#44). A hand-written `description = "Adds 2 power"` is the #552 failure: the justification goes stale while the rule stays right, and the code compiles either way.

The dev's requirement, which this serves rather than fights:

> "Players will 100% want and need the full numerical changes any decision they make brings."

## Preview-at-decision

The stronger form, and the one that actually answers that sentence: hovering an item shows its effect **on the unit who would carry it**, not just its own stats. `STR 6 → 4 · DEF 3 → 7`. Fire Emblem does this; `jobs.md` already names preview-at-decision as the pattern for stat ceilings.

It is cheap here because #S4 makes every roster unit a real `Unit`, so the "with this equipped" number comes from the same `get_effective_stat` walk as the current one.

## The flavor half is a content pass, not a schema change

`Item` already has `description` and `icon`, and both already render — `inventory_panel.gd:241` puts `item.description` in a tooltip today, `info_panel.gd:234` does it for abilities. `WeaponData extends Item`, so weapons have had the field all along; it is simply unauthored.

So the work is writing, on a deliberately narrow scope. Dev, 2026-09-04:

> "most things won't get descriptions for now, sans the base versions of the weapons, almost everything is currently placeholder."

**Flavor never states a number** — that is what the derived block is for, and a flavor line that restates a value is the stale-description trap coming back in through a different field.

One wire to check: `WeaponInstance` copies `description` from its template at creation (`WeaponInstance.gd:98`), so authoring a line on a `WeaponData` does not reach instances already saved in scenarios.

## Verify

Change a mod's `power_delta` in a fixture and assert the rendered tooltip text changes with it — that is the property the whole ticket exists for, and it is the one a snapshot test of expected strings would quietly pin the wrong way round.

[...]

PR #839, opened 2026-09-08
#835: derive a generic instance of every weapon template (slice 1)

Phaazoid:
Closes half of #835 (slice 1 of 2). Slice 2 — `ItemCatalog`, the curated roster stash, `RosterTool` — follows.

The template layer was invisible to every picker. `get_editable()` returned `get_saved()`, so the only bridge from a template to something a unit can hold was hand-authoring a mod-less variant per template. Six families were bridged in `9afb0803`; **eight designed weapons were still unreachable** — the Prosthetic base and all seven prototypes (Bow, Sweeper, POINT, Shock Rod, Water Staff, Firespitter, The Jaw). Nothing warned about it: a weapon simply was not in a list.

`WeaponCatalog.generics()` derives one plain `WeaponInstance` per template instead. A new template file appears and its generic exists the same instant.

**Your #80 ruling is kept, not repealed** (recorded on #835, 2026-09-08): what a picker offers is a real `WeaponInstance`, never the shared template. Only who *creates* the instance changed — the catalog derives it instead of you saving a file.

## What this does NOT change

**Nothing duplicates, and no existing row moves.** An authored variant wins a name clash — the specific thing beats the fallback — and your saved variants already occupy all six family names. So the derived Carbine/Chainsword/Chemical Spitter/Drill/Kinetic Mace/Springspear are suppressed, and exactly **8 rows are new**. That is also what lets you delete the five generics from `9afb0803` whenever you like, as a separate step, without the list changing underneath.

## A defect this had to fix on the way

`UnitEditorTool._entry_matches` assumed a catalog entry is a **template**, so a generic entry never matched and it fell through to the name compare — where a generic's `display_name` is `""` by design, making the test `"" == "" and "" != ""` → false. A held generic would have read as **"(empty)"** beside a `holds:` label: the exact #80 symptom that label was typed for.

Two instances now match when **indistinguishable** — same template, same fitted mods. The mods half is not optional, or a fitted `Gun` would match the plain Carbine entry. It is also stricter than the name compare it replaces, which let two variants sharing a `display_name` match each other — the latent half of #833's `vroom` collision.

`Item.shown_name()` is promoted beside `describe()`, for `describe()`'s own stated reason (*"the field is storage, and for a WeaponInstance the stored value is usually empty on purpose"*, #745). Required rather than cosmetic: a generic has no name of its own, so `ItemCatalog.label_for` would otherwise fall back to the filename and render `"Chemical_Spitter"` / `"ChainSword"`. It collapses three `is WeaponInstance` branches that only wanted a name.

## #624 landed mid-flight, and it moved two things underneath this

Rebased onto it. Both interactions would have left assertions **passing for the wrong reason** rather than failing, so they are called out rather than quietly fixed (second commit).

- **`space()` no longer grows the array** — a weapon holding nothing answers with a throwaway, and `fit()` is the one door that grows. So `space(i).append(mod)` fitted *nothing*, and both suites' mod-fitting helpers were silently comparing two empty weapons. They fit through `fit()` now, asserted true at the door.
- **The lazy-grow case lost its premise.** Its hazard is real and is now stated as what it actually is: an instance saved *before* #624 carries `[[], [], []]`, which is the state that ticket's own `copy_for_grant` normalization exists for — so a weapon out of an old save must still match its own entry rather than reading as "(empty)". `_same_weapon`'s shape-immunity is re-justified against that.

## How to check this

**What to do.** Launch, F1 for dev tools, pick a unit, open **Unit Editor → Gear & Jobs**, and open an inventory slot dropdown. Then try the **Spawn** tab's weapon dropdown.

**What you should see.** Eight entries that were never there before — **Prosthetic**, and the seven prototypes **Bow / Sweeper / POINT / Shock Rod / Water Staff / Firespitter / The Jaw**. Everything you had before is still there exactly once. Pick one; the unit gets a real weapon with that family's main attack and empty mod spaces. Close the editor and reopen it on that unit: the slot should read the weapon's name, **not "(empty)"**.

**What a failure looks like.** A slot reading `(empty)` with `holds: Bow` beside it means `_entry_matches` is wrong again. A row labelled `Chemical_Spitter` or `ChainSword` (underscore, or that odd capital S) means the name is coming from the filename rather than reading through the template. Two `Carbine` rows means the authored-wins clause broke. A crash on picking means `instantiate_entry` got something that is not an instance.

**What the tests cover, and what they do not.** 18 cases pin the derivation, the sharing, the grant independence and the matching rule. **Six mutants run, six caught** — including one that reverts `get_editable()` and reds naming **exactly those eight weapons**, and one (`copy_for_grant` sharing its array) that *only* the ad-hoc fixture case catches while the disk sweep passes, which is why that case was split out.

But the suite asserts through `_item_catalog()` and `_entry_matches()`, **never through the built dropdown** — a row that resolves correctly and never draws would pass every case here (the #804 trap). So the dropdown actually listing them is yours to confirm. Two further gaps, stated rather than left to be found: the null-drop in `generics()` (an unmapped `weapon_type`) has **no test**, because `WeaponTemplateLint` BLOCKS that at the save door and authoring one would mean writing a fixture into a scanned content folder — the dirty-tree rule; and the **player-facing stash is untouched in this slice**, so the pre-mission loadout still offers only authored variants until slice 2.

## Canon sweep

Deferred to #835's close, per the policy's own rule for multi-session issues — sweep against the final state, not a slice. What it will owe is already named: `CLAUDE.md`'s `items/` bullet (`Item` gains `shown_name`), its `get_editable` line, and the #80 residue entry under Known debt.

## Testing

`weapons items dev flow ui law` — see the run below. No content file changed.


Phaazoid:
🤖 Claude says:

**The dropdown blind spot is closed** — the body's *"yours to confirm"* line no longer holds, and that is the only thing that changed. Rebased onto `9238462c` at the same time (no conflicts; #837/#687/#53 touched none of these files).

You confirmed the rows draw by hand. This makes that permanent, in the reachability law rather than beside the matching cases, because a name that never becomes a row is unreachable — which is that law's whole subject:

```
test_every_offered_item_draws_a_row_in_the_inventory_picker
```

It drives `_add_inventory_section` into a plain container, walks it for the picker, and requires every offered name to be listed. No overlay and no `editing_unit` needed — the section reads `_item_catalog()` and `_inventory` and nothing else. Sweeping what is *offered* rather than what is authored means one claim now covers vials, armour, runes, weapons and the derived generics at once, and it passes vacuously on an empty catalog.

**Falsified.** The mutant is the realistic version of the fault, not a blunt one — a filter that drops entries with no `display_name`, which is exactly what a derived generic has:

```
'Prosthetic' is offered by the catalog and no row draws it
'Bow' is offered by the catalog and no row draws it
... Firespitter, POINT, Shock Rod, Sweeper, The Jaw, Water Staff
```

Eight names, and the suite was green before it. That is #804's shape — a row that resolves correctly and silently never draws — and it is now the one thing in this PR that could not have been caught by reading the data layer.

`law dev` — 747 cases, 0 failures, 0 orphans.

— Claude (Opus 5) · 2026-09-08

Issue #80, opened 2026-07-21
Weapons broken: empty inventories, all-1-range attacks, no counters, dev-tool offers families not instances


Phaazoid:
Discovered during manual testing 2026-07-21 (unrelated to the action-registry work landing alongside this — combat weapons appear broken across the board). Filing symptoms only; **root cause not yet investigated.**

## Symptoms observed

1. **Dev-tool inventory shows empty for pre-authored units.** Selecting a unit in the Unit Editor that was authored with weapons in its inventory shows nothing there.
2. **Every attack shows a 1-cell Manhattan range**, regardless of the equipped weapon's actual `AttackPattern` (forward line, forward wide, etc. all collapse to the same 1-range fallback).
3. **No unit can counter-attack**, ever.
4. **The Item Editor's equip picker offers weapon FAMILIES, not instances** — when adding a weapon to a unit's inventory, the list being chosen from is `WeaponData` templates (`ChainSword`, `Springspear`, ...) rather than `WeaponInstance`s a unit could actually carry.

## Suspected area (unconfirmed)

All four symptoms plausibly point at the same root cause: fallout from the `WeaponData`/`WeaponInstance` template-vs-instance split (#59) and the subsequent multi-attack model migration (#72), which moved `power`/`attack_pattern`/`can_counter`/etc. off `WeaponData` onto per-attack `WeaponAttackData` (`main_attack` + `extra_attacks`). Symptom 2 (range always 1) and 3 (no counters) both look like reads that should resolve through a unit's `main_attack`/live-picked `AttackData` are instead hitting an unset or default `AttackPattern`. Symptom 1 (empty inventory) and 4 (families not instances) both look like the dev tooling (`Classes/dev/UnitEditorTool.gd` / `ItemEditorTool`) reading/offering the wrong layer of the template/instance split. Existing `tests/weapons/` suites are green (357/357 full suite passes), so this is either an in-game/dev-tool-only path the test suite doesn't cover, or a gap between scenario-authored data and what the split expects at load time — needs actual investigation, not assumed from this description.

## Scope note

Explicitly NOT to be worked in the same pass as whatever's currently active — filed as its own tracked item per the dev's request.

Phaazoid:
🤖 Claude says:

[...]

PR #844, opened 2026-09-08
#835 slice 2: a template is nameable, and grants as an instance

### LINKED ISSUE #835: A generic instance of every weapon template, derived rather than authored

Phaazoid:
**Stacked on #839** — base is `feat/835-weapon-generics`, so the diff here is slice 2 alone. GitHub retargets it at `main` when #839 merges. Closes #835 with it.

Slice 1 derived a generic per template and reached the dev pickers. This reaches the **player**: the pre-mission stash and the roster editor could offer only saved variants, so a mission's gear could not contain the seven prototypes or the Prosthetic base at all.

## Two lines, and the second is what makes the first safe

**`ItemCatalog` gains the two TEMPLATE folders.** It answers what a mission may **offer**, which is a question about *files* — a roster stores its picks as `ext_resource` paths, and a weapon's plain form has no file of its own; its template's **is** its identity.

**`WeaponData.copy_for_grant()` answers with `WeaponInstance.make(self)`.** `Item.copy_for_grant` is *"the one door for how a unit receives its OWN copy"*, and for a shared family template the honest answer to that question was never a copy of the template — a unit cannot hold one, and the default deep copy handed over a **forked** one, which is the fault #80 closed by taking templates out of the pickers. Closing it at the door instead means a template may be authored into any of the three `Item`-typed kit doors and arrive as the weapon it names.

**So your #80 ruling is intact rather than bent:** a bare template still reaches no unit and no stash. It is *nameable* in a catalog and *instantiated* on the way out.

`instantiate_entry` collapses to that door — its `is WeaponData` branch had become the same rule spelled twice.

## The guard that turned out to matter

Three grant doors gained a null check, and **not for tidiness**: `make()` answers an unmapped `weapon_type` with null, `Item.copy_for_grant` has never returned one, and — measured — **`Unit.add_item(null)` puts the null into the first free slot and returns `true`**. Shipped content cannot reach it (`WeaponTemplateLint` BLOCKS that at the save door), but without the guards a hole would be carried in silence. Mutant M10 covers it.

## Content

Per your call: the five generics from `9afb0803` **plus Drill and TorvArm**, which are the same shape and merely predate them. Fable's review caught Drill — I had found two referrers and there were **eight `ext_resource` lines across six files**, two of them load-bearing (Noemie's Carbine, Torv's mace).

Each repointed at its template, each verified plain first, and each written **without a uid** because every existing referrer of those templates has none — I checked the convention rather than choosing one. `WeaponVariants/` now holds only what `weapons.md` says it holds: named fitted instances and pet-named ones. No `by_file` key collides across the six sources any more.

**`Roster` and `RosterTool` needed no edit at all** — `Loadout.from_roster` already copies at the phase boundary and is `offered_stash`'s only caller, and RosterTool's rows are typed `Item` and store the ref. Two planned edits deleted as Law #4 hazards on review.

## How to check this

**What to do.** Launch, F1, **Project → Rosters**. Look at the Stash column. Then load a mission with a roster and open the pre-mission loadout.

**What you should see.** The stash column now lists every weapon **template** — the seven prototypes and the seven family bases — alongside the saved variants. Set a count on one, Update, then start that mission: the item in the phase stash is a **real weapon**, and its chip opens the fitting card with mod spaces you can drag into. Noemie still starts with a Carbine, Torv with his mace and both limbs, Bandit and Ross with Drills.

**What a failure looks like.** A stash item whose chip opens no fitting card, or a card with no spaces, means a template reached the phase uninstantiated. Torv missing an arm or a leg means the prosthetic repoint was wrong. Any mission failing to load at all means a dangling `ext_resource` — that would be loud, not silent.

**What the tests cover, and what they do not.** 10 new cases and **three mutants, three caught** (reverting the grant door, forking the template, dropping the null guard). `tests/flow/test_stash_grants.gd` asserts the **wire** — a `Roster` naming a template through to what `Loadout` hands the phase — rather than either end. But the suite drives `Loadout.from_roster` and `_seed_starting_kit` directly; **it never opens the RosterTool stash column or the fitting card**, so *that those rows draw and that chip opens* is yours. The content half is stronger: `test_scenario_load_integrity` loads every shipped mission and `test_resource_uid_references` grades every reference, so the eight repoints are machine-checked rather than eyeballed.

## Residual, your call

**`vroom.tres` still carries `display_name = "Chainsword"`**, so it occupies that name in the dev picker and the derived Chainsword generic does not list. Functionally identical — vroom is a plain Chainsword wearing a pet name, and a pet name is an authored decision — but the row you see is vroom. Say so if you would rather it went the way of the other seven.

## Testing

`items flow dev law weapons unit ui presentation` — **2591 cases, 236 suites, 0 failures, 0 errors, 0 orphans**. `presentation` included deliberately: content changed, which is the class CLAUDE.md warns reddens those suites.


Phaazoid:
🤖 Claude says:

**Retargeted at `main`, and the reason is a defect in how I set this up.** The workflow triggers on `pull_request: branches: [main]` only, so while this was based on `feat/835-weapon-generics` it got **no CI at all** — `gh pr checks` reported none. That is the stacked-PR trap, and I would rather not rely on a later auto-retarget re-triggering it: changing a base fires `pull_request` with action `edited`, which is not in the default trigger set.

**What that costs you in review surface:** this PR's diff now shows **both slices**, because slice 1 has not landed yet. The body's "the diff here is slice 2 alone" no longer holds.

- **Slice 2 alone is commit `76a41597`** — that is the one to read here.
- **Slice 1 is #839**, already reviewed, CI-green, and tested by hand.

Either merge order works and neither double-applies: merge #839 first and this diff shrinks to slice 2 on its own; merge this one first and #839 goes empty and closes.

Nothing about the code changed — this is a base and a body correction.

— Claude (Opus 5) · 2026-09-08

Issue #887, opened 2026-09-10
Electricity has no look: the sky strike, and the current arcing above the water it travels


Phaazoid:
SHOCK landed a mechanic in #884 and nothing on screen says so. A shock into a river pulls half a squad into the volley, the queue rows explain it perfectly, and the board plays three separate lunges with no current anywhere between them.

**The ask** (dev, 2026-09-10): *"The next step for this is giving electricity a visual effect. Something HD, popping, dramatic and strong feeling."* The rulings that shape it are in the comments below, one per comment, in his words.

## Why this is not one of the four tickets next to it

- **#455** is the *terrain* channel — what a BURNING or FROZEN tile looks like while it stands. This is an event that lasts half a second and touches no tile state.
- **#358** is the *sprite* stack — what a WET or frostbitten body looks like between turns. The arc's victims come away wet, and #358 is what will say so; this is the blow, not the mark it leaves.
- **#188** is *damage feedback* — shake, flash, numbers, for any hit. Already built for every attack via `UnitMirror`'s HP poll. This is the attack's own signature, which that poll cannot see: it knows an HP number moved and nothing about elements, routes or aims.
- **#278** is the *grading* stack — presets, weather, the look of a whole mission. An attack effect is what interrupts one, not part of one.

## What it actually builds

**The first per-attack event channel in the game.** Today the diorama learns that a blow landed by watching HP fall (`UnitMirror.report_impact`), which knows nothing about what hit or why, and `battle3d` connects to nothing on `OrderExecutor`. `ElementalReaction.vfx_tag` has been declared since the elemental system landed and is read by nothing. So the seam is most of this ticket, and every later attack effect inherits it — the same way #656 minted the particle seam that #551/#455/#358 now inherit.

A shock volley's impact publishes one event carrying its `AttackAction`, once per blast (`is_secondary_hit`'s own gate — the same one readiness, vial and watch spends already use). From that action the effect reads the aim, the attack, and the current's own route tree.

## The look

Ordered by the dev's ruling: bolts **above the water** first, the sky strike with them, the crawl through the water last. Every one is a knob on the Game tab's Elemental page, so the whole thing can be dialled to nothing and back.

| ingredient | what it is | slice |
| --- | --- | --- |
[...]

Issue #656, opened 2026-08-29
Battle zoom: the dust shockwave when the tiles thud home


Phaazoid:
## What

The exit thud lands silently. [#521](https://github.com/Phaazoid/Godoiosis/issues/521)'s ask was *"tiles thud back into their sockets **with a dust shockwave**"* — slice B built the thud and cut the dust, deliberately and on the dev's call (2026-08-28: **motion in, particles out**), because it is new machinery rather than more of the same.

Filed rather than parked: #521 closed with this in its Ask, so it needs a home of its own.

## Why it was cut

**There is no particle system anywhere in this project.** `grep` for `GPUParticles`/`CPUParticles` across `Classes/` and `Scenes/` returns nothing. The only burst that exists is `HealthBlockDebris`, which is hand-rolled cubes on its own `_process(delta)` clock.

So this is a fork, not a task:

1. **Hand-roll it on `HealthBlockDebris`'s pattern.** Consistent with the one precedent, no new dependency, and it already survives #520's hitstop correctly (it runs on a scaled delta). Costs a second bespoke particle-ish thing.
2. **Introduce `GPUParticles3D` properly**, and make it the answer for this, for #358's status effects, and for whatever the art pass wants. Bigger, and it is a decision about the whole presentation stack rather than about dust.

## What it can hang off, already built

**The landing moments are already discrete and already published.** Slice B bumps `BoardSpace.staging_version` once per tile **landing** — not per frame — precisely so version-gated pollers can react to a tile arriving. A dust puff is exactly that kind of consumer, and needs no new signal:

- `BoardSpace.advance_flight()` returns `true` on the step a tile landed
- `BoardSpace.flying_cells()` / `in_flight(cell)` say who is still in the air
- `BoardMirror.surface_point(cell)` is the one placement seam, and already carries the staged offset

The **exit** is where the ticket wants it (tiles hitting their sockets), but the entry's slam is the louder moment and has the same shape — worth deciding whether one effect serves both.

## Notes

- Timing has a home already: `Pacing`'s tear-out block, beside the four held beats slice B added.
- Whatever this becomes must read [#217](https://github.com/Phaazoid/Godoiosis/issues/217) if it flashes, like the white-out does.
- The dev's timings run long by design (~7–12s of transition per Execute), so a puff that reads at 0.3s in isolation may vanish against them — tune against his settings, not the old defaults.

Cross-refs: #521 (the transition), #410 (umbrella), #358 (the other customer for a particle layer).

[...]

PR #889, opened 2026-09-10
Sparks, a flash, and the current in the water (#887 slice 2)

Phaazoid:
Slice 2 of #887, stacked on #888 — **its base is that branch, so merge #888 first and this diff is just the three channels.** You said you would merge both before playing either.

The three ingredients that each cross a system the bolts do not: a GPU particle system, a white-out that already had a driver, and the water shader. What they share is **the arc's own clock** — every one is measured against the same elapsed seconds the bolts are, so a hitstop freezes the lot rather than three of four.

## Sparks

They mark a **body**, which is the half the bolts cannot say: the arcs draw the current over the ground, so they answer *which tiles are live*, and a burst on each caught unit answers *who was caught*. It is what stands in for the sprite blowout the plan opened with and had to drop.

They fire on the hop that reaches each body, and **the delay is read back off the stamped tree** rather than tracked beside it — whichever hop ends on a victim's cell already carries the step the current took to get there. So the sparks travel with the bolts by construction and there is no second schedule to keep in step.

**The scatter is shared with the slam dust, not copied.** `ParticleFan.scatter` is #656's fan with every number lifted into parameters; what genuinely differs between dust and sparks is the material and the knobs (additive rather than alpha, up rather than out). Two emitters that both mean *throw N particles from a point* are two answers to one question the moment either is tuned. `BoardSpace.effect_volume` is the same move for the cull box, and there it is more than tidiness — a stale copy of that volume draws **nothing at all** and no test can see it, which is exactly the zero-lit-pixel failure #656 measured.

## The screen flash

The sharpest thing in the slice. The white-out was built for the tear-out with **one** writer, so its driver's idle branch pushed a flat 0 — which with a second driver is one caller clearing the channel out from under the other. Each driver now keeps its own level, one function composes them, and the louder wins.

**Max rather than a sum**, because both are the same physical thing: adding them would make a shock struck during a tear-out brighter than either can be alone, past the cap the photosensitivity setting is enforcing. And the shock inherits that cap and the muted tint for free by driving the existing channel rather than a rect of its own — one accessibility gate, where it already lived.

## The crawl

The current seen **in** the water, which you ranked below the arcing above it. The filaments are the water shader's own caustic net at a tighter scale and a faster rate — one function, so a lake bed's light and a discharge cannot disagree about what a filament looks like — driven by a mask holding one texel per cell with that cell's own hop count, spreading on the same ring delay the bolts use.

It decodes through the board mask's **existing** rect (one board, one picture of where its cells are) and samples **nearest** where the board mask samples linear, because a hop is a category and a step interpolated between two cells is a hop that never happened.

Five new shader globals, declared as a **named closed category** on the water law rather than as an exception to it: they describe something happening *to* the water rather than a property of either water, so a deep/shallow pair could mean nothing and a Water knob row would be a second answer to values the effect already owns. That law's own comment predicted this shape.

**The `project.godot` edit round-trips through `ProjectSettings.save()` with zero churn** — measured with a throwaway tool scene, run and deleted, rather than argued from the format looking right.

## How to check this

The Ford again, with the Electric Rune. Everything is on by default; each ingredient has its own toggle on **Game tab → Elemental → Shock**, so the way to judge one is to turn the other two off.

1. **Sparks.** Shock the river with bodies in it. Each caught unit should throw a burst *as the current reaches it* — near ones first, far ones a beat later. Turn `Arc over the water` off and they should still travel in that order.
2. **The flash.** The screen should tick brighter as the blow lands, once, briefly. `Flash strength` is the dial; at 1 it is a full white-out, which is almost certainly wrong for a hit.
3. **The crawl.** Filaments spreading over the water surface under the bolts, outward, lingering after the bolts have gone. It only appears on **water** — a wet body standing on dry land gets a bolt and no crawl, which is correct and worth knowing before it reads as a bug.
4. **Photosensitivity on** (Settings): the flash should go muted and capped rather than disappearing, the bolts should hold one shape, and the sparks and the crawl are unchanged.
5. **A shock during a tear-out.** Both flashes are live and the screen should take the louder, never the sum.

**What a failure looks like:** sparks all firing at once (the stagger); no flash outside a tear-out (the frame-path push — see below); a crawl that freezes the moment the bolts die (the clock's liveness test); a crawl on the wrong cells or a cell short (the mask's encoding); or sparks that never appear at all, which on past form is the cull box rather than the emitter.

**What the tests cover.** 17 new cases. The tree read as a hop count per cell, the spark delay read off it, the stagger driven on an explicit schedule, the flash popping and dying, the crawl outliving the bolts *and its age still advancing after they do*, the mask encoding a hop per cell with zero everywhere else, the clock reaching the mirror's own push funnel, the crawl wearing the element's hue, and the two fans coming out different from their own knobs.

**Five mutants, five caught** — the stagger flattened, the mask's +1 dropped, the crawl left out of the clock's liveness test, the shared fan ignoring its parameters, and the frame-path push deleted.

**A sixth mutant PASSED, and that is the finding worth reading.** I first wrote the flash case calling the composition function directly, and restoring the old single-writer idle branch left it green — because `_process` recomputes from both fields afterwards anyway, so that mutant is behaviourally equivalent. The real hazard was never the `max`: it was that the transition's driver returns early the moment nothing is flying, i.e. on **every frame a shock is ever struck on**, so the second driver needed a push on the ordinary frame path. The case now drives `_process` and the deleted push reds it. *When a channel gains a second writer, test the frame that reaches it, not the function that composes it.*

**Blind spots:**

- **Everything visual is yours**, and more so than in slice 1: a GPU particle is simulated on the card and never read back, and headless a global shader parameter *registers* but stores nothing (`global_shader_parameter_get` answers null whatever anyone sets — #552 measured that, and it is why the crawl's cases assert through the mirror's push funnel instead). So nothing here has seen a spark or a filament.
- **The crawl's filament look is unmeasured.** Its scale and rate are consts in the shader, chosen against the caustics' numbers rather than against a screenshot; if it reads as caustics rather than as electricity, those two are the dials and they are the only values in this slice with no panel row.
- **The spark defaults are guesses**, particularly `Spark rise` and `Spark size` — dust's neighbours, which may be the wrong reference for something meant to read as light.
- **The crawl never appears on a wet body standing on dry ground.** The water shader draws it, so it is water-only by construction. The bolt still arcs there; only the crawl stops at the bank.

Ran locally: `presentation dev` (1250 cases), the water law with `elemental` (69), and the new suite alone after each mutant. CI owns the tree.

**Canon:** `presentation-effects.md` gains the slice-2 half of its #887 section, including the passing mutant. `terrain.md` notes the water shader's second kind of tenant and keeps its marker, the sweep having been for this topic only.


Phaazoid:
🤖 Claude says:

**Merge order, and why the base changed.**

This opened against `feat/887-arc-lightning` so its diff would be slice 2 alone. It is retargeted at `main` instead, because **CI only runs on a PR whose base is `main`** (`.github/workflows/tests.yml`'s `pull_request: branches: [main]`) — so as a stacked PR it would have merged with no verdict at all, which is the one thing a stacked pair must not do quietly.

The cost is cosmetic and temporary: until [#888](https://github.com/Phaazoid/Godoiosis/pull/888) lands, the diff here shows **both** slices, because this branch is built on that one. The moment #888 merges, this collapses to slice 2 by itself.

**So: merge #888 first, then this.** Merging this one alone would take slice 1 with it, which is harmless but leaves #888 showing as unmerged with nothing in it.

— Claude (Opus 5) · 2026-09-10

Phaazoid:
🤖 Claude says:

**Rebased, green, and there was nothing to resolve by hand.**

The conflict was history rather than content. [#888](https://github.com/Phaazoid/Godoiosis/pull/888) merged as a **squash**, so none of this branch's slice-1 commits are ancestors of `main` — the same changes now arrive by two unrelated routes, and every file both slices touched reads as a conflict even though the two never disagree about a line.

`git rebase --onto origin/main <slice-1 tip> <this branch>` replays only slice 2's own three commits onto the squashed main. All three applied with **zero conflicts**, and the diff here is now slice 2 alone: 19 files, no slice-1 leftovers.

Your `authoring` commit is untouched — it is content (`Small Cone.tres`, the two Chemical Spitter files) and overlaps nothing here.

**Verified rather than assumed.** Ran `presentation`, the water law and `elemental` on the rebased tree: **790 cases, 0 failures**, which includes slice 1's own suite off `main` beside slice 2's. CI is **6/6 green**.

I also deleted the landed `feat/887-arc-lightning`, after checking by CONTENT rather than by SHA: a squash makes `git diff main...<branch>` list the parent's whole diff and read exactly like unlanded work, so the check was grepping `main` for symbols only that diff introduced (`volley_struck`, `add_beam_strip`, `EFFECT_RENDER_PRIORITY`, `arc_links`, `trace_point` — all present).

Ready to merge whenever you are.

— Claude (Opus 5) · 2026-09-10

Issue #937, opened 2026-09-12
Replace the Fire Emblem placeholder sprites before this build is listed anywhere


Phaazoid:
🤖 Claude says:

**Filed at the dev's call, 2026-09-12, at P0.** His words, after weighing shipping them once more against swapping now:

> *"Let's file it under P0, probably better to bite the bullet earlier rather than later. We can replace them with Sara's sprites for now."*

The first half stands. **The second half does not execute** — see the blockers below, which is the substance of this ticket.

## What ships today

`Art/Units/MapSprites/` — **108 files, 36 sprite triples** (idle / `_Moving` / `_Downed`), rips from *Fire Emblem: The Blazing Blade* (GBA). `Art/Units/ZoomAnimations/` — **4 battle-scene sheets** plus the archived downloads, feeding `tools/zoomanim/gen_zoom_animations.gd` and its generated `<Unit>_Frames.png`.

Neither folder is in `exclude_filter`, so all of it is in the pack. These are [#139](https://github.com/Phaazoid/Godoiosis/issues/139)'s reddest column: **the one class of asset that cannot be licensed at any price.**

## Why P0 rather than "before launch"

The risk is a function of **reach, not existence** — an unlisted page shared with a few friends is genuinely low-risk, and the realistic bad outcome is a takedown of the itch page rather than anything worse. But the exposure **scales with exactly the outcome the demo is for**: if it lands, word spreads, and the swap becomes urgent at the moment it is most disruptive. Doing it while the build is still pre-release costs a content pass; doing it after costs a content pass plus a live page.

## Why Sara's stash does not cover it

Permission is not the issue — [#139](https://github.com/Phaazoid/Godoiosis/issues/139) records her grant as unrestricted with a credit condition. **Coverage is.** Measured against `C:\Iosis\old-game-sprites`:

| blocker | detail |
|---|---|
| **4 unit looks against 14 needed** | Her humanoid sets are sword / spear / drill / alchemist (plus a boss and `square`, a board overlay). Authored content references **14 distinct looks**: Brigand, Archer, Mage, Recruit, Fighter, Priest, Pirate, Myrmidon, General, Druid, Sage, Basic_Soldier, Knight, Dancer |
| **No downed frame exists** | Nothing in the stash is a downed/death/hurt pose. `UnitData.downed_sprite` is a required field and downed is a live state with its own countdown |
| **Incomplete sets** | `alchemist` has idle and attack but **no move frames** |
[...]

Issue #139, opened 2026-08-04
Asset licensing for public distribution


Phaazoid:
🤖 Claude says:

**Demo milestone.** An audit and a decision, not a feature.

## The ask

Every asset shipped in the exported build needs a license permitting **public distribution**. That covers:

- The placeholder art currently in `Art/`
- Any audio added for the demo (see the audio issue — it has none yet, which makes this the cheap moment)
- Fonts
- Anything pulled from an asset pack or a free-asset site

## Why now rather than later

From the milestone note: *cheap now, expensive after the build is out.* The asymmetry is real — before distribution, a wrongly-licensed asset is a swap; after, it is a takedown, a rebuild, and a re-send to everyone who has the zip. The demo's whole point is that the zip goes to people you do not control.

## Deliverable

A per-asset provenance record — where each came from and under what terms — plus whatever attribution those terms require, shipped in the build. A `CREDITS.md` or an in-menu credits screen; the main menu from the player-shell issue is the natural home for the latter.

Assets you made yourself still belong in the record. The record's job is to answer "where did this come from" a year from now, and your own memory is not that record.

## Related

- Audio (its assets are the ones not yet acquired — do this before, not after) · the player-facing shell (credits screen home) · the art pass

**Milestone:** Demo — a build you can hand to a stranger

— Claude (Opus 5) · 2026-08-04

Phaazoid:
🤖 Claude says:

**Audio provenance — 15 assets, the first rows for the per-asset record.**

Recovered 2026-09-10 from `c3potheds/iosis` @ `master`, path `Iosis.gmx/sounds` — the old GameMaker
game, private repo. Staged locally, not imported. Extraction note and the inventory-vs-scope mapping are on #136.

## Music — 5 tracks

Original work, **Simeon Anfinrud**, album tagged "Iosis", 2014–2015. ID3-tagged on every file, so the
attribution is self-documenting.

| asset | track title | year |
|---|---|---|
| `snd_music_title` | Splendor of Adventure | 2014 |
| `snd_music_overworld` | Journey | 2014 |
| `snd_music_boss1` | Boss Battle 1 | 2014 |
| `snd_music_enemy_battle` | Enemy Approaching | 2014 |
| `snd_music_battle1` | Battlefield | 2015 |

**Terms: unknown.** A named composer's original work, made for the old project. Whether that extends to
the Godot project, and on what terms, is a question for him — the files cannot answer it.

## SFX — 10 attack sounds

[...]

PR #938, opened 2026-09-12
#139: a credits screen, and the two names a licence makes mandatory

Phaazoid:
**You are on `feat/139-credits-screen`** — boot it, hit **Credits** on the title screen.

Advances [#139](https://github.com/Phaazoid/Godoiosis/issues/139). The deliverable that issue has named since 2026-08-04, and the thing that converts two grants into compliance rather than a plan.

## Two names ship because a licence says so

Neither is a courtesy, and the second was found while building this:

| who | for | the condition |
|---|---|---|
| **Sara Shen** — @stargarnishstudio | the element runes, weapon icons, the logo | *"fine with us using whatever, and just to include her name and insta link somewhere"* — dev, 2026-09-10 |
| **Jamie Brownhill** — World of Solaria | `Art/Board/Solaria Demo Tiles.png`, the live 2D board tileset | **licence §2(b)(6)**: *"The Licensee agrees to grant proper credit to the Licensor where due"* |

**The Solaria licence also resolves a whole #139 row that was open this morning** — unlimited use in commercial and non-commercial products, so it clears public distribution outright. The pack is at `C:\Iosis\World of Solaria Demo Pack Update 04`; its terms also forbid resale and AI training, neither of which this build does.

## Shape

`Glossary.gd` → `GlossaryScreen.gd` one domain over, because that is the surface this is:

- **`Classes/core/Credits.gd`** — the data. `required` is a licence condition, not emphasis.
- **`Classes/ui/CreditsScreen.gd`** — `extends ModalCard`, pure projection, bounded scroll, Esc closes.
- `MissionSelectScreen` gains `credits_chosen` + a row; `MissionController` gains one connect line.

**Title screen only, not the pause menu** — a deliberate break from the Glossary's two doors. [#724](https://github.com/Phaazoid/Godoiosis/issues/724) already calls that list *"a build history, not a decision"* and leaves the ordering as your open call; a ninth row makes your own ticket harder to settle. One line in `PauseMenu` whenever you want it in both.

## Two documents, not one

#139's body says *"a `CREDITS.md` **or** an in-menu credits screen"*. It is both, because they answer different questions and merging them puts *"terms: unknown"* in front of a player:

- **`Credits.gd`** — whose name must appear in the build. Player-facing, only what ships.
- **The provenance record** (#139's other half) — where each file came from and whether we may ship it. Internal, complete, and it necessarily holds rows that get **no** credit because they cannot be licensed at all.

## The Fire Emblem sprites are deliberately absent

Recorded in `Credits.gd`'s header so the omission is not "fixed" later. Crediting an asset that cannot be licensed does not license it, and naming *The Blazing Blade* on a public page is an admission that ships with the zip for no benefit. They are [#937](https://github.com/Phaazoid/Godoiosis/issues/937) and #139's record.

`gdUnit4` is likewise absent because `exclude_filter` keeps it out of the pack — not an oversight.

## How to check this

- **Do:** launch, click **Credits** on the title screen. Scroll it, press **Close**, then press **Esc** on a second opening.
- **You should see:** five sections; Sara's handle and instagram link under her name; Jamie Brownhill under Board tileset. Closing either way returns you to the title screen with the game responsive.
- **A failure looks like:** a name with no link under it (the grant's condition unmet), the page opening over a frozen title screen that stays frozen after Close, or Esc doing nothing (the #418 room-with-no-door shape).
- **What the tests already cover:** the wire (row exists → `credits_chosen` → page opens), the thaw on close, Esc, and **every `required` row's name AND detail reaching the page** — asserted as a property over `Credits.required_entries()`, never as an expected list, so re-wording a credit or adding a fourth grant does not red anything. The law suite guards that set being non-EMPTY, without which that loop passes having asserted nothing.
- **What is NOT covered:** whether it LOOKS right — spacing, the section rhythm, whether the link is readable at the detail colour. That is your eye.

## Verification

`ui` + `law` — **698/698, exit 0**.

**5 mutants, 5 caught:**

| mutation | caught by |
|---|---|
| sever `credits_chosen.connect` | the wire case |
| rename the `Credits` row | the row assertion |
| `required_entries()` returns empty | the law's non-empty guard |
| `_add_entry` skips the detail line | named both required credits by name |
| `_on_cancel` returns false | the Esc case |

One seam was removed during the build: a `shown_text()` on the screen, which would have been a second answer to a question the test's own label walker already asks.


Phaazoid:
**Follow-up commit: equal development credit, and Daniel's own name.**

Dev, 2026-09-12:

> *"Simeon and I need equal credits at the top for development. This most recent build of the game may be 90% coded by me, but the previous one from 10 years ago was 90% him, and the ideas synthesis has been a fully collaborative effort."*

So the Development section is now two rows with an identical role, **alphabetical by surname** so the order signals equality rather than a lead — and the reasoning is in `Credits.gd`'s header rather than only in this thread, because the tempting "correction" later is to rank them by who wrote the code in *this* repo.

**Simeon keeps his Music row.** Composing the score is a separate contribution, not a second billing.

`Phaazoid` → **Daniel Manzella**, at his request. Grepped: no other shipped file named the handle.

Both credits suites re-run — **8/8, exit 0**.

PR #960, opened 2026-09-14
CI: one Godot import at a time on a shared host

Phaazoid:
Three shards importing at once on the self-hosted VM crashed Godot's import **three times today** — exit 134, *"caller thread can't call propagate_notification()"* in the filesystem scan — every time in the first wave, on a cold VM twice and a 40-minute-old one once; never on GitHub's image, where each job had a VM to itself. No OOM kill, no kernel or journal error: a thread race under three concurrent imports on eight vCPUs.

`flock /tmp/iosis-import.lock` around the import step serialises that phase across jobs sharing a host (~20 s each, so ~40 s of wall clock for the second and third) and is a no-op on separate hosts. The first pass retries once; the second pass keeps its existing reason (the UID registry). Tests still run in parallel — only the phase that crashes is serialised.

Its own CI run is the first test: three first-wave imports queued behind one lock.


Phaazoid:
**The lock, measured.** This PR's own run was not a valid test — #959's rerun held the other runners, so its six jobs started one at a time and no two imports ever overlapped. A dispatched run with all three runners idle ([34890077218](https://github.com/Phaazoid/Godoiosis/actions/runs/34890077218)) is the proof: three first-wave jobs started in the same second and their imports queued behind the lock.

| shard | job start | import window | import step |
|---|---|---|---|
| pres-rest | 19:58:19 | 19:58:24 → 19:58:44 | 20 s |
| rest | 19:58:19 | 19:58:25 → 19:59:04 | 39 s (waited ~19) |
| ui-flow | 19:58:19 | 19:58:25 → 19:59:24 | 59 s (waited ~39) |

Non-overlapping, one at a time, no crash. Six of six green, **357 s wall** — inside the range of the unlocked runs (384–393 s), so the ~40 s the third job spends waiting is absorbed.

Also on the record from today: five import crashes before the lock (abort 134 ×3, segfault 139 ×1, one packed-boot error), every one a first-wave job with three imports running at once; zero under the lock so far.

Issue #982, opened 2026-09-16
Dev tools: create a Dialogic timeline from the Dialog page, instead of leaving for Dialogic's tab


Phaazoid:
Authoring a mission briefing is the one piece of board content that still requires leaving the dev overlay. The Dialog & Tutorial page (#397) can attach a timeline to a beat but cannot create one, so the sequence is: Godot editor -> Dialogic tab -> new timeline -> save into `Scenarios/dialog/` -> back to the overlay -> pick it in the beat row. The same door reopens one step later for any speaker who is not Torv: `dch_directory` holds exactly one entry.

Found live 2026-09-16, authoring the Terraces briefing. Dev: *"If preferable, I'd like the whole sequence doable in my editor."* Ruled 2026-09-16: **timelines, speakers, delete and preview all land in ONE PR.**

Terraces and The Dry Field are the two boards with no briefing. Causeway, Ford and Quarry each carry a `PRE_MISSION_START` beat pointing at their own `<mission>_intro.dtl`.

## Scope

On the Dialog page: a **TIMELINES** section (pick / New / Edit / Delete / Play, with a speaker + text line editor and Save) and a **New speaker from cast** action that mints a `.dch` from a `Resources/Units` character. Beat rows are unchanged except their dropdown now lists what was just made.

## The seam: `Classes/dev/DialogicSource.gd`

One answer to "how does a Dialogic resource get created, registered and removed from the running overlay" - generic over `.dtl`/`.dch`, the way `KnobSource` sits beside `GameTool`. Pure text rewrites where possible; the page owns no file logic. Plus `ScenarioDirector.preview(timeline)`: refuses while anything is talking, sets the latch, calls `_start`; bypasses the `SHOW_DIALOG` gate (`_fire`'s, for players).

## Rules the writer obeys - each one is a bug the first plan would have shipped

1. **The beat holds the `load()`ed, path-bearing resource, never an in-memory one.** `DialogicTimeline.events` is not an `@export`; an in-memory timeline EMBEDS into the mission `.tres` on Update and loads back empty (`tests/flow/test_pre_mission_draw.gd:365`, the #177 shape). Edits update the already-loaded object in place (`from_text` + `process`) so the reference stays file-backed and the change plays without a restart.
[...]

PR #994, opened 2026-09-16
Dev tools: a timeline name is a plain identifier, and the picker starts empty (#982 follow-up)

Phaazoid:
Follow-up to #988, for the bug found in play on #982: *"any dialog I add to it gets destroyed."*

## What was actually happening

It was never destroyed. `_suggested_name()` built the prefill from `ScenarioManager.display_name()`, which keeps the folder -- so on Terraces the name field read `missions/terraces_intro`, and the first Save wrote an **empty** `Scenarios/dialog/missions/terraces_intro.dtl`. You fixed the name, wrote your line, saved: that went into the flat file. Two files.

On the next editor Play, Dialogic's own `_build()` re-registered from disk exactly as designed -- which is why the option carried over -- but its scan visits the subfolder first, so the **empty twin took the plain name** `terraces_intro` and your real file was registered as `dialog/terraces_intro`. Picking `terraces_intro` opened the empty one.

A headless reproduction of the whole sequence with a single file is clean at every step. The fault exists only once the twin does.

## The fix

- `DialogicSource.name_block_reason` -- a name must be a valid identifier (the tightest rule that admits every shipped timeline and speaker), refused at BOTH writers, before anything touches disk.
- The prefill uses the file's basename.
- The picker's `select(-1)` moves to after the add loop. `add_item` auto-selects entry 0, so ahead of the loop it was undone at once and the page opened showing `causeway_intro` with nothing loaded.

## How to check this

**Do:** on Terraces, press **New**. The name should prefill `terraces_intro` -- no folder. Type `missions/x` into the name and press Save.

**You should see:** a refusal naming slashes, and no new folder under `Scenarios/dialog/`. Opening the page fresh, the timeline picker should show **no selection** until you pick one.

**A failure looks like:** a `missions/` folder appearing under `Scenarios/dialog/`, or the picker opening on `causeway_intro` with an empty line list.

**If you already have the stray folder from the earlier session:** delete `Scenarios/dialog/missions/`, then press Play once so Dialogic re-registers -- your real `terraces_intro.dtl` will take the plain name back.

**Covered by tests:** the name rule (slashed, dotted, empty and space-padded names refused; every shipped name accepted), the prefill for a subfolder mission, a refused save writing nothing, and the empty picker. Falsified: reverting the `select(-1)` order reds the picker case alone; letting the gate always allow reds both name cases and writes the stray folder.

**Not covered:** the layout, as before.

One correction on the record: my first diagnosis in chat blamed the editor overwriting `project.godot`. It does rewrite that file, but Dialogic's re-registration repairs it on the next Play, so the registry was never the missing half -- the name was.


Phaazoid:
🤖 Claude says:

Second commit pushed; the suites are green now and the first push was not -- it went out on a `grep` exit code instead of the runner's, which is mine.

The two red cases were **correct**: main already carried the bug's artifacts from `609afdce` ("authoring") -- the empty `Scenarios/dialog/missions/terraces_intro.dtl`, a registry with `terraces_intro` pointing at that twin and the real file under `dialog/terraces_intro`, and **`Terraces.tres` referencing the empty twin by path**, so the briefing you saved played silence.

The repair is one diff (the #596 rule -- a deleted content file sweeps its referrers): the twin and its `.uid` are gone, `terraces_intro` names the flat file holding your line, and the mission's `ext_resource` is repointed to it with its own uid. Verified by `test_dialogic_registries`, `test_resource_uid_references` and the shipped-mission load sweep.

**One content call I made rather than asked**, stated so you can reject it: the two candidates were a 0-byte file and the one holding `torv: gjh`, and the beat pointing at the empty one is the bug you reported. If you meant something else by either file, say so and I will redo it.

— Claude (Fable 5.1) · 2026-09-16

Issue #1042, opened 2026-09-19
Threat lines read as an aim: they need direction, motion, and a vocabulary of their own


Phaazoid:
Filed by Claude from the dev's notes watching a first-time player stream the demo (2026-09-18), not written by the dev directly. #710's threat lines had only just shipped.

**What he watched happen:** the lines confused the player, who read one as another way of aiming a fireball.

**Dev's call:** *"They need to pulsate from the enemy to the unit the enemy targets, they need an arrow indicating direction, and they need to be visually different from the targeting lines."*

## What is drawn today

`ThreatLines2D` is a pure renderer over a two-point polyline per intent:

```gdscript
const LINE_WIDTH := 1.5
static var INTENT_LINE_COLOR := Color(1.0, 0.8, 0.2, 0.95)
static var INTENT_FELL_COLOR := Color(1.0, 0.2, 0.15, 1.0)
```

`segment(from_cell, to_cell, board)` builds two points at `Reach.EYE_HEIGHT` and `_draw` strokes them. Three consequences, each matching one of the three asks:

1. **No motion.** `_draw` runs on `queue_redraw` and nothing animates between redraws, so the line is a static stroke.
2. **No direction.** The polyline is geometrically symmetric and both endpoints are built by the same loop. Nothing on screen says which end is the enemy and which is the victim — the player has to already know.
3. **It is the wrong colour to be distinct.** The aim footprint fill is `HOVER_MODULATE = Color(1, 1, 0)`, pure yellow. The intent line is `Color(1.0, 0.8, 0.2, 0.95)`, amber. Two members of one hue family, one meaning "I am aiming this" and the other "an enemy is aiming at you" — which is exactly the mistake the player made.

## The three asks, and what each one costs

**Pulsate from the enemy toward the target.** The direction of travel is the second half of ask 2 — a pulse that runs enemy → victim says which way the threat points without an arrowhead having to. It needs a clock; the project's own rule from #314 round 5 is **one clock and N offsets, never N tweens**, and `OverlayManager.intent_version` already exists as the mirror's change signal, so the store side is in place. Note the 3D path: `OverlayMirror` lifts the same points into the diorama, so any animation has to be expressed in the shared trace space or the two views will disagree.

[...]

PR #1061, opened 2026-09-20
#1059: the threat mark leaves from the body, bows, and ends in a cone

### LINKED ISSUE #1059: The threat mark leaves from over a unit's head, runs straight, and ends in an arrowhead that does not read
🤖 Claude says:

Filed off the dev's play of [#1042](https://github.com/Phaazoid/Godoiosis/issues/1048) (2026-09-20), in his words:

> They should be shallow arcs rather than straight lines, and the point of origin needs to come from the actual unit rather than over its head. The arrows also just don't look great in practice. Perhaps a narrow cone at the end instead?

## The origin complaint is exact, and measurable

In world units above a unit's feet:

| | world units | from |
|---|---|---|
| the median character's head | **0.625** | `MapSpriteInk.INK_RECT` — 20 texels ÷ `UnitSprite3D.texels_per_unit` 32 |
| **the mark, today** | **0.5** | `Reach.EYE_HEIGHT` 1.0 × `BoardSpace.ROW_HEIGHT` 0.5 |
| the crown | **0.85** | `BoardOverlays.billboard_lift` |

The mark leaves at **80% of body height**. `Reach.gd`'s own header records the dev's 2026-08-20 ruling that the line "should originate from the center of the sprite" — which is 0.3125. It has never been that.

## Two records this corrects

**The crown is ABOVE the mark, not under it.** [#1042's ruling comment](https://github.com/Phaazoid/Godoiosis/issues/1042) says the arrowhead's 0.4-cell inset was needed to clear the marker band, because the band sits "just under" the mark. It does not — `billboard_lift` 0.85 is 0.35 world units *above* the mark's 0.5. The round-1 mockup that produced that finding drew `EYE_HEIGHT` as a whole cell rather than half a level, and the map sprites at half their real size; both errors ran the same way. The inset is a taste value, not a clearance.

**A ramp mark leaves from the feet by coincidence, and the fix would break it.** `ThreatLines2D.segment` measures from `board.elevation_at(cell)`, which `BoardHeights` documents as a cell's **low side** — "for a ramp, the floor it connects upward from". A body on that ramp stands at the slope's visual midpoint, half a level higher (`verticality.md`; it is why `BoardSpace.surface_point` exists). Today's `EYE_HEIGHT` of 1.0 rule unit happens to equal that half level exactly, so the mark lands on the feet. Drop it to the sprite's centre and it sits ~0.19 world units *under* the slope, occluded, on every ramp cell in the game. The ends have to measure from the cell-centre surface (`Terrain.height_at_uv`, #427's one surface) instead.

## What was settled against a drawn mockup

Two rounds, 2026-09-20. The dev's rulings:

- **The mark leaves from the sprite's CENTRE** — 0.3125 world, his own 2026-08-20 ruling, measured.
- **It bows**, at the deeper of the two lifts drawn (0.45 of a cell on the card he ruled on). Drawn against a sideways swing as well; the vertical lift won.
- **A narrow converging cone replaces the arrowhead**, at **half the length it was first drawn at** — *"It should be subtle."* Settled at 0.45 long by 0.13 across, the narrowest of three widths offered, after it was pointed out that halving the length alone makes a cone blunter rather than subtler.
- **The travelling bead stays exactly as it is on main** — *"If by the bead, you mean the current glow that travels through on main, yes keep it. If you mean adding something new, no, good how it is."*

## The architectural note

`Reach.EYE_HEIGHT` is a **rule** constant — it defines what a wall is for `Reach.sight_trace` and the vertical aim gate. It cannot move for a look. So the mark gets its own height, and the split is declared rather than discovered: the sight trace draws the *trajectory the rule judges* and must hang where the rule says; an intent mark is not a trajectory — it is a statement about who attacks whom — so it may hang where the body is. The same split answers the ramp question, which is why `segment()` stops reading `elevation_at` while `sight_trace` keeps it.

## Cross-refs

`Classes/board/ThreatLines2D.gd` · `Classes/presentation/BoardOverlays.gd` (`add_beam_strip`, `set_marks`) · `Classes/presentation/sight_beam.gdshader` · `Classes/presentation/OverlayMirror.gd` (`_threat_intents`) · `Classes/weapons/Reach.gd` (`EYE_HEIGHT`) · `Classes/terrain/Terrain.gd` (`height_at_uv`) · [#1042](https://github.com/Phaazoid/Godoiosis/issues/1042) · [#710](https://github.com/Phaazoid/Godoiosis/issues/710) · [#674](https://github.com/Phaazoid/Godoiosis/issues/674) · [#44](https://github.com/Phaazoid/Godoiosis/issues/44)

— Claude (Opus 5) · 2026-09-20

Phaazoid:
Closes #1059.

Your three asks, settled against a two-round mockup rather than prose: the mark leaves from the sprite's **centre**, it **bows** (the deeper of the two lifts drawn, authored per cell of run), and the arrowhead becomes a **narrow cone** at half the length it was first drawn — *"It should be subtle."* The travelling bead is untouched.

## What the measurement found, including two things I had recorded wrongly

In world units above a unit's feet:

| | world | from |
|---|---|---|
| the median character's head | **0.625** | `MapSpriteInk.INK_RECT`, 20 texels ÷ `texels_per_unit` 32 |
| the mark, before this | **0.5** | `Reach.EYE_HEIGHT` 1.0 × `BoardSpace.ROW_HEIGHT` 0.5 |
| the crown | **0.85** | `BoardOverlays.billboard_lift` |

So the mark left at **80% of body height** — your report is exact — and `Reach.gd`'s header recorded your 2026-08-20 ruling that it should be the sprite's centre, which is 0.3125. It had never been that.

**#1042's record said the arrowhead's inset was needed to clear the crown. It never cleared anything.** `EYE_HEIGHT` is in RULE units and `billboard_lift` is in WORLD units, so the crown hangs a third of a cell *above* the mark rather than under it, and the guard ward is a ground sprite that was never in that band. The round-1 mockup drew `EYE_HEIGHT` as a whole cell and the sprites at half size; both errors ran the same way. The inset is a taste value. `Reach.gd` and `CLAUDE.md` both called `EYE_HEIGHT` "the sprite's centre" and `CLAUDE.md` still carried its pre-#427 value of 0.5 — both corrected.

## The one architectural decision, and the bug it uncovered

**`Reach.EYE_HEIGHT` cannot move for a look** — it defines what a wall is for the sight trace and the vertical aim gate. So the mark gets its own height, and the split is **declared**: a sight trace draws the trajectory the rule judges and must hang where the rule says; an intent mark is a statement about who attacks whom and may hang where the body is.

**That uncovered a latent bug.** `segment()` measured from `board.elevation_at`, a cell's **low side** — for a ramp, the floor it climbs *from* — while a body on that ramp stands at the slope's visual midpoint, half a level higher. `EYE_HEIGHT`'s 1.0 rule unit happens to *equal* that half level, so a ramp mark left at the feet **by coincidence**; at the sprite's centre it would have sunk 0.19 world units into the slope on every ramp cell in the game. It reads `Terrain.height_at_uv` now, #427's one surface, which agrees with `elevation_at` on flat ground.

## How it is built

- **The cone rides `UV2.y` as `scale − 1.0`.** A ribbon's width is a per-material uniform, so a shaft and a cone on one layer cannot differ by one; `UV2.y` was the only free per-vertex channel. The **offset** encoding is the design — zero means *unchanged*, so the sight bead and `ArcLightning`'s bolts are untouched by construction rather than by care, the same inert-by-default discipline #1042 gave the bead and flash uniforms.
- **A mark is two strokes** (the sampled arc, then the cone) sharing their join point, so the bead's chained world measure stays continuous with nothing double-counted. Two strokes rather than one polyline for #1042's reason: `beam_tangents` averages at a joint, and the arc→cone join is a deliberate width discontinuity.
- **The bow is per CELL of run.** "Shallow" is a property of the shape, so one absolute lift is a hump on an adjacent-cell mark and nearly flat on a nine-cell one.
- **The cone's width is a MULTIPLE of the shaft**, not an absolute — `intent_width` is a `BoardOverlays` export a 2D node cannot read, and an absolute would go stale anyway, since that setter re-pushes material parameters without bumping `intent_version`.

## Knobs

Mark **height**, **bow** and **inset**; cone **length** and **width**. Game tab, *Board markup colours*, all re-deriving the geometry through the existing restyle sweep. `HEAD_SIZE`/`HEAD_INSET`/`HEAD_SPREAD` are deleted rather than retuned, so a stale caller is a parse error.

## How to check this

**Do:** launch the branch, put a player unit in an enemy's reach, press **T** to see intents. Then find a ramp and stand a unit on it.

**You should see:** a pink mark leaving each enemy from the middle of its body, arcing gently over the gap and converging to a narrow point just short of its victim, with the bead still running enemy → victim. On a ramp the mark leaves from the feet, not from inside the slope. A felling mark still flashes white. Turn **photosensitivity** on: everything stops, the cone and the arc still draw, the lethal mark stays the brighter.

**A failure looks like:** a mark leaving at hair height again; a mark on a ramp starting below the ground; a cone that is a blunt wedge rather than a dart, or one pointing back at the enemy; a bead that restarts where the cone begins instead of arriving through it; a long mark bowing over the crown band.

**What the tests cover:** the height is asserted with a *probe value* rather than the tuned number, so it survives every retune and still reds if it goes back to reading `EYE_HEIGHT`; a ramp mark measures from the cell-centre surface; the arc bows above its own chord and the lift scales with the run; the cone converges *on the victim* and tapers to zero; the taper reaches the **mesh** (`UV2.y` read off `surface_get_arrays`), which is the case a static-only assertion would have missed entirely; a stroke with no widths asks for no change; the bead still chains and restarts. **Five mutants, each died to its named case** — height back at `EYE_HEIGHT`, a ramp measured from `elevation_at`, an absolute bow, a reversed taper, and `widths` dropped inside the strip builder.

**What they do not cover, and is yours:** whether the bow reads at the zoom you play at, whether the cone is subtle enough in motion, and whether dropping to the sprite's centre puts the mark behind bodies it used to clear. A headless suite cannot see motion at all — it can only assert the parameters reached the material. All five are sliders.

Suites run: `test_threat_preview` 29/29, `test_board_overlays` 51/51, `test_overlay_mirror` 54/54, `test_game_knobs` 52/52.


Phaazoid:
🤖 Claude says: a declared residual from the dev's look at this build, recorded here so it is not lost between slices.

**The cone is a see-through triangle, and it should be a solid 3D shape around the line** (dev, 2026-09-20: *"I was imagining a solid 3D shape around the line, not a see through triangle"*). Deliberately NOT fixed in this PR — his call: *"let's wrap it into the next slice"*.

**Why it looks that way, so the next pass does not re-derive it.** The cone is the same camera-facing RIBBON as the shaft (`sight_beam.gdshader`), tapering via the per-vertex width scale this PR added. That shader's whole look is a bright core fading to nothing at the rim — `falloff = pow(1.0 - edge, beam_softness)` — so every ribbon in the game is translucent and edgeless by construction. Widening or darkening it cannot make it a volume; a flat shape seen from any angle is still flat.

**What it will take, sketched rather than decided.** A solid cone of revolution about the mark's own axis is real geometry rather than a ribbon, and there are two seams it could ride:

- **The same `ImmediateMesh`, as its own surface.** The vertex stage widens by `beam_width * max(1.0 + UV2.y, 0.0)`, so `UV2.y = -1.0` leaves a vertex exactly where the mesh put it — real geometry can share the material without being billboarded. `UV.y = 0.5` puts it at the falloff's peak, i.e. solid. It would keep the bead and the flash for free, and cost no new layer. What it would NOT get is shading: the material is `unshaded`, so a cone would read as a flat silhouette of one colour unless the build fakes a gradient into the vertex COLOUR — which is baked in board space and therefore wrong the moment the camera orbits.
- **A real `MeshInstance3D` marker with a lit material** — a new `Kind` on `BoardOverlays.LAYERS` beside FILL/BRACKET/SPRITE/BILLBOARD/LINE. Honest shading, at the cost of a marker kind and of the bead/flash not applying to it.

**And the flat 2D view is a fork either way.** `ThreatLines2D` draws the cone as a filled polygon today, which is the right flat twin of a ribbon and cannot be the flat twin of a volume — so the next slice either keeps the flat triangle as a declared #292 asymmetry (the flash and the bead already are) or says something else there.

— Claude (Opus 5) · 2026-09-20

